Issue #281, opened 2022-09-14
Feature: Don't return error on empty queue (context-box)


Despire:
### Motivation:

After correctly printing error msgs to the console (#280) the `builder` and `scheduler` service are printing empty queue errors everytime they pull the `context-box` and the queue is empty

```bash
12:33PM ERR internal/worker/error.go:8 > error="error while getting config from the Builder: error getting builder config: rpc error: code = Unknown desc = empty Builder queue" module=builder
12:33PM ERR internal/worker/error.go:8 > error="error while getting config from the Builder: error getting builder config: rpc error: code = Unknown desc = empty Builder queue" module=builder
```

```bash
12:34PM ERR internal/worker/error.go:7 > error="error while getting Scheduler config: error getting scheduler config: rpc error: code = Unknown desc = empty Scheduler queue" module=scheduler
12:34PM ERR internal/worker/error.go:7 > error="error while getting Scheduler config: error getting scheduler config: rpc error: code = Unknown desc = empty Scheduler queue" module=scheduler
```

An empty queue shouldn't be an error. 

If we return nil from the `context-box` on an empty queue in either `GetConfigScheduler` or `GetConfigBuilder` it will fail with the following error
```bash
2022/09/14 13:08:45 ERROR: [core] [Server #1] grpc: server failed to encode response: rpc error: code = Internal desc = grpc: error while marshaling: proto: Marshal called with nil
```

### Description:

Review if we really need to return an error if the `builderQueue` or `schedulerQueue` is empty

### Exit criteria:
- [ ] Don't return an error if the queue is empty

PR #298, opened 2022-09-27
return empty config instead of an error

jaskeerat789:
# Description

This PR resolves the issue pointed in #281. We don't want to send an error when scheduler/builder tries to get configs from empty queues in c-box. c-box now sends a gRPC response with an `nil` config.


Despire (approved):
LGTM. Nice simple solution.

PR #301, opened 2022-09-29
Feat/terminatio grace period

### LINKED ISSUE #283: Feature: Termination grace period of all pods
### Description:

As was mentioned in the discussion https://github.com/Berops/claudie/issues/192#issuecomment-1199244554. Review having an termination grace period for pods.

Termination grace period in K8s ensures that upon a grace termination signal is sent to a pod, the service running inside this pod has been given enough time to finish processing the work unit. This prevents a service disruption from the client perspective when the service pods need to be terminated. This is the scenario that happens during normal operation activities, i.e. loading new service versions, pod eviction, deployment update, etc...

What happens technically, is that the container receives a SIGTERM signal. Then the grace termination period starts and once it times up (unless the service has terminated by then) it receives a SIGKILL signal that kills the service and eventually the whole container.

**We need to make sure the SIGTERM is passed from the process with PID 1 in container (typically the shell) to the golang service that should process it.**
Probably [this stackoverflow question](https://unix.stackexchange.com/questions/146756/forward-sigterm-to-child-in-bash) migh help.

In context of Claudie, probably the services which are typically asking for new work-units (I'm looking at you `builder`) should stop asking for a new work unit. For services that are not asking for new work units, but instead are contacted by clients for processing new work unit (I'm assuming e.g. `terraformer`), should not have difficulties to comply, because I expect that the pod endpoint in `k8s` is removed and therefore the particular pod under the grace termination period is removed from the list of pods behind the particular service and no new work will be sent towards this pod. Subject to a double-check.

### Exit criteria:
- [x] Implement a logic for the pull-based services (e.g. `builder`) to switch into an idle state upon receiving SIGTERM
- [x] Ensure the SIGTERM signal is passed by shell to all golang processes
- [x] Ensure the `kube-proxy` properly takes out the pod endpoint from behind the push-based services (e.g. `terraformer`) so that services become idle before they're terminated
- [x] Figure out if a desired behavior for a service that becomes idle after receiving SIGTERM is to terminate. If so, implement this behavior for all our services.

Despire:
Closes #283 

- [x] Ensure the SIGTERM signal is passed by shell to all golang processes

    All of our services are started as PID 1 inside the container thus the SIGTERM signal will be propagated to all of them correctly. I've researched why this is the case and it turns out if we use the exec form in the Dockerfile it will get assigned a PID 1, however if the shell form is used it will be spawned as a child process.
    more on this can be found here:
     - https://docs.docker.com/engine/reference/run/#pid-settings---pid
     - https://phoenixnap.com/kb/docker-cmd-vs-entrypoint
     - https://docs.docker.com/engine/reference/commandline/kill/#description
  
- [x] Ensure the kube-proxy properly takes out the pod endpoint from behind the push-based services (e.g. terraformer) so that services become idle before they're terminated

    I've closely followed the iptables in the `kube-proxy` for the specified services when they're terminated. After a `SIGTERM` signal is received by the pod, for example by calling `kubectl delete pod`, The endpoint and also the entries in the iptables will be removed for this particular pod and new entries will be added after the newly spawned pod is assigned an ip address.
    

For example, the following 2 screenshots are made right after `terraformer` received a SIGTERM signal, all of the entries from the iptables and endpoints are removed

deletion_iptable

deletion_pods

The next 2 picutures are made right after a new pod for `terraformer` was created and assigned an IP. The old pod is still terminating and since the IP tables contains ip address for the new pod it will also receive any new traffic leaving the old one idle until it terminates gracefully

after_deletion_iptable

after_deletion_pods

- [x] Implement a logic for the pull-based services (e.g. builder) to switch into an idle state upon receiving SIGTERM
- [x] Figure out if a desired behavior for a service that becomes idle after receiving SIGTERM is to terminate. If so, implement this behavior for all our services.

The above 2 points were implemented such that every service, including the builder, performs a graceful shutdown either on SIGTERM or SIGINT, [info regarding signals can be found here](https://www.gnu.org/software/libc/manual/html_node/Termination-Signals.html), meaning they always finish whatever work they're currently working on and after that they terminate, as leaving them idle until kuberentes issues a SIGKILL would waste resources.

Since the default termination grace period in kubernetes is 30 sec [ref](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination) I've changed the timings for each service as 30 seconds simply isn't enough for them to gracefully terminate.
- Builder - 90 mins for graceful shutdown, this is on par with the builderTTL, also including some time for potential retries
- ContextBox - 5 mins for graceful shutdown
- Terraformer - 22 mins for graceful shutdown, also including some time for potential retries
- Kuber - 22 mins for graceful shutdown, also including some time for potential retries
- KubeEleven - 22 mins for graceful shutdown, also including some time for potential retries
- Frontend - 5 mins for graceful shutdown
- Ansibler - 22 mins for graceful shutdown, also including some time for potential retries
- Scheduler - 5 mins for graceful shutdown

I've chosen these numbers to give enough time for each service to gracefully terminate. These numbers may be revisted in the future if they're too short.

### DIFF
--- a/services/builder/configProcessor.go
+++ b/services/builder/configProcessor.go
@@ -20,6 +20,7 @@ We can have three cases of a operation within the input manifest
 
 import (
 	"fmt"
+	"sync"
 
 	"github.com/Berops/claudie/internal/utils"
 	"github.com/Berops/claudie/proto/pb"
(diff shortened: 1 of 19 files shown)


bernardhalas:
Very good PR description @Despire , thank you. While I let the others to comment on the code changes themselves, I just have one change I'd like to reflect on. I think at some point in time we need to look into potential ways how to decrease the TTL values (and graceful periods too). Waiting 3 hrs for a SW load to finish in case of 2 replicas, this will be a lot.

Despire:
@bernardhalas It depends on what size the cluster being build is.
I've been actually curious about this and made some build-time tests:

 9:30 min - 1 hetzner 1 gcp node.
 16:40 min - 20 hetzner nodes
 27:00 min - 30 hetzner nodes 1 GCP node
 
 The larger the cluster is the more time it requires, (and if more cloud providers are used e.g gcp, aws, azure... this may also add some time to building the cluster)
 
Thus for large clusters the current builderTTL and termination grace periods will simply not be enough, though I might be just overthinking this.

bernardhalas:
That is interesting data. Frankly speaking, I think we can design this flow better. Asynchronous calls to the rescue.

`builder` should be getting a new work-unit from the `c-box` and then task `terraformer` with the request to build the infra, but then it should be able to drop the context. Once `terraformer` is done, any builder instance should be able to load the context again a and call `ansibler` and drop the context again. This way we:
- would be able to work with much shorter grace periods
- increase resiliency (as the context is loaded just for a short period of time in case of any interruption)
- we can build larger clusters (as based on your data above, it seems like every TTL will result in some cluster size limit, right?)

If you guys agree with this discussion, probably a new task is worth it. I wonder what's your feedback @jaskeerat789 @MiroslavRepka

MiroslavRepka:
I would very much like to have a discussion about it @bernardhalas.

>builder should be getting a new work-unit from the c-box and then task terraformer with the request to build the infra, but then it should be able to drop the context. 

Just a few clarifications. The bottleneck is not in the "control" units of a Claudie (cbox, builder). While it is true, the builder is holding a context for each manifest, this is done in separate goroutines, thus manifests are not blocking each other. The optimization you proposed would be optimizing just the resources used in Builder, but they would have zero to no effect on the whole workflow.

services/scheduler/scheduler.go
```diff
@@ -41,34 +43,47 @@ func processConfig(config *pb.Config, c pb.ContextBoxServiceClient) error {
...
 		return nil
 	}
+
+	if wg != nil {
+		// we received a non-nil config thus we add a new worker to the wait group.
+		wg.Add(1)
+	}
+
+	go func() {
+		if wg != nil {
```
MiroslavRepka, line 70: is this second if necessary?

services/builder/configProcessor.go
```diff
@@ -36,62 +37,72 @@ type nodeCount struct {
...
+		return nil
+	}
+
+	if wg != nil {
+		// we received a non-nil config thus we add a new worker to the wait group.
+		wg.Add(1)
+	}
+
+	go func() {
+		if wg != nil {
```
MiroslavRepka, line 62: ditto

MiroslavRepka (commented):
Hi @Despire, please have a look at the comments I left.

Overall, I think it's quite good, however, I would just like to clarify a few things.

If the pod gets a termination signal, is it able to start working on new Configs, or will it get blocked?

If the Builder gets a termination signal, it will wait until all current calls return, but will not initiate new calls (i.e. Terraformer returned but Ansibler will not be called), rather it will save the current state of the config into the DB?

services/scheduler/scheduler.go
Despire: Yes, calling wg.Done() on a nil receiver would panic

MiroslavRepka:
Thanks for the clarifications!

>If builder receives a SIGTERM, it will finish whatever it's working on, so if it recieves the signal during a gRPC call to terraformer it will continue with ansibler, kube-eleven kuber and finishes with an update to the DB.

Is this the behaviour we want? @jaskeerat789

services/scheduler/scheduler.go
MiroslavRepka: Yes, but 3 lines above, you are already checking for a nill `wg`. Is there a chance that by the time it will trigger a goroutine, it will become nil?

Despire: no, the WaitGroup will always outlive the spawned go-routines, I added this check if someone wanted to call the function without a WaitGroup, basically if the parameter is `nil` these calls will be ignored, if not `nil` they will use the WaitGroup

jaskeerat789:
I am satisfied with the implementation as well. Thanks to @Despire for the clarification as well as the amazing work.

jaskeerat789 (approved):

Issue #295, opened 2022-09-22
Proposal: Modify the `.tpl` files


MiroslavRepka:
### Motivation:

The way, we define the infra in terraform is kind of clumsy at the moment. We define the K8s and LB node pools similarly, however, we use 2 different `.tpl` (`<provider>.tpl` and `<provider>-lb.tpl` ) files for them and in each, we define a VPC, subnet, firewall and node specification. 

I propose to change the way we use `.tpl` files to become more flexible. 

We can define two templates per provider, one for node pool definition and one for networking. 
- the node pool definition would only define the instances (VMs) + ssh keys if supported
- the networking definition would define a VPC, subnet and firewall rules

This would give us flexibility when defining the networking side of things (firewall rules) while lifting the burden of maintaining 2, similar files.

In the end, each provider would have 
- `<provider>-node.tpl`  for node pool definition of K8s and LB nodes
- `<provider>-networking.tpl` for networking definition for K8s and LB nodes

### Description:

The `.tpl` files would be modified to implement the proposed way of structuring terraform code.

### Exit criteria:
- [ ] The new `.tpl` files were designed
- [ ] The new `.tpl` files were tested

Despire:
Was never groomed

PR #297, opened 2022-09-26
Feat/terraformer retry policy

Despire:
Closes #282 by introducing  a retry policy for terraform commands with up to 10 retries (similar to ansible retry count).

PR #302, opened 2022-10-03
Feature/add oci provider

MiroslavRepka:
This PR adds OCI provider to a list of supported providers by Claudie.

Please, carefully review the templates, as they are the most important part of this PR.

Couple of things to note:
- The `oci.tpl` and `oci-lb.tpl` are almost identical. Issue #295 is already trying to fix this issue, however, for now, we will have a duplicates
- There was a connectivity issue between the OCI nodes, which was due to `iptables`. Since we are already defining the firewall via Security lists, I chose to disable them for now


Despire (commented):
I've gone through the PR and left some comments, just out of curiosity.

I've gone trough the templates and didn't find any error (I did compare it with the hetzner, and gcp templates and also read the OCI parts of the docs)

services/scheduler/state-builder.go
```diff
@@ -132,7 +132,7 @@ func createKeys(desiredInfo *pb.ClusterInfo) error {
 // makeSSHKeyPair function generates SSH key pair
 // returns key pair if successful, nil otherwise
 func makeSSHKeyPair() (keyPair, error) {
-	privateKey, err := rsa.GenerateKey(rand.Reader, 2042)
+	privateKey, err := rsa.GenerateKey(rand.Reader, 2048)
```
Despire, line 135: Nice catch

services/terraformer/Dockerfile
```diff
@@ -50,6 +50,8 @@ COPY --from=build /go/services/terraformer/templates /bin/services/terraformer/t
 COPY --from=build /bin/grpc_health_probe /bin/grpc_health_probe
 COPY --from=build /bin/grpc_health_probe-Liveness /bin/grpc_health_probe-Liveness
 
+RUN apk add bash
```
Despire, line 53: Why are we adding bash to the docker image ?

services/terraformer/templates/gcp.tpl
```diff
@@ -57,7 +57,7 @@ resource "google_compute_instance" "{{ $nodepool.Name }}" {
   allow_stopping_for_update = true
   boot_disk {
     initialize_params {
-      size = 10
+      size = "{{ $nodepool.DiskSize }}"
```
Despire, line 60: Why was the size 10 originally ?

services/terraformer/templates/oci.tpl
```diff
@@ -0,0 +1,153 @@
...
+
+resource "oci_core_subnet" "claudie_subnet" {
+  provider = oci.k8s-nodepool
+  vcn_id = oci_core_vcn.claudie_vcn.id
+  cidr_block = "10.0.0.0/24"
+  compartment_id = var.default_compartment_id
+  display_name = "{{ $clusterName }}-{{ $clusterHash }}-subnet"
+  security_list_ids = [oci_core_vcn.claudie_vcn.default_security_list_id]
+  route_table_id = oci_core_vcn.claudie_vcn.default_route_table_id
+  dhcp_options_id   = oci_core_vcn.claudie_vcn.default_dhcp_options_id
```
Despire, line 34: Why are we explicitly setting these values ? Shouldn't it be set by default to the VCN values ? [ref](https://docs.oracle.com/en-us/iaas/tools/terraform-provider-oci/4.95/docs/r/core_subnet.html)

services/terraformer/templates/oci.tpl
```diff
@@ -0,0 +1,153 @@
...
+
+resource "oci_core_internet_gateway" "claudie_gateway" {
+  provider = oci.k8s-nodepool
+  compartment_id = var.default_compartment_id
+  display_name   = "{{ $clusterName }}-{{ $clusterHash }}-gateway"
+  vcn_id         = oci_core_vcn.claudie_vcn.id
+  enabled = true
+}  
+
+resource "oci_core_default_security_list" "claudie_security_rules" {
```
Despire, line 45: Is there any difference between [oci_core_security_lists](https://docs.oracle.com/en-us/iaas/tools/terraform-provider-oci/4.95/docs/d/core_security_lists.html) and `oci_core_default_security_list` ?

services/terraformer/Dockerfile
MiroslavRepka: Bash is a requirement for retry policy and is missing since #297

services/terraformer/templates/gcp.tpl
MiroslavRepka: No idea, it has been there for ages

services/terraformer/templates/oci.tpl
MiroslavRepka: Probably as a result of a days-long debugging, I specifically added them. We can omit it, but I would rather be explicit.

services/terraformer/templates/oci.tpl
MiroslavRepka: The `oci_core_default_security_list` is used to configure the default resource created for every VCN, whereas `oci_core_security_list` defines new security list.

services/terraformer/templates/oci.tpl
Despire: Understandable, we can keep them

services/terraformer/Dockerfile
Despire: Thanks for spotting this mistake.

Despire:
@jaskeerat789 you can give the green light after your review if you don't find any errors.

services/testing-framework/longhorn.go
```diff
@@ -11,8 +11,8 @@ import (
 )
 
 const (
-	maxLonghornCheck = 240 // max allowed time for pods of longhorn-system to be ready [seconds]
-	sleepSecPods     = 10  // seconds for one cycle of longhorn checks (the node and pod checks)
+	maxLonghornCheck = 10 * 60 // max allowed time for pods of longhorn-system to be ready [seconds]
```
jaskeerat789, line 14: Do we see change in time taken to deploy all longhorn-system pods with increase in number of nodes?

jaskeerat789 (approved):
Nice work Miro, I am happy with the changes.

services/testing-framework/longhorn.go
MiroslavRepka: Yes, I saw increased time to deploy pods into the cluster, however, I wouldn't say it was due to an increased number of nodes, more. Maybe the lack of compute power on the nodes chosen

docs/config/config.md
```diff
@@ -4,144 +4,149 @@ Config is a datastructure which holds all of the data for a Claudie microservice
...
-  | Credentials | string | Credentials of the provider |
-  | GcpProject | string | GCP project (only required when using GCP as DNS provider) |
-  | CloudProviderName | string | Cloud provider name. e.g. `gcp` and `hetzner`
+  | Name              | Type   | Description                                                |
+  | ----------------- | ------ | ---------------------------------------------------------- |
+  | SpecName          | string | Provider name                                              |
+  | Credentials       | string | Credentials of the provider                                |
+  | GcpProject        | string | GCP project (only required when using GCP as DNS provider) |
+  | CloudProviderName | string | Cloud provider name. e.g. `gcp`, `hetzner` or `oci`        |
+  | UserOcid          | string | OCID of the user                                           |
```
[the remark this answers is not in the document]
MiroslavRepka: This might be a good idea. I already thought about appending prefixes to variables, like `oci_`, `gcp_`, and `general_`, but I didn't want to "pollute" the PR more

Issue #279, opened 2022-09-13
Feature: Individual cluster deletion


MiroslavRepka:
### Motivation:

Currently, Claudie is able to manipulate multiple node pools on multiple clusters. However, due to my "error" during testing, I found out that Claudie is not able to delete the whole cluster. Here is an example:
1. Define an input manifest with a single k8s cluster in it and wait until it is applied
2. Add a second k8s cluster to your input manifest and wait until it is applied
3. Now delete one of the clusters from the input manifest, apply it and see what happens

In step 3, the Builder will throw a panic since it will try to access some `nil` reference, and the process will never finish. 

The obvious solution would be to add nodes from deleted cluster into slice of nodes to delete, but that would not work, since the Builder would try to gracefully remove the nodes from the cluster but 
- what if it will remove the control nodes first? The `kubectl` would not work and worker nodes would be left dangling with error.
- why do we need to gracefully remove them, when the whole cluster should be deleted? It would be faster to just schedule them for deletion in terraformer.

### Description:

Claudie should be able to delete individual clusters, not just nodepools or whole input manifest.

### Exit criteria:
- [x] Design for this scenario was proposed
- [x] The proposed design was implemented and verified
- [x] The new test-set for more "advanced" nodepool manipulation was designed and added (Not mandatory, just nice to have)

PR #303, opened 2022-10-05
Add support for individual cluster deletion

jaskeerat789:
# Description
This PR targets the feature described in #279. Now you can delete a cluster just by removing cluster definition from the input-manifest. 

# Task
- [x] Implement logic for cluster deletion
- [x] Update test-sets
- [x] Update docs


services/builder/configProcessor.go
```diff
@@ -216,3 +222,54 @@ func mergeDeleteCounts(dst, src map[string]*nodeCount) map[string]*nodeCount {
...
+// isEqual function checks if the two cluster from desiredState and Current state are same by comparing
+// names and hashes
+// return boolean value, True if the match otherwise False
+func isEqual(dsClusterInfo, csClusterInfo *pb.ClusterInfo) bool {
+	if dsClusterInfo.Name == csClusterInfo.Name && dsClusterInfo.Hash == csClusterInfo.Hash {
+		return true
+	}
+	return false
+}
+
```
Despire, line 269: This can be simplified, i.e.
```go
return dsClusterInfo.Name == csClusterInfo.Name && dsClusterInfo.Hash == csClusterInfo.Hash
```
But do we need this as a separate function? It's used only in one place and personally the bool expression `dsClusterInfo.Name == csClusterInfo.Name && dsClusterInfo.Hash == csClusterInfo.Hash` is clear to me.

services/builder/configProcessor.go
```diff
@@ -216,3 +222,54 @@ func mergeDeleteCounts(dst, src map[string]*nodeCount) map[string]*nodeCount {
...
+// updateCurrentState function will update the Clusters and LoadBalancerClusters for the config passed.
+// the function has side effects since it updated the config which is passed as a pointer reference
+func updateCurrentState(config *pb.Config, k8sCluster []*pb.K8Scluster, lbClusters []*pb.LBcluster) {
+	config.CurrentState.Clusters = k8sCluster
+	config.CurrentState.LoadBalancerClusters = lbClusters
+}
```
Despire, line 275: Ditto. Do we need this as a separate function? It's only used in one place and the body of the function are 2 assignments, It would be better to be inlined. i.e. don't abstract it in a function.

services/builder/serviceCaller.go
```diff
@@ -75,12 +75,27 @@ func buildConfig(config *pb.Config, c pb.ContextBoxServiceClient, isTmpConfig bo
...
 // destroyConfig destroys existing clusters infra for a config using terraformer and kuber
 func destroyConfig(config *pb.Config, c pb.ContextBoxServiceClient) error {
+	err := destroy(config, c)
+	if err != nil {
+		return err
+	}
+	// save changes to DB
+	err = cbox.SaveConfigBuilder(c, &pb.SaveConfigRequest{Config: config})
+	if err != nil {
+		config.CurrentState = config.DesiredState
```
Despire, line 85: Why are we doing this assignment ? `config.CurrentState = config.DesiredState`, The config its not being used after calling `destroyConfig` so this has no effect.

`destroy`, and `destroyConfig` are ambiguous names when first looking at them, maybe a more suiting name would be `destroyWithoutSave` and `destroyWithSave` as they essentially do the same thing but one saves to the DB and the other doesn't

services/builder/configProcessor.go
jaskeerat789: I just wanted to avoid having copies of this little long expression since we use this function twice( one in k8s loop and one in lb loop ). Although I agree that we can simplify the functions itself.

services/builder/serviceCaller.go
jaskeerat789: I agree with you but I was just following the pattern of updating `currentState` before calling the `SaveConfigBuilder` function. I think we should move the `currentState` updation to `SaveConfigBuilder` fucntion since fields like checksums and TTLs are being updated there.

Let me know your thoughts @MiroslavRepka

Despire: I've actually addressed this line of code as part of fixing the document deletion #309

Despire (commented):
I've tried the scenario described in the issue 
```
1. Define an input manifest with a single k8s cluster in it and wait until it is applied
2. Add a second k8s cluster to your input manifest and wait until it is applied
3. Now delete one of the clusters from the input manifest, apply it and see what happens
```

The service no longer panics and the cluster gets deleted.

Regarding the 2nd part of the issue

```
The obvious solution would be to add nodes from deleted cluster into slice of nodes to delete, but that would not work, since the Builder would try to gracefully remove the nodes from the cluster but

what if it will remove the control nodes first? The kubectl would not work and worker nodes would be left dangling with error.
why do we need to gracefully remove them, when the whole cluster should be deleted? It would be faster to just schedule them for deletion in terraformer.
```

@MiroslavRepka could you review this and check if the PR addresses this ? As I'm not 100% sure what is meant by that.

MiroslavRepka:
As far as I understand the code, the clusters, which need to be deleted are passed to the `destroy()` function in some sort of temporary config, which will delete them. The function `updateCurrentState()` will deal with the references to the already deleted clusters in the current state, so the remark in the issue is correctly addressed imo.

Despire (approved):

services/builder/configProcessor.go
MiroslavRepka: We can compromise; don't use the function and just add a comment about it so the behaviour is documented

services/builder/configProcessor.go
```diff
@@ -216,3 +222,51 @@ func mergeDeleteCounts(dst, src map[string]*nodeCount) map[string]*nodeCount {
 	}
 	return dst
 }
+
+// prepareConfigToDelete function queries for cluster those needs ro be deleted from current state.
+// It also updated the config object to remove the clusters to be deleted from current state.
+// returns *pb.Config which contains clusters (both k8s and lb) that needs to be deleted.
+func prepareConfigToDelete(config *pb.Config) *pb.Config {
```
MiroslavRepka, line 229: The name `prepareConfigToDelete` should be changed imo. The function is not taking care of the deletion of individual nodes from the node pool, it only checks for the whole cluster deletion. I would prefer if the name would reflect that, i.e. `getDeletedClustersConfig`

services/builder/configProcessor.go
```diff
@@ -63,14 +63,20 @@ func configProcessor(c pb.ContextBoxServiceClient, wg *sync.WaitGroup) error {
...
 			if err := destroyConfig(config, c); err != nil {
 				log.Error().Err(err).Send()
 			}
 			return
 		}
 
+		// check for cluster deleting
+		configToDelete := prepareConfigToDelete(config)
+		if err := destroy(configToDelete, c); err != nil {
+			log.Error().Err(err).Send()
```
MiroslavRepka, line 77: I wonder why we are changing `log.Error().Msgf()` to `log.Error().Err().Send`?

Afaik, the `Msgf()` already calls `Send()`, plus it gives us ability to format the message as we please.

From [docs](https://pkg.go.dev/github.com/rs/zerolog@v1.28.0#Event.Msgf):
```
Msgf sends the event with formatted msg added as the message field if not empty.
```

MiroslavRepka (commented):
LGTM, please see the comments. Good job @jaskeerat789

services/builder/configProcessor.go
Despire: I've used `log.Error().Err().Send` in #280, Although I think we should stay uniform across all service and just stick to either `log.Error().Msgf()` or `log.Error().Err().Send` personally since we mostly use the former I'd stick to it

jaskeerat789: I am personally fine with any approach but we need to stick to one for a consistent pattern.
@Despire

MiroslavRepka: Since we mostly use `log.Error().Msgf()`, it would make more sense to stick to it, as @Despire said.

@jaskeerat789, please make appropriate changes in your PR, we will deal with the rest later.

PR #304, opened 2022-10-05
Feat/aws

MiroslavRepka:
This PR adds AWS provider to a list of supported providers by Claudie.

Please, carefully review the templates, as they are the most important part of this PR.

As mentioned in #302, the `aws.tpl` and `aws-lb.tpl` are almost identical. Issue #295 is already trying to fix this issue, however, for now, we will have a duplicates


docs/config/config.md
```diff
@@ -117,6 +117,7 @@ Provider represent a single provider from manifest
   | TenancyOcid       | string | OCID of the tenancy                                        |
   | OciFingerprint    | string | Fingerprint of the private key saved in `Credentials`      |
   | CompartmentOcid   | string | OCID of the compartment                                    |
+  | AccessKey         | string | AWS access key                                             |
```
MarioUhrik, line 120: Isn't  AWS Secret key missing from here ?

services/terraformer/templates/aws.tpl
```diff
@@ -0,0 +1,160 @@
...
+  to_port           = 51820
+  protocol          = "udp"
+  cidr_blocks       = ["0.0.0.0/0"]
+  security_group_id = aws_security_group.claudie-sg.id
+}
+
+resource "aws_security_group_rule" "allow-icmp" {
+  provider = aws.k8s-nodepool
+  type              = "ingress"
+  from_port         = -1
```
MarioUhrik, line 111: This could be hardened to only allow `ping`, but I don't know if it's important or worth doing.
Might as well, though.
https://blog.jwr.io/terraform/icmp/ping/security/groups/2018/02/02/terraform-icmp-rules.html

MarioUhrik (changes requested):
This is a very partial review, I only looked on some files to a limited extent.

docs/config/config.md
MiroslavRepka: No, for the secret key, we are using the `Credentials` field, as mentioned in
```md
| OciFingerprint    | string | Fingerprint of the private key saved in `Credentials` 
```

jaskeerat789 (approved):
LGTM, Thanks for the PR @MiroslavRepka

MarioUhrik (approved):
LGTM. I didn't check thoroughly, but I see nothing wrong

Despire:
Since we now support more Cloud providers and Azure is on the way, is there a reason why we still only allows GCP for the [DNS zone](https://github.com/Berops/claudie/blob/master/docs/input-manifest/input-manifest.md#dns) in the input manifest ? (the validation only allows GCP)

MarioUhrik:
> Since we now support more Cloud providers and Azure is on the way, is there a reason why we still only allows a GCP provider for the [DNS zone](https://github.com/Berops/claudie/blob/master/docs/input-manifest/input-manifest.md#dns) in the input manifest ? ([the validation only allows GCP](https://github.com/Berops/claudie/blob/feat/aws/internal/manifest/validate_load_balancer.go#L70-L77))

This is a good topic to bring up again.
Our LB architecture requires a DNS record, because the DNS is what provides us with the HA aspect of the LB (as still shown in the docs).
Back then, Hetzner had no DNS, and we only had GCP besides that.
- Now, possibly, we might add support for non-GCP DNS.
- Or even redesign the entire LB architecture to avoid this burden, if somehow possible.

Note that the DNS thing is also the only part of Claudie architecture that isn't naturally on-premises compatible.

FYI @jaskeerat789 @MiroslavRepka @Despire

MiroslavRepka:
>Note that the DNS thing is also the only part of Claudie architecture that isn't naturally on-premises compatible.

Could you please elaborate @MarioUhrik?

MarioUhrik:
> > Note that the DNS thing is also the only part of Claudie architecture that isn't naturally on-premises compatible.
> 
> Could you please elaborate @MarioUhrik?

Everything else has been constructed with on-premises clusters in mind - storage, LB clusters, the way we create K8s...
DNS is the only aspect for which it is currently mandatory to actually use a cloud provider.

MiroslavRepka:
>Everything else has been constructed with on-premises clusters in mind - storage, LB clusters, the way we create K8s...
DNS is the only aspect for which it is currently mandatory to actually use a cloud provider.

Thanks for the clarification

Despire (approved):
LGTM! I don't see any issues with the changes

MiroslavRepka:
@Despire could you please review the changes in the recent [commit](https://github.com/Berops/claudie/pull/304/commits/c3fb929ffdc34b2d2c3b5679598f0ac8e63f6c24), regarding dynamic firewall creation?

After the review is done, we can merge this PR.

Despire:
Looks good to me. Great work in this PR @MiroslavRepka

Issue #205, opened 2022-07-25
Security hardening of firewall rules of Claudie-made clusters


MarioUhrik:
Within https://github.com/Berops/platform/tree/master/services/terraformer/templates, we could significantly improve security of the Claudie-made clusters by closing unnecessary ports from unnecessary IP ranges.

- Do we really need to use source IP range `0.0.0.0/0`, or can we narrow it down, e.g. to the Wireguardian CIDR?
- Close the protocol/ports that aren't needed (e.g. here we currently allow all TCP ports: https://github.com/Berops/platform/blob/master/services/terraformer/templates/gcp-lb.tpl#L33)

jaskeerat789:
Currently being worked in #306 so might be outdated

MarioUhrik:
We're leaving ICMP open, the above PR is merged

PR #306, opened 2022-10-06
feat: dynamic firewall creation

### LINKED ISSUE #294: Feature: Dynamic firewall creation
### Motivation:

Currently, the LB have static firewall rules defined in their `.tf` code. However, if we want to have a functioning LB, we need to dynamically create firewall rules to allow the user-defined ports/protocols in the LB specification (the `roles` section) in the firewall.  

### Description:

The firewall rules for LBs should be dynamically created based on the specification provided in the input manifest (the `roles` section).

### Exit criteria:
- [ ] LB define the firewall rules dynamically, based on the input manifest spec.

Despire:
Closes #294 

1. The firewall rules for LBs are dynamically created based on the specification provided in the input manifest.
   (by using the `protocol` and `port` fields from the roles section)
2. The port `6443` for the API server is created as a firewall rule for a k8s cluster only if no LB is attached to the cluster with a `target_port` of value `6443`


MarioUhrik:
Hi guys, just a question - I see the logic that if the LB config lacks a rule for 6443, then the worker nodes are opening port 6443 to everyone. That's fair.
But if we are using an LB with 6443, does the cloud provider accept 6443 traffic on the worker nodes coming via the LB?
Is that because you're using the Wireguard IP and the traffic is going via the Wireguard port ?

Despire:
That's what we were discussing with @jaskeerat789, The traffic should go via wireguard, but I haven't managed to test it yet.

Despire:
I’ve tested this with a cluster of 2 load balancers, 2 control nodes and 2 compute nodes. 
I’ve used `telnet` to ping port `6443` on LBs. The LBs load balanced the traffic between the two control nodes via Wireguard. 

To watch this I’ve used `tcpdump -ttttni any 'udp port 51820' and src host ‘ip address of load balancer’` from one of the control nodes.
 I’ve also watched the `wg0` interface for the traffic on the same control node for the same packets.
 
The traffic does reach the control nodes via Wireguard if the firewall rule is disabled for port `6443` on the cluster.

Also kubectl didn't had any problems communicating with the nodes via the LoadBalancers through port `6443`.

services/terraformer/templates/hetzner-lb.tpl
```diff
@@ -35,15 +35,17 @@ resource "hcloud_firewall" "firewall" {
...
   }
 
+  {{range $role := index .Metadata "roles"}}
   rule {
     direction = "in"
-    protocol  = "tcp"
-    port      = "6443"
+    protocol  = "{{ $role.Protocol }}"
+    port      = "{{ $role.Port }}"
     source_ips = [
```
MiroslavRepka, line 43: Can we limit source IPs to the CIDR block of Wireguard VPN? I think @MarioUhrik suggested that on the last grooming.

internal/utils/cluster.go
```diff
@@ -71,3 +71,22 @@ func FindName(realNames []string, name string) string {
...
 	return ""
 }
+
+// ExtractTargetPorts extracts target ports defined inside the role in the LoadBalancer.
+func ExtractTargetPorts(loadBalancers []*pb.LBcluster) []int {
+	ports := make(map[int32]struct{})
+
+	for _, c := range loadBalancers {
+		for _, role := range c.Roles {
+			ports[role.TargetPort] = struct{}{}
```
MiroslavRepka, line 81: Can we do something like 
```go
if _, ok := ports[role.TargetPort]; !ok {
	result = append(result, int(role.TargetPort))
	ports[role.TargetPort] = struct{}{}
}
```
thus omitting the second `for` loop

internal/utils/utils.go
```diff
@@ -0,0 +1,36 @@
+package utils
+
+import (
+	"fmt"
+	"strings"
+)
+
+// IsMissing checks if item is missing in the list of items.
+func IsMissing[K comparable](item K, items []K) bool {
```
MiroslavRepka, line 9: Nice to see some generics :slightly_smiling_face:

MiroslavRepka (commented):
LGTM, please see my comments about CIDRs. 

One more thing to add though. On Friday, we discussed that we could probably close the ICMP protocol in the firewall since the Ansible ping module uses ssh to ping the instance. The question is if we want to include it in this PR.

Please, see the reference in [slack](https://berops.slack.com/archives/C0190PTVA0Y/p1665147282310249).

services/terraformer/templates/hetzner-lb.tpl
MarioUhrik: I think Matus has perfectly nailed it as it is, as far as I could see.

> Can we limit source IPs to the CIDR block of Wireguard VPN?

This is the LB machine. It needs to open these ports to the entire internet. What you're suggesting is nonsense.

The flow is:
- LB nodes open 6443 to the entire Internet
- cluster nodes don't open 6443 at all
- LB nodes forward 6443 to the cluster nodes via their wireguard IP, so only the wireguard firewall rule needs to be present for the cluster nodes

Also:
- In case there is no LB, then cluster nodes open 6443 to the Internet in the same way the LB otherwise would

Please let me know if my understanding is correct @Despire @MiroslavRepka @jaskeerat789

MarioUhrik:
@MiroslavRepka @jaskeerat789 @Despire 

> One more thing to add though. On Friday, we discussed that we could probably close the ICMP protocol in the firewall since the Ansible ping module uses ssh to ping the instance. The question is if we want to include it in this PR.

In a separate thread later, Bernard pointed out he would like us to keep ICMP open for the time being.
There were no clear arguments for/against it.
So, for the time being, the conclusion is that we'll keep ICMP open.
The discussion happened here https://github.com/Berops/claudie/issues/205
BTW @Despire , I think that you'll probably complete https://github.com/Berops/claudie/issues/205 via this PR, please check

services/terraformer/templates/hetzner-lb.tpl
MiroslavRepka: >This is the LB machine

My bad, should have finished my coffee first :smile:.

I see the flow the same way you described it, therefore, please disregard my comment

internal/utils/cluster.go
Despire: Yes, that's a more simpler way

services/terraformer/templates/hetzner-lb.tpl
Despire: Exactly as @MarioUhrik described.

Despire:
> BTW @Despire , I think that you'll probably complete https://github.com/Berops/claudie/issues/205 via this PR, please check

Regarding the `0.0.0.0/0` I don't think we can narrow it down, ssh, icmp and also `6443` for the API server wouldn't work if the used the Wireguardian CIDR
Regarding the second task `Close the protocol/ports that aren't needed` yes, this PR also solves that, so this PR closes #205

MiroslavRepka (approved):

PR #311, opened 2022-10-13
remove tf statefile after manifest deletion

### LINKED ISSUE #227: Feature: Remove the state file when the whole manifest is deleted
### Motivation:

Once the input manifest is deleted from claudie, the terraform state file of the infra is useless and should be deleted.

### Description:

When the input manifest is deleted, the terraform state file for that manifest should be cleaned up

### Exit criteria:
- [ ] Delete the terraform state files once they are no longer needed

### Note:
Terraformer can take care of this since it has the minio go client already installed. ([`mc.RemoveObject()`](https://github.com/minio/minio-go/blob/master/docs/API.md#removeobjectctx-contextcontext-bucketname-objectname-string-opts-minioremoveobjectoptions-error))

Despire:
Closes #227 

This PR adds deletion of the statefile from minIO after the cluster has been deleted.

This PR also includes some code clean up


Despire:
Regarding using `Id()`, I don't think we can use it [here](https://github.com/Berops/claudie/blob/master/services/terraformer/server/clusterBuilder/clusterBuilder.go#L58) as we use `DesiredState` in that function and the `Id()` uses current state (as desired state doesn't exists when the cluster is being deleted)

MiroslavRepka:
>Regarding using Id(), I don't think we can use it [here](https://github.com/Berops/claudie/blob/master/services/terraformer/server/clusterBuilder/clusterBuilder.go#L58) as we use DesiredState in that function and the Id() uses current state (as desired state doesn't exists when the cluster is being deleted)

 What's more, `ClusterBuilder` does not even use the `Cluster` interface, which made me remove my comment.
 
 LGTM, good job!

MiroslavRepka (approved):

PR #335, opened 2022-10-19
fix panic when accessing Id()

Despire:
Fixes logic from #311 where the server would panic in the BuildInfra due to current state not existing

This will fix the issue by preffering DesiredState (in the BuildInfra) and using CurrrentState (in the DestroyInfra)


MiroslavRepka (approved):
LGTM, thanks for the fix!

PR #336, opened 2022-10-20
Feature/future domain check

### LINKED ISSUE #305: Feature: Move `CheckLengthOfFutureDomain` to Frontend
### Motivation:

Function `utils.CheckLengthOfFutureDomain` was implemented in issue #112, as a result of domains being too long ([explained here](https://github.com/Berops/claudie/issues/112#issuecomment-1015432224)). However, it was added to the Scheduler, since at the time, we did not have a Frontend. Now, it is time to move the check into the Frontend validation part.

Why open an issue for a simple fix? Well, the function must be redesigned since it consumes `*pb.Config`, but in the Frontend, we do not work with `*pb.Config`. 

### Description:

Redesign and move the logic of a function `utils.CheckLengthOfFutureDomain` into the Frontend validation part.

### Exit criteria:
- [x] Function have been redesigned to consume `manifest.Manifest`, instead of `*pb.Config`
- [x] Function have been moved to Frontend

MiroslavRepka:
This PR moves validation of a future domain to Frontend.

Closes #305


jaskeerat789 (approved):
LGTM Miro

Issue #241, opened 2022-08-16
Ensure that Terraformer has state file locking (Minio backend)


MarioUhrik:
By default, normal S3 buckets don't support state file locking.
But Minio buckets supposedly do.
Still, it's not really clear whether our current Terraformer implementation supports state file locking.
Just in case we'd have multiple concurrent Terraformers, we should have a look at this.

The objectives of this task are to:
- Check whether we already have fully functional Terraform state locking in Terraformer
- If not, explore what options do we have given the circumstances (Minio), then choose and implement a solution

PR #345, opened 2022-10-25
Add DynamoDB for TF state locking

jaskeerat789:
# Description
The issue #241 raised the question if state was enabled by minIO, which after investigation we found out that it was not. We use `s3` backend type in terraformer. `s3` backend depends on DynamoDB to enable state locking according to the [docs](https://developer.hashicorp.com/terraform/language/settings/backends/s3). Thus, this PR deploys local version of dynamodb and configures it so that it can be used by terraform `s3` backend


manifests/claudie/dynamo/job.yaml
```diff
@@ -0,0 +1,63 @@
...
+              "--table-name",
+              "claudie",
+              "--attribute-definitions",
+              "AttributeName=LockID,AttributeType=S",
+              "--key-schema",
+              "AttributeName=LockID,KeyType=HASH",
+              "--provisioned-throughput",
+              "ReadCapacityUnits=1,WriteCapacityUnits=1",
+              "--endpoint-url",
+              "http://dynamo:8000",
```
jaskeerat789, line 36: @MarioUhrik I wanted to use env vars for `endpoint-url` argument but didn't get it to work. I am unable to pass the variable in the command. I think i am using wrong syntax.

manifests/claudie/dynamo/dynamodb.yaml
```diff
@@ -0,0 +1,74 @@
...
+    targetPort: 8000
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: dynamo-pvc
+spec:
+  resources:
+    requests:
+      storage: 10Gi
```
jaskeerat789, line 72: @MarioUhrik 10Gi looks like a bit of overkill, should we have a lower value, something like 500Mi ?

MarioUhrik: Let's go for 1 Gi, but please make a little bit of a research to verify whether this is enough.

My unproven theory is that the only reason why DynamoDB has to exist, is because it may have a feature that allows atomic writes/reads. So perhaps, in theory, even a nearly empty file would suffice.

manifests/claudie/dynamo/job.yaml
MarioUhrik: If you don't get this to work, don't worry about it too much, it's not a dealbreaker at all

Makefile
```diff
@@ -45,10 +45,10 @@ dockerUp:
 	docker-compose --env-file ./manifests/claudie/.env up
 
 dockerDown:
-	docker-compose --env-file ./manifests/claudie/.env down
```
MarioUhrik, line 48: Pls double-check the diffs in the Makefile, they seem to bring inconsistencies

manifests/claudie/dynamo/job.yaml
MarioUhrik: But you can get it to work via setting an env variable from the configmap, and then referencing that env variable in the `args`

See https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#use-configmap-defined-environment-variables-in-pod-commands

manifests/claudie/dynamo/job.yaml
```diff
@@ -0,0 +1,63 @@
...
+                configMapKeyRef:
+                  name: env
+                  key: DYNAMO_HOSTNAME
+            - name: DYNAMO_PORT
+              valueFrom:
+                configMapKeyRef:
+                  name: env
+                  key: DYNAMO_PORT
+            - name: AWS_ACCESS_KEY_ID
+              value: fake
```
MarioUhrik, line 54: How is the authentication handled? Are the passwords `fake`, or is there no auth at all?

I'm thinking we could do this similarly as for mongodb.
I.e. kustomize secretgenerator from a file, which contains some `changeme` passwords, which we encourage users to change as they like.

MarioUhrik (changes requested):

manifests/claudie/dynamo/job.yaml
jaskeerat789: AWS cli needs some cred to run the create table command, but since its a locally deployed dynamodb, it does not have any auth. So i need to pass some random value to aws cli for access key and secret key using env vars just to pass the validation check.

Makefile
jaskeerat789: `docker-compose` has be deprecated and the project has moved under `docker` tool. More details can be found [here](https://stackoverflow.com/questions/66514436/difference-between-docker-compose-and-docker-compose)

manifests/claudie/dynamo/job.yaml
jaskeerat789: As suspected, i was using wrong syntax,`{}` instead of `()`. Thanks for guiding me here.

Makefile
MarioUhrik: Shouldn't `dockerUp:` be updated too ?

manifests/claudie/dynamo/job.yaml
MarioUhrik: ok, sounds fine

manifests/claudie/dynamo/dynamodb.yaml
jaskeerat789: After doing some rough calculations, we can have more than 2M clusters in 1Gi of storage.

We create two string items with each `terrafom init` command(LockID and Digest), The combined size of both the string field is 200 bytes and since a cluster with LB have 2 `terraform init` executions (one for k8s nodepools and other for LB nodepools) so that make the 400 bytes of data for a cluster with  LB attached. Adding a buffer of 100 bytes for each cluster brings up the total to 500 bytes or half a kilobyte. Since there are a Million KB in one GB, we can have 2 Million entries in the table

MarioUhrik: Thanks for the info :)
I think 1Gi is fine. It's near-nothing. The sadder thing is that some storageclasses have a minimum permissible capacity per PV much higher than that, e.g. I've seen OCI block volumes have a minimum of 50Gi size, even if you only request 1Gi.

MarioUhrik (approved):
LGTM, great work @jaskeerat789

PR #354, opened 2022-10-28
Feat/subnet per nodepool

### LINKED ISSUE #310: Feature: Create subnet per region
### Motivation:

While working on AWS support, I found out that VMs from region A cannot join a subnet in region B. Therefore, currently, we only support one region for the node pools in AWS, (which might be true for other providers as well). If we want to support multiple regions per provider, we need to create a subnet based on the region and assign it to the particular node pool.

### Description:

Subnets should be dynamically created per region so that we have 1 VCN and N subnets for N regions where we have current node pools. Node pools should be also assigned to the correct subnet.

### Exit criteria:
- [ ] Create an algorithm which will identify used regions and creates one subnet per region
- [ ] Assign node pools to the correct subnet

MiroslavRepka:
This PR changes the templates so every node pool has its own subnet in the VPC. 

Closes #310


jaskeerat789 (commented):
LGTM, It was interesting to see that you can call functions inside gotemplate. Nice work.

jaskeerat789 (approved):

internal/templateUtils/template_test.go
```diff
@@ -0,0 +1,8 @@
+package templateUtils
+
+import "testing"
+
+func TestGetCIDR(t *testing.T) {
+	cidr := GetCIDR("10.0.0.0/16", 2, 1)
+	t.Log(cidr)
+}
```
Despire, line 8: Do we even need a test that will never fail?

MiroslavRepka: This test was used to check if CIDR changes without running it in the template. We can remove it at this point, if you wish.

Despire: actually lets keep it, I don't mind maybe it might catch some errors in future

PR #356, opened 2022-10-28
Add readme under `/manifest`

jaskeerat789:
# description
Update docs to add clear instructions on how to deploy claudie. 

# tasks
- [x] add testing framework warning


MarioUhrik:
Hi guys, just a quick point from me.
Let's try not to duplicate information, if we can help it.
Lots of this information is already present elsewhere, e.g. the deployment command is in `README.md`.

jaskeerat789:
> uys, just a quick point from me.
> Let's try not to duplicate information, if we can help it.
> Lots of this information is already present elsewhere, e.g. the deployment command is in `README.md`.

hey @MarioUhrik thanks for pointing out duplicate information. So should only remove the deployment command from this docs  or remove it from the root README.md. I don't think we should remove it from root README. Also do you have any further comments on the doc here.

MarioUhrik:
If the point of this PR is to prevent people from deploying `testing-framework`, then I think the content of `manifests/README.md` shouldn't be much more than:
```
Warning: you probably don't want to deploy `testing-framework`. 
Please only deploy the manifests from the `claudie` folder.
```

I don't have much of an opinion on the rest of the content.
Except that if we want something like that, e.g. docs for devs that tell them how to get started, then they should be under `/docs` with the rest of the docs.

docs/input-manifest/example.yaml
```diff
@@ -59,7 +59,7 @@ providers:
       compartment_ocid: ocid2.compartment.oc2..aaaaaaaaa2rsfvlvxc34o060kfdygsds21nske76ksjkko21lpsdfsf
   aws:
     - name: aws-1
-      accces_key: SLDUTKSHFDMSJKDIALASSD
+      access_key: SLDUTKSHFDMSJKDIALASSD
```
MarioUhrik, line 62: This is now already fixed in master

manifests/README.md
```diff
@@ -0,0 +1,3 @@
+## Warning 
+Please make sure you are in `claudie` directory before using `kustomize` to deploy Claudie.
+Running `kustomize` in this directory will deploy `testing-framework` along with the other Claudie components.
```
MarioUhrik, line 3: I like this, but could it include the information that `testing-framework` isn't a real part of Claudie, and that the user probably doesn't want to deploy it?

Also, what do you think about putting this exact content also into `manifests/testing-framework` ?

README.md
```diff
@@ -37,6 +37,8 @@ Deploy Claudie Kubernetes [manifests](https://github.com/Berops/claudie/tree/mas
 kustomize build | kubectl apply -f -
```
MarioUhrik, line 37: I think this would be nice too:

Replacing
```
Deploy Claudie Kubernetes [manifests](https://github.com/Berops/claudie/tree/master/manifests/claudie) into a Kubernetes cluster.
```
by
```
Deploy Claudie Kubernetes [manifests/claudie](https://github.com/Berops/claudie/tree/master/manifests/claudie) into a Kubernetes cluster.
```

manifests/README.md
jaskeerat789: > I like this, but could it include the information that testing-framework isn't a real part of Claudie, and that the user probably doesn't want to deploy it?
Cool, I'll update it.

> Also, what do you think about putting this exact content also into manifests/testing-framework ?
yeah we can add it imo.

MarioUhrik (approved):

PR #359, opened 2022-10-28
Add smart pipelines

MiroslavRepka:
This PR adds a mechanism to pipelines which will check whether they need to run E2E tests or not.

Closes #221 #348

Issue #348, opened 2022-10-27
Feature: Dont run the E2E unnecessarily in CD


MiroslavRepka:
### Motivation:

Now that we are polishing up the docs, it would be nice if E2E testing would be skipped in the CD, as they are already skipped in the CI when nothing functionally important changed.

### Description:

The CD should not run E2E tests if only the non-functional part of the repo was changed. (Makefile, README.md, /docs, etc.)

### Exit criteria:
- [ ] Improve the skipping logic in the CI
- [ ] Add the skipping logic to the CD

PR #360, opened 2022-10-28
Improve pipeline logic

### LINKED ISSUE #221: Bug: CI doesn't run if only manifests are changed
### Current Behavior:
If only K8s manifests are changed, the CI pipeline doesn't run the `deploy-and-test` step.

### Expected Behavior:
CI pipeline should run in full if any of the K8s manifests are changed.

### Anything else to note:
This is more of a big deal than it looks.
We've already introduced a handful of bugs, because we falsely assumed everything is good just because of the :heavy_check_mark:  symbol of the pipeline (which skipped the important steps :smile: )

MiroslavRepka:
This PR adds a mechanism to pipelines which will check whether they need to run E2E tests or not.

Closes #221 #348

Tested in #359

### DIFF
--- a/.github/workflows/CD-pipeline-dev.yml
+++ b/.github/workflows/CD-pipeline-dev.yml
@@ -27,15 +27,15 @@ jobs:
           git fetch origin master
           for i in "$(git diff  --name-only HEAD^ HEAD)"
           do
-            if [[ $i == *"internal/"* ]] || [[ $i == *"proto/"* ]] || [[ $i == "go.mod" ]] || [[ $i == "go.sum" ]] && [[ $i != *".md"*]]; then
+            if [[ $i == *"internal/"* ]] || [[ $i == *"proto/"* ]] || [[ $i == "go.mod" ]] || [[ $i == "go.sum" ]] && [[ $i != *".md"* ]]; then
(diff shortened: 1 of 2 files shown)


.github/workflows/CD-pipeline-dev.yml
```diff
@@ -13,8 +13,41 @@ env:
...
+                break
+                fi
+              done
+            fi
+          done
+    outputs:
+      RUN_TESTS: ${{ steps.change.outputs.RUN_TESTS }}
+  #--------------------------------------------------------------------------------------------------
   deploy-and-monitor:
-    if: github.event.pull_request.draft == false
```
MarioUhrik, line 17: Hi @MiroslavRepka , I see we're getting rid of this, but I don't see a "replacement" anywhere, is it intended ?

MiroslavRepka: Yes, I deleted it as it was completely unnecessary condition. The CD is triggered on `push` to `master`, thus this event does not even exist as the PR is closed at that point.

MarioUhrik (approved):
Superficially LGTM :D

As long as it appears to work, let's roll with it, and fix it if any bugs appear

PR #367, opened 2022-11-01
Remove default http server conf from nginx

### LINKED ISSUE #364: Remove Nginx default HTTP server on LB nodes
### Current Behavior:
Nginx comes with default http server on port 80. If the user tries to create a LB role that balances traffic on port 80, they nginx will fail to run due to the conflict. 
### Expected Behavior:
Nginx should only run the `proxy` server. This can be achieved by deleting the  `/etc/nginx/sites-available` and `/etc/nginx/sites-enabled` files.  We also want to update the `nginx.conf` to only enable proxy mode.

jaskeerat789:
# Description
Fixes #364 
We are deleting default http server conf files in order to prevent spinning up the default http server on port `80`. We only want to use proxy sever from nginx.

### DIFF
--- a/manifests/claudie/kustomization.yaml
+++ b/manifests/claudie/kustomization.yaml
@@ -21,7 +21,7 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: claudieio/ansibler
-  newTag: 2962f5f-918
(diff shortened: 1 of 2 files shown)


Despire (approved):

Despire:
Thx, for debugging this issue @jaskeerat789

PR #368, opened 2022-11-02
Increase CPU of context-box and mongodb to help prevent initial crashlooping

MarioUhrik:
We've had some reports of context-box crashlooping forever during startup.
I believe this is a very cheap and quick way to help prevent this.
Jas will open a more proper issue that'll address resource requests/limits of all the pods with more closer attention.

Note that we can check resource usage metrics in the cloud console.

### DIFF
--- a/manifests/claudie/context-box.yaml
+++ b/manifests/claudie/context-box.yaml
@@ -21,10 +21,10 @@ spec:
           image: claudieio/context-box
           resources:
             requests:
-              cpu: 5m
+              cpu: 40m
               memory: 50Mi
             limits:
-              cpu: 25m
+              cpu: 80m
               memory: 100Mi
           env:
             - name: DATABASE_PORT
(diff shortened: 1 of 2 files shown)


jaskeerat789 (approved):
Thanks for a quick PR @MarioUhrik

MiroslavRepka (approved):

MarioUhrik:
Hi @MiroslavRepka, unfortunately no, we have no proof this'll get fix it forever.
I just think this is a good thing to do whether it truly helps or not.
And the chances are good :)

Despire:
I'm just wondering how to get the right numbers for CPU limits. I've just watched the current CI pipeline with kubectl top and context-box doesn't need that much of CPU 
Screenshot 2022-11-02 at 10 00 47

Despire (approved):

PR #371, opened 2022-11-02
Feat: LB create/update/delete

### LINKED ISSUE #307: Remove LB cluster for a k8s cluster
## Description:
While working on #279, I had faced a scenario where the user would delete the LB cluster for k8s cluster. This scenario was not mentioned in the original.  The removal of LB won't be a simple task since we will have to change the API endpoint for the cluster and also open firewall ports for nodes (according to #306 ). 

## Exit criteria:
- [x] delete LB infra only 
- [x] change API Endpoint via ansibler
- [ ] change firewall policies for the nodes to expose ports

Despire:
Closes #307 

Currently the ApiEndpoint for the Load-Balancers wasn't working correctly. When Attaching/Dettaching/Changing roles the endpoint wouldn't be correctly updated. This PR fixes the described issue by correctly identifying the cases when an ApiEndpoint needs to be updated and executing the UpdateApiEndpoint function.

I've tested this out with the following scenarios:
1. start with a k8s-cluster with no LB
2. attach ApiServer LB
3. attach another LB (not for ApiServer)
4. Swap roles of the Load-Balancers
5. Destroy the ApiServer LB
6. Change role for the other LB to be the ApiServer
7. Destroy the ApiServer LB

8. Change the role from the APIServer to some other role (edited)

In all the scenarios the endpoint was correctly updated

This PR also includes manifest validation that only allows a single ApiServer LB per k8s cluster.


services/ansibler/server/loadbalancers.go
```diff
@@ -40,106 +40,316 @@ const (
...
+			}
+
+			return os.RemoveAll(k8sDirectory)
+		})
+	}
+
+	return errGroupK8s.Wait()
+}
+
+func HandleAPIEndpointChange(apiServer *LBData, k8sCluster *LBInfo, k8sDirectory string) error {
```
MiroslavRepka, line 238: This function should probably not be exported ~`HandleAPIEndpointChange`~ -> `handleAPIEndpointChange`

services/ansibler/server/loadbalancers.go
```diff
@@ -40,106 +40,316 @@ const (
...
 
-//setUpLoadbalancers will set up and verify the loadbalancer configuration including DNS
-//returns error if not successful, nil otherwise
+// setUpLoadbalancers will set up and verify the loadbalancer configuration including DNS
+// returns error if not successful, nil otherwise
 func setUpLoadbalancers(lbInfos map[string]*LBInfo) error {
 	var errGroupK8s errgroup.Group
 	//iterate through k8s clusters LBs
 	for k8sClusterName, lbInfo := range lbInfos {
-		func(lbInfo *LBInfo, k8sClusterName string) {
```
MiroslavRepka, line 73: What is the reason to change the function around the error group to `var := var` ?

services/ansibler/server/loadbalancers.go
```diff
@@ -40,106 +40,316 @@ const (
...
-	log.Info().Msgf("Changing the API endpoint for the cluster %s", lbInfo.LbCluster.ClusterInfo.Name)
-	log.Info().Msgf("New endpoint is %s", lbInfo.LbCluster.Dns.Endpoint)
-	flag := fmt.Sprintf("--extra-vars \"NewEndpoint=%s OldEndpoint=%s\"", lbInfo.LbCluster.Dns.Endpoint, lbInfo.CurrentDNS.Endpoint)
-	ansible := ansible.Ansible{Playbook: apiChangePlaybook, Inventory: inventoryFile, Flags: flag, Directory: directory}
-	err := ansible.RunAnsiblePlaybook(fmt.Sprintf("EP - %s", lbInfo.LbCluster.ClusterInfo.Name))
-	if err != nil {
-		return fmt.Errorf("error while running ansible for EP change for %s : %w ", lbInfo.LbCluster.ClusterInfo.Name, err)
+// FindAPIEndpointNode searches the NodePools for a Node with type ApiEndpoint.
+// If no such node is found an error is returned.
+func FindAPIEndpointNode(nodepools []*pb.NodePool) (*pb.Node, error) {
```
MiroslavRepka, line 321: ditto,  should not be exported

MiroslavRepka, line 321: I can never decide what is better in these types of functions; return `nil` or return error :smile:.

What do you think @jaskeerat789?

services/ansibler/server/loadbalancers.go
```diff
@@ -40,106 +40,316 @@ const (
...
+		return nil
+	case EndpointRenamed:
+		oldEndpoint = apiServer.CurrentLbCluster.Dns.Endpoint
+		newEndpoint = apiServer.DesiredLbCluster.Dns.Endpoint
+	case RoleChanged:
+		newEndpoint = apiServer.DesiredLbCluster.Dns.Endpoint
+		// 1st check if any other LB was previously an ApiServer.
+		var oldAPIServer *LBData
+
+		for _, lb := range k8sCluster.LbClusters {
```
MiroslavRepka, line 259: Can we put this logic into separate function for better readability, i.e. `func findApiLB()`?

MiroslavRepka (commented):
LGTM @Despire, kudos to you for a very nice implementation of this issue. I am really impressed. 

Would it make sense to add some to the E2E tests? @jaskeerat789

internal/manifest/validate_load_balancer.go
```diff
@@ -21,21 +24,30 @@ func (l *LoadBalancer) Validate(m *Manifest) error {
...
+			apiServerRole = role.Name
+		}
+
 		if _, ok := roles[role.Name]; ok {
 			return fmt.Errorf("name %q is used across multiple roles, must be unique", role.Name)
 		}
 		roles[role.Name] = true
 	}
 
+	apiServerLBExists := make(map[string]bool) // [Targetk8sClusterName]bool
```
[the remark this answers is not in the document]
Despire: It's used [here](https://github.com/Berops/claudie/blob/5eb96bf827b719ee78c68f793b57afc3859291b2/internal/manifest/validate_load_balancer.go#L70), to not make the if more verbose than it needs to be by checking if `if _, ok := ....`

internal/manifest/validate_load_balancer.go
```diff
@@ -53,6 +65,16 @@ func (l *LoadBalancer) Validate(m *Manifest) error {
 			if _, ok := roles[role]; !ok {
 				return fmt.Errorf("role %q used inside cluster %q is not defined", role, cluster.Name)
 			}
+
+			// check if this is an ApiServer LB and another ApiServer LB already exists.
+			if role == apiServerRole && apiServerLBExists[cluster.TargetedK8s] {
```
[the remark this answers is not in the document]
Despire: Yes, actually we can do even better
```go
			if role == apiServerRole {
				// check if this is an ApiServer LB and another ApiServer LB already exists.
				if apiServerLBExists[cluster.TargetedK8s] {
					return fmt.Errorf(...)
				}

				// this is the first LB that uses the ApiServer role.
				apiServerLBExists[cluster.TargetedK8s] = true
			}
```

services/ansibler/server/loadbalancers.go
Despire: started with the `func(a, b)(c,d)` wrapping as it was at the start, each time I looked at the function took me a while to to read the code due to the nested function creation and indentation, I just created local variables instead which has the same effect as passing it as arguments to the function wrapper, tbh I don't really care about which to choose I can revert it back to the `func(a,b)(c,d)` wrapper.

services/ansibler/server/loadbalancers.go
Despire: Sure

services/ansibler/server/loadbalancers.go
MiroslavRepka: We used a wrapper every time we needed this logic, thus for consistency's sake, I would choose one and roll with it. I don't mind either.

Despire: 👍  I'll revert it back to the function wrapper since its used in other places

Despire:
I've implemented the suggestion from #385 , further I've also identified one more case where the endpoint could change which I didn't though of before, I've tested the same scenarios described in the header of this issue and they still work with these changes.
Please re-review and let me know what you think @jaskeerat789, @MiroslavRepka

internal/manifest/validate_load_balancer.go
```diff
@@ -53,6 +65,16 @@ func (l *LoadBalancer) Validate(m *Manifest) error {
 			if _, ok := roles[role]; !ok {
 				return fmt.Errorf("role %q used inside cluster %q is not defined", role, cluster.Name)
 			}
+
+			if role == apiServerRole {
+				// check if this is an ApiServer LB and another ApiServer LB already exists.
+				if apiServerLBExists[cluster.TargetedK8s] {
+					return fmt.Errorf("role %q is used across multiple Load-Balancers for k8s-cluster %s. Can have only one ApiServer Load-Balancer per k8s-cluster", role, cluster.TargetedK8s)
```
jaskeerat789, line 72: @Despire Can we make this clear in the docs as well? we should let users know about this limitation

jaskeerat789 (approved):
Good work with the `oldAPIEndpoints` @Despire.

jaskeerat789:
For E2E, we can have a new issue, do you plan to add the tests in this PR or create a new one ?

Despire:
> For E2E, we can have a new issue, do you plan to add the tests in this PR or create a new one ?

I would prefer a new PR,

internal/manifest/validate_load_balancer.go
Despire: Yes.

PR #377, opened 2022-11-02
Add fix for CI/CD

MiroslavRepka:
This PR fixes issues reported in #371. For some reason, when adding `[[ $i != *".md"* ]]` the bash interpreted `$i` as the whole array. I changed the condition and array creation, which was tested on #376 and should be fixed now.

### DIFF
--- a/.github/workflows/CD-pipeline-dev.yml
+++ b/.github/workflows/CD-pipeline-dev.yml
@@ -26,19 +26,21 @@ jobs:
         run: |
           SERVICES=( ${{ env.SERVICES }} )
           git fetch origin master
-          echo "Files changed since last commit"
-          git diff  --name-only HEAD^ HEAD
-          for i in "$(git diff  --name-only HEAD^ HEAD)"
+          changes=($(git diff  --name-only HEAD^ HEAD))
+          for i in "${changes[@]}"
           do
-            if [[ $i == *"internal/"* ]] || [[ $i == *"proto/"* ]] || [[ $i == "go.mod" ]] || [[ $i == "go.sum" ]] && [[ $i != *".md"* ]]; then
+            if [[ $i == *".md"* ]]; then
+              echo "Skipping document $i"
+              continue
+            elif [[ $i == *"internal/"* ]] || [[ $i == *"proto/"* ]] || [[ $i == "go.mod" ]] || [[ $i == "go.sum" ]]; then
               echo "RUN_TESTS=true" >> $GITHUB_OUTPUT
               break
-            elif [[ $i == *"manifests/"* ]] && [[ $i != *".md"* ]]; then
+            elif [[ $i == *"manifests/"* ]]; then
               echo "RUN_TESTS=true" >> $GITHUB_OUTPUT
               break
             else
               for SERVICE in "${SERVICES[@]}"; do
-                if [[ $i == *"services/$SERVICE"* ]] && [[ $i != *".md"* ]]; then
+                if [[ $i == *"services/$SERVICE"* ]]; then
                 echo "RUN_TESTS=true" >> $GITHUB_OUTPUT
                 break
                 fi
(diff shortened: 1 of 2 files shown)


.github/workflows/CD-pipeline-dev.yml
```diff
@@ -26,19 +26,21 @@ jobs:
...
           SERVICES=( ${{ env.SERVICES }} )
           git fetch origin master
-          echo "Files changed since last commit"
-          git diff  --name-only HEAD^ HEAD
-          for i in "$(git diff  --name-only HEAD^ HEAD)"
+          changes=($(git diff  --name-only HEAD^ HEAD))
+          for i in "${changes[@]}"
           do
-            if [[ $i == *"internal/"* ]] || [[ $i == *"proto/"* ]] || [[ $i == "go.mod" ]] || [[ $i == "go.sum" ]] && [[ $i != *".md"* ]]; then
+            if [[ $i == *".md"* ]]; then
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 32: I don't mean to pick or anything; could this perhaps be better future-proofed by quoting variables ("$stuff") to prevent possible word splitting scenarios? not sure if it's sufficient that the array being iterated over is quoted, but IMO it wouldn't hurt to quote all variables for bash..

bernardhalas: I think you wanted to do regexp comparison here, didn't you? By no means showing off here, but as we have issues with this CI, let's make it a bit more robust using the regexes:
```
if [[ "${i}" =~ .*\.md.* ]]; then
  ...
elif [[ "${i}" =~ .*(internal|proto|manifests)/.* ]] || [[ "${i}" =~ go\.(mod|sum) ]]; then
...
else
  for SERVICE in "${SERVICES[@]}"; do
    if [[ "${i}" =~ .*services/${SERVICE}.* ]]; then
...
```
Also, I've combined lines 35 and 38. You can take similar approach with the regexes on line

MiroslavRepka: By no means, please leave any feedback you think will add value. FIY @wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf 

Tbf, my regex skills are quite poor, so thanks for the examples/improvements, will definitely add it.

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf (approved):
looks good to me! :rocket:
thanks for implementing the little suggestion, @MiroslavRepka and for the regex help @bernardhalas.

.github/workflows/CD-pipeline-dev.yml
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf: looks fine to me now;

it'd be great if we could test the CI config with some kind of unit tests... :thinking: :laughing:

.github/workflows/CI-pipeline.yml
```diff
@@ -44,18 +44,22 @@ jobs:
...
+          changes=($(git diff  --name-only origin/master))
+          for i in "${changes[@]}"
           do
-            if [[ $i == *"internal/"* ]] || [[ $i == *"proto/"* ]] || [[ $i == "go.mod" ]] || [[ $i == "go.sum" ]] && [[ $i != *".md"* ]]; then
+            if [[ "${i}" =~ .*\.md.* ]]; then
+              echo "Skipping document "$i""
+              continue
+            elif [[ "${i}" =~ .*(internal|proto)/.* ]] || [[ "${i}" =~ go\.(mod|sum) ]]; then
               arr=(${SERVICES[@]})
               echo "All services needs to be built"
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 55: while we're at it, the following could be fixed, too:
* typo/grammar: should be "All services *need* to be built.".
* also L41, should instead read "Check the repository for changes".
* also L24 of `CD-pipeline-dev.yml`, should instead read "Check the PR for changes since the latest commit.".

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf (commented):
looks great :rocket: 
just added a couple of typo fixing suggestions, while we're dealing with these files.

Despire (approved):

PR #391, opened 2022-11-07
Build multi-arch container images in CI

### LINKED ISSUE #366: Build multi-arch container images in CI
### Motivation:
After having a discussion in the Slack channel, the team was convinced that we want to support `arm64` arch along with currently supported `amd64` arch. This makes sense since the popularity of arm based personal machine are becoming popular and the end users might want to run claudie locally on these arm64 machines. 
We can use the official docker plugin [`buildx`](https://github.com/docker/buildx) and pass list of archs that we want to support.

### Exit criteria:
- [ ] build multi-arch images in CI

jaskeerat789:
# Description
Resolved #366 by using `buildx`

### DIFF
--- a/internal/DELETE.ME
+++ b/internal/DELETE.ME
@@ -1,2 +1,0 @@
-// DELETE ME
-// added to trigger docker build for all services
--- a/internal/DELETE.ME
+++ b/internal/DELETE.ME
@@ -1,0 +1,2 @@
+// DELETE ME
+// added to trigger docker build for all services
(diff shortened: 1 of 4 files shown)


jaskeerat789:
i was successfully able to build and push multi arch images 
image

Despire:
The images are being build for over an hour already, compared to previously ~3-4min that's a 10x slowdown, is there a chance to speed this up?

MarioUhrik:
optimization would be nice, but i don't see it as necessary, LGTM

MarioUhrik (approved):

internal/DELETE.ME
```diff
@@ -0,0 +1,2 @@
+// DELETE ME
+// added to trigger docker build for all services
```
MarioUhrik, line 2: let's not forget about this :D

jaskeerat789: yeah i am just waiting for containers to spin up once in order to test if everything is fine with them. I'll revert the commit after  necessary checks have been made

PR #392, opened 2022-11-07
Feat/acc net az

### LINKED ISSUE #337: Feature: Enable `accelerated networking` on Azure VMs which support it
### Motivation:

As described in the discussion [here](https://github.com/Berops/claudie/pull/334#issuecomment-1286738922) the performance of Azure VMs is rather poor. One of the reasons might be the networking bandwidth. We might want to try to enable [`accelerated networking`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/network_interface#enable_accelerated_networking) on the VM types that support it, which should increase the bandwidth and boost the performance.

### Description:

Enable `accelerated networking` on Azure VMs, which support this feature.

### Exit criteria:
- [ ] Function to decide whether VM type supports `accelerated networking` is implemented
- [ ] The Azure templates will enable `accelerated networking` for VMs that support it

### Example template 

```hcl
resource "azurerm_network_interface" "{{ $nodepool.Name }}-{{ $clusterHash }}-ni" {
  ...
  enable_accelerated_networking = {{ enableAccNet $nodepool.ServerType }}
  ...
}
```

MiroslavRepka:
This PR enables accelerated networking on the VMs in Azure which supports it as described in #337.

The function which checks if the accelerated networking can be enabled uses regex to match the VM sizes to the list of supported ones.

I omitted regions and OS checks since OS support is for anything newer than Ubuntu 16.04 and all regions are supported.

Closes #337


internal/utils/utils.go
```diff
@@ -55,3 +63,23 @@ func ProtocolNameToAzureProtocolString(protocol string) string {
 func AssignPriority(index int) int {
 	return basePriority + index
 }
+
+// EnableAccNet will check if accelerated networking can be enabled based on conditions
+// specified here https://azure.microsoft.com/en-us/updates/accelerated-networking-in-expanded-preview/
+// we will look only at VM sizes, since all regions are supported now all reasonable operating systems
+func EnableAccNet(vmSize string) string {
+	if !checkContains(vmSizes, vmSize) {
```
Despire, line 71: This can be just `return checkContains(vmSizes, vmSize)`

Despire: Nvm, I just realized it's a string.

Despire (approved):
LGTM

jaskeerat789 (approved):

PR #405, opened 2022-11-08
Feat/fix skip manifest on validation

### LINKED ISSUE #402: Bug: infra gets deleted after config is reapplied with errors
### Current Behavior:

After applying a config successfully  and making a modification to it that doesn't pass through the validation stage in the frontend service the infra of the previously created config gets deleted
### Expected Behavior:

A deletion of the infra shouldn't happen, if new changes to the config are invalid the old infra shouldn't be changed
### Steps To Reproduce:

1. apply config
2. make a change that will fail on the validation stage in the frontend service
3. See terraformer logs for deletion of the infra

Despire:
Closes #402


jaskeerat789 (approved):
Thanks for the quick fix @Despire

PR #411, opened 2022-11-11
Fix example inputManifest AMI

MarioUhrik:
It was ARM64, which is clearly wrong.
The new link is AMD64, the same as in E2E pipelines.


MiroslavRepka (approved):
LGTM, thanks for the fix @MarioUhrik

PR #418, opened 2022-11-11
Create own ansible base image

MiroslavRepka:
This PR creates our own ansible base image for `ansibler`, due to issues seen in recent pipeline runs.

### DIFF
--- a/manifests/claudie/kustomization.yaml
+++ b/manifests/claudie/kustomization.yaml
@@ -21,7 +21,7 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: claudieio/ansibler
-  newTag: 561b5f8-1045
+  newTag: 22560da-1053
 - name: claudieio/builder
   newTag: 561b5f8-1045
 - name: claudieio/context-box
--- a/services/ansibler/Dockerfile
+++ b/services/ansibler/Dockerfile
@@ -1,3 +1,45 @@
+#Ansible base image inspired by https://github.com/cytopia/docker-ansible
+FROM alpine:3.16 as ansible-base
+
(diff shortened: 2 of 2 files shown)


MarioUhrik:
Guys, is there truly no `ansible` image out there that we can use ?
Nothing from here?
https://hub.docker.com/r/ansible/ansible/tags
I see those aren't `arm64` compatible, so fair enough.
But these still are, aren't they?
https://hub.docker.com/r/cytopia/ansible/tags

It sounds absurd to me that we have to create our own Ansible image.
But we can do it if there's no other way, of course.

Despire:
Since this Dockefile is inspired by https://github.com/cytopia/docker-ansible, can't we just use it as the base image ? Like Mario suggested.

MiroslavRepka:
Hi @MarioUhrik 
>https://hub.docker.com/r/ansible/ansible/tags 

The last time this was updated was 5 years ago

>https://hub.docker.com/r/cytopia/ansible/tags

They merge the support for `arm64` [2 days ago](https://github.com/cytopia/docker-ansible/pull/111) FYI @Despire, the reason for this PR was missing support for arm

MarioUhrik:
> The last time this was updated was 5 years ago
You are right, let's not use 5 year old stuff, sorry for suggesting that, I didn't notice

Well, if the issues with `cytopia` are now gone, I'd vote to return to the previous approach :D

MiroslavRepka:
>Well, if the issues with cytopia are now gone, I'd vote to return to the previous approach :D

I am rerunning the CI which failed because of a missing arm image. If that step will succeed now, we can close this.

MiroslavRepka:
Closing due to recent addition of  cytopia's of arm-based images

Issue #370, opened 2022-11-02
DesiredState and CurrentState does not match after cluster creation


jaskeerat789:
### Current Behavior:
After a successful claudie run, if you try to retrieve config from DB using `printConfig` function in [`/services/context-box/client/client.go`](https://github.com/Berops/claudie/blob/master/services/context-box/client/client_test.go#L118) but it will not print any info related to nodes. After attaching a debugger to this function, you could see that `DesiredState` have empty lists for `Nodes` under `NodePools` whereas current state has an non-empty list of `Nodes`. I am attaching screenshots for better representation. 

Desired State            |  Current State
-------------------------|-------------------------
<img src="https://user-images.githubusercontent.com/38110126/199462159-bf74b077-f7ed-4b2b-8492-6f864b865f10.png" width="400"/> | <img src="https://user-images.githubusercontent.com/38110126/199462189-4ec7d6fa-48bf-4410-85c4-f052c8cac3de.png" width="400"/>

PR #422, opened 2022-11-14
Feat/cs ds equality

MiroslavRepka:
This PR fixes the issue described in #370 

The problem was in the `context-box` since it did not update the Desired State after `builder` finished processing. This lead to the difference between these two.

This PR also improves the `printConfig()` function, with a small test to check if the prints are equal for both states.


Despire (approved):
LGTM! Thanks for the fix @MiroslavRepka

MiroslavRepka:
@Despire could you please check my latest commits before we merge it?

Despire:
why is the check `if dbConf.DesiredState != nil {`  needed ? shouldn't the config from builder also have desired state set to nil ?

MiroslavRepka:
Thanks for feedback @Despire :) 
>why is the check if dbConf.DesiredState != nil { needed ? shouldn't the config from builder also have desired state set to nil ?

Image a case, where Builder will start building the infra, and while Builder is working, a request for deletion comes. 

The deletion request will set `DesiredState` to `nil` and would wait until the Builder finishes. 

When the Builder is finished we need to save the current state as well as desired state, since Builder updates both of them. So if we would not check, if the config in a database does not have a `DesiredState` set to `nil`, the `SaveConfigBuilder` would overwrite it with the desired state from before the deletion request, resulting in a `DesiredState` not being `nil`, even though it should have been because of the deletion request.

Despire:
thx for the explanation @MiroslavRepka, didn't considered the deletion at all :), 
the changes looks good 👍

Issue #353, opened 2022-10-28
Feature: Move `region` field from nodepool to provider spec


MiroslavRepka:
### Motivation:

Most of the providers define the `region` used on the provider level in terraform, only sometimes on the resource level. 
[GCP](https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/provider_reference#example-usage---basic-provider-blocks), [OCI](https://docs.oracle.com/en-us/iaas/Content/API/SDKDocs/terraformproviderconfiguration.htm#APIKeyAuth), [Azure](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/resource_group#location) (under RG spec), [AWS](https://registry.terraform.io/providers/hashicorp/aws/latest/docs#region), Hetzner does not even use `region`, only `zone`

However, in our input manifest, the user is defining the region in the node pool spec, so every node pool has to define it, regardless if it is used or not since we usually look at the `region` of the first node pool. 

I propose to move `region` to provider spec and leave `zone` in node pool spec. If the user would like to use a different region, they would add a provider into their spec, with the same credentials, only a different `region`, as they would have to do it in terraform anyway.

Example:
```yaml
providers:
  aws:
    - name: aws-1
      #some creds
      region: "eu-central-1"
    aws:
    - name: aws-2
      #same creds as in aws-1
      region: "eu-central-2"
  azure:
    - name: azure-1
      ...
      region: "Europe West"
  gcp: 
    - name: gcp-1
      ...
      region: "eu-central1"
  oci:
    - name: oci-1
      ...
      region: "eu-frankfurt-1"

nodePools:
   dynamic:
    - name: pool-1
      providerSpec:
        name: aws-1
        zone: eu-central-1c
      ...
    - name: pool-2
      providerSpec:
        name: oci-1
        zone: hsVQ:EU-FRANKFURT-1-AD-1
      ...
``` 

### Description:

The `region` field should move from node pool spec to provider spec.
### Exit criteria:
- [ ] Change the parsing of the manifest in the frontend/scheduler appropriately
- [ ] Change config proto spec to accommodate the change
- [ ] Change template to accommodate the change

bernardhalas:
I know this makes sense from the TF perspective (as one of the things I assume it affects the URL of the cloud provider API endpoint to which TF is talking to). I'm not convinced this makes sense from the end-user perspective.

[...]

PR #429, opened 2022-11-18
Clarify docs

MiroslavRepka:
This PR clarifies the docs about permissions accounts need to function correctly. Azure is still missing due to the rework in #353.


MarioUhrik:
Beautifully done @MiroslavRepka.
No negative comments.

I'm just wondering, would it be better or worse to put the info into `input-manifest.md` under the sections of individual providers, instead of starting new files ?
I wouldn't want to bloat the folder. But I wouldn't want to bloat the one file, either.

MiroslavRepka:
Thanks for the feedback.

>I'm just wondering, would it be better or worse to put the info into input-manifest.md under the sections of individual providers, instead of starting new files ?

I was thinking the same, however, I wasn't sure much footprint will description of each provider take.

We can put them under the subdirectory, i.e. `docs/input-manifest/providers/<provider>.md`. Would that solve the issue?

MarioUhrik:
> We can put them under the subdirectory, i.e. docs/input-manifest/providers/<provider>.md. Would that solve the issue?

Yes please :)

MarioUhrik (approved):

PR #430, opened 2022-11-18
Jaskeerat789/issue353

### LINKED ISSUE #353: Feature: Move `region` field from nodepool to provider spec

jaskeerat789:
# Description
Fixes #353 by using mukltiple provider definition in terrform scripts. Also updated the test-set to create multi-region clusters.


internal/utils/cluster.go
```diff
@@ -40,6 +40,24 @@ func GetNodePoolByName(nodePoolName string, nodePools []*pb.NodePool) *pb.NodePo
...
+		regionSet[nodepool.Region] = struct{}{}
+	}
+
+	// extract value of set and create a slice
+	regions := make([]string, len(regionSet))
+	i := 0
+	for k := range regionSet {
+		regions[i] = k
+		i++
+	}
```
Despire, line 57: You request the capacity in the slice and then just append to it
```go
regions := make([]string, 0, len(regionSet))
for k := range regionSet {
regions = append(regions, k)
}
```

services/terraformer/server/clusterBuilder/clusterBuilder.go
```diff
@@ -201,6 +208,12 @@ func (c ClusterBuilder) getNodes() []*pb.Node {
 	return oldNodes
 }
 
+// ReplaceAll is a wrapper function to replace all occurrence of a substring with another
+// substring. Returns modified string
+func (nd NodepoolsData) ReplaceAll(s, old, new string) string {
+	return strings.ReplaceAll(s, old, new)
+}
+
```
Despire, line 216: This function should be part of `NodepoolsData` struct, move the function to this [file](https://github.com/Berops/claudie/blob/ef31f9951082f7d8b2f22df8694884306054f583/internal/templateUtils/templates.go#L1) and then add it [here](https://github.com/Berops/claudie/blob/ef31f9951082f7d8b2f22df8694884306054f583/internal/templateUtils/templates.go#L70) to call it from the templates

Despire: Actually I think you can just add something like this to the FuncMap [here](https://github.com/Berops/claudie/blob/ef31f9951082f7d8b2f22df8694884306054f583/internal/templateUtils/templates.go#L71) and you shouldn't even need a function definition.
```go
"ReplaceAll": strings.ReplaceAll
```

jaskeerat789: yeah i was not aware of the function map, is sounds like i dont need to define a special wrapper function, i will pass the `strings.ReplaceAll` in the function map

services/terraformer/templates/aws-lb.tpl
```diff
@@ -1,139 +1,141 @@
...
+{{- $clusterHash := .ClusterHash }}
+{{- $index :=  0 }}
 
+{{- range $i, $region := .Regions }}
 provider "aws" {
-  region     = "{{(index .NodePools 0).Region}}"
-  access_key = "{{(index .NodePools 0).Provider.AwsAccessKey}}"
-  secret_key = file("{{(index .NodePools 0).Provider.SpecName}}")
-  alias = "lb-nodepool"
+  access_key = "{{(index $.NodePools 0).Provider.AwsAccessKey}}"
```
MiroslavRepka, line 7: >$.NodePools

why do we need `$`?

services/terraformer/templates/hetzner.tpl
```diff
@@ -29,7 +29,7 @@ resource "hcloud_firewall" "defaultfirewall" {
     ]
   }
 
-  {{ if index .Metadata "loadBalancers" | targetPorts | isMissing 6443 }}
+  {{- if index .Metadata "loadBalancers" | targetPorts | isMissing 6443 }}
```
MiroslavRepka, line 32: > {{ - if index .Metadata ... }

Why do we need `-` before `if`?

MiroslavRepka (approved):
LGTM, great job @jaskeerat789 !

MiroslavRepka:
I just realized, since we are no longer supplying `resource_group`, we might as well delete it from `config.proto`

services/terraformer/templates/hetzner.tpl
jaskeerat789: It's just for cleaner tf files, `-` does not result in new line creation.

Issue #387, opened 2022-11-04
Clear DynamoDB field after deletion


jaskeerat789:
### Description:
After the cluster is delete, the dynamoDB document is still present in the database. After successful deletion of infra, we should remove the corresponding state lock data from the table too. I was expecting TF to handle this out the box. 

### Exit criteria:
- [ ] Handle dynamodb cleanup after cluster deletion

MarioUhrik:
Hi Jas, why is this a problem? Does this lead to some negative effects?

jaskeerat789:
> Hi Jas, why is this a problem? Does this lead to some negative effects?

According to me knowledge it has not resulted in any problem. I wanted to added this as a good practice

MarioUhrik:
If it's not a problem, I'd actually consider simply not ever cleaning DynamoDB contents

MarioUhrik:
We are clear to proceed with this task, as decided at the meeting

PR #432, opened 2022-11-18
Add removing lockfile from DynamoDB

cloudziu:
Draft PR for #387 issue.

I'm not sure about two things:
- Using five AWS libraries just to make one request seems to be a bit overkill. I'm considering craving a manual request to Dynamo
- Is the approach I took with type assertion correct [in this case](https://github.com/Berops/claudie/blob/14be2910e79ddac9e21644f18f0a19901adece52/services/terraformer/server/server.go#L174)? I wonder if there are another `types` that won't meet the requirements of getting `hash` and `name` from it. 
https://github.com/Berops/claudie/blob/14be2910e79ddac9e21644f18f0a19901adece52/services/terraformer/server/server.go#L175


MiroslavRepka:
Hey @cloudziu, thx for the addition

>Using five AWS libraries just to make one request seems to be a bit overkill. I'm considering craving a manual request to Dynamo

AFAIK, we use only one "library" (module), but a few packages. I would not worry about it too much though.

>Is the approach I took with type assertion correct [in this case](https://github.com/Berops/claudie/blob/14be2910e79ddac9e21644f18f0a19901adece52/services/terraformer/server/server.go#L174)? I wonder if there are another types that won't meet the requirements of getting hash and name from it

You can use `Id()` function from interface `Cluster` to get `<name>-<hash>` string, similarly as we did [here](https://github.com/Berops/claudie/blob/feature/clear-dynamodb-fields/services/terraformer/server/server.go#L172)

services/terraformer/server/server.go
```diff
@@ -151,7 +172,23 @@ func (*server) DestroyInfrastructure(ctx context.Context, req *pb.DestroyInfrast
...
 
-				return mc.RemoveObject(ctx, "claudie-tf-state-files", key, minio.RemoveObjectOptions{
+				// Key under which the lockfile id is stored in dynamodb
+				dynamoLockId, err := attributevalue.Marshal(
+					fmt.Sprintf("%s/%s/%s-md5", minioBucket, projectName, c.Id()),
+				)
+				if err != nil {
+					log.Warn().Msgf("Error composing state lockfile id for cluster %s : %s", c.Id(), config.Name)
+				}
+				fmt.Println(dynamoLockId)
```
MiroslavRepka, line 182: I just noticed this print. Do we need it? If so, can we make it `log.Debug().Msgf("some message")`?

MiroslavRepka (approved):

services/terraformer/server/server.go
cloudziu: This signals when the `Sprintf` output could not be marshalled to DynamoDB request. I think we can make it `debug`

PR #445, opened 2022-11-29
Add diagram of an example infra

bernardhalas:
Outline an example of an infrastructure created by Claudie. A picture is worth more than a paragraph.


MarioUhrik (approved):
This is nice Bernard, thanks :)

PR #446, opened 2022-11-29
Streamline the deployment step

bernardhalas:
`kustomize` can be pointed to any folder in the repo. That makes the `note` superfluous.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Claudie comes with a pre-configured storage solution, with ready-to-use Storage 
 Deploy Claudie Kubernetes [manifests/claudie](https://github.com/Berops/claudie/tree/master/manifests/claudie) into a Kubernetes cluster.
 
 ```
-kustomize build manifests/claudie | kubectl apply -f -
+kubectl apply -k manifests/claudie
 ```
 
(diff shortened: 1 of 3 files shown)


MarioUhrik (approved):
Thanks @bernardhalas , this is great.

I added a couple commits, please check

bernardhalas:
Thank you for complementing me.

bernardhalas:
Btw, @MarioUhrik , would you know if `kubectl apply -k manifests/claudie` would work the same way? Because that would drop the need for a standalone `kustomize` binary.

MarioUhrik:
> Btw, @MarioUhrik , would you know if `kubectl apply -k manifests/claudie` would work the same way? Because that would drop the need for a standalone `kustomize` binary.

Yes, let's do that.
One way or another, you need a relatively recent version of Kustomize - or in the case of your suggestion, Kubectl.
I just tested that:
- kustomize v3 is too old
- kustomize v4 suffices
- kubectl 1.20 is too old
- kubectl 1.25 suffices

Despire (approved):

MarioUhrik (approved):

PR #447, opened 2022-11-30
Increase patience of Context Box when waiting for MongoDB

MarioUhrik:
I think that increasing the liveness probe delay & the number of reconnection attempts will make the startup more robust.
Please let me know your thoughts


Despire (approved):
I dont see any issues with this 👍

jaskeerat789 (approved):
LGTM 👍

Issue #443, opened 2022-11-29
Error while cleaning lock in DynamoDB after infra deletion


jaskeerat789:
### Current Behavior:
After deleting the infra, terraformer tries to delete the `lock` from DynamoDB but it fails with the following error:
>  Failed to destroy the infra for project TestSet-2 : failed to remove state lock file ts2-lb-2-s8u3l79 : operation error DynamoDB: DeleteItem, failed to sign request: failed to retrieve credentials: failed to refresh cached credentials, no EC2 IMDS role found, operation error ec2imds: GetMetadata, http response error StatusCode: 404, request to EC2 IMDS failed 

This error is pretty consistent in CI.

Despire:
closed in #449

PR #449, opened 2022-12-01
Feat/node metadata

### LINKED ISSUE #407: Feature: Nodes information and credential
### Motivation:
When trying to debug issues I had to manually retrieve ssh keys directly from MongoDB and lookup node IP's in Terraformer logs. User should have easier access to LB and K8S Nodes. 

### Description:
User should be able to obtain information about nodes created by Claudie. I see it like this: Claudie after provisioning, next to the `secret`  containing `kubeconfig` will create another one with Node metadata ( ip addresses at minimum ) and credentials to it. ( username and ssh key ).

### Exit criteria:
- [ ] user can retrieve nodes ssh keys
- [ ] user can retrieve nodes metadata ( IP address at least )

Despire:
Closes #407 

This PR introduces a change where an additional secret is created to the kubeconfig that contains additional information about the created cluster:
- Private Key SSH key for nodes
- Public IP addresses for all nodes

The secret is created with the `metadata` suffix and contains data in JSON format in base64 encoding.
Screenshot 2022-11-30 at 16 41 58

**Is there any other information you would like to have (or is private key and IP addresses sufficient) ?**


MarioUhrik:
Hi @Despire , thanks a lot for this :) 
I think that private key and IP addresses are sufficient for now.

Couple questions:
- What happened with the CI? Did it run forever?
- What happens if the secret is removed by accident? Does Claudie recreate it?

Despire:
encountered issue #443 , so fixing it currently as the CI will fail...

if we delete the secret either kubeconfig or the metadata it won't be recreated automatically it needs to be triggered by either editing the input-manifest so the workflow starts again.,

MarioUhrik:
> encountered issue #443 , so fixing it currently as the CI will fail...

Thanks for fixing the CI issue :) 

> if we delete the secret either kubeconfig or the metadata it won't be recreated automatically it needs to be triggered by either editing the input-manifest so the workflow starts again.,

OK, this should suffice in the meantime

services/kuber/server/server.go
```diff
@@ -31,6 +32,11 @@ const (
 	outputDir        = "services/kuber/server/clusters"
 )
 
+type ClusterMetadata struct {
+	NodeIps    map[string][]net.IP `json:"node_ips"`
```
MiroslavRepka, line 36: What are the key and values here? From the code it looks like `[nodepool-name]<public ips in nodepool>[]`, but I would expect it to be something like `[node-name]<public IP, private IP>[]`, or something along those lines.

It would be easier for a user to find IP for individual node, not to look for it in nodepool.

services/kuber/server/server.go
```diff
@@ -55,96 +61,121 @@ func (s *server) SetUpStorage(ctx context.Context, req *pb.SetUpStorageRequest)
...
+
+	for _, pool := range req.GetCluster().GetClusterInfo().GetNodePools() {
+		for _, node := range pool.GetNodes() {
+			md.NodeIps[pool.Name] = append(md.NodeIps[pool.Name], net.ParseIP(node.Public))
+		}
+	}
+
+	b, err := json.Marshal(md)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal cluster metadata: %w", err)
```
MiroslavRepka, line 88: An error can be a little more specific, i.e. `fmt.Errorf("failed to marshal %s cluster metadata: %w", req.GetCluster().GetClusterInfo().GetName(), err)`

MiroslavRepka (commented):
LGTM, thanks for the PR @Despire!

services/kuber/server/server.go
Despire: Yes currently it's [nodepool-name][ array of all public ips for each node]
`[node-name]<public IP, private IP>[]` works as well 👍 , I think it might be better.

Despire: the data will now look like this

```json
{
    "node_ips": {
        "matus-t1-qa47cb5-hetzner-compute-1": {
            "public_ip": "192.168.2.2",
            "private_ip": "192.168.2.2"
        },
        "matus-t1-qa47cb5-hetzner-control-1": {
            "public_ip": "192.168.2.1",
            "private_ip": "192.168.2.1"
        }
    },
    "private_key": "...."
}
```

MiroslavRepka (approved):

services/kuber/server/server.go
```diff
@@ -31,6 +32,20 @@ const (
...
 
+type (
+	IPPair struct {
+		PublicIP  net.IP `json:"public_ip"`
+		PrivateIP net.IP `json:"private_ip"`
+	}
+
+	ClusterMetadata struct {
+		// NodeIps maps node-name to public-private ip pairs.
+		NodeIps map[string]IPPair `json:"node_ips"`
```
jaskeerat789, line 43: Do you think we should add private IP to this metadata as well? I am not sure if we can get this info using kubectl.
@Despire @MarioUhrik

jaskeerat789 (approved):
Thanks for feature addition and also some refactoring, changes LGTM @Despire

services/kuber/server/server.go
Despire: Not sure what you mean @jaskeerat789  there already is a private/public IP https://github.com/Berops/claudie/blob/acbfaad00155d0ad0faac82604df90b7f5b3f222/services/kuber/server/server.go#L91-L95

Despire:
Please Re-review the latest commit @MiroslavRepka https://github.com/Berops/claudie/pull/449/commits/b4758447e62003e20ea8cbdc3894cb4ab88302d3

I had to remove the OCI/GCP combination due to errors building clusters.
I had to remove this line https://github.com/Berops/claudie/blob/5a92ab8e1feaea51cef4a91d4841e1ad82a811bc/services/terraformer/templates/azure.tpl#L108

As terraformer would just fail to correctly resolve the association if the zones were changed in the public_ip, we still have the zone in the VM so there shouldn't be any problem with removing this line.

I've also remove 1 manifest from test-set2 and test-set1 that just increased the counts for each nodepool by one which I believe didn't really test much, I instead just increased/decreased counts in the other manifests.


The CI pipelines is still failing due to issues with deleting the kubernetes nodes... (Since I've rewritten the tests this issue watch catched in the CI pipeline the lonhorn nodes won't setup correctly, compute nodes get marged as master nodes for some unknown reason.)

Screenshot 2022-12-13 at 11 28 23

```bash
NAME                            STATUS   ROLES                  AGE   VERSION
ts1-gcp-wenvdk7-gcp-compute-1   Ready    <none>                 51m   v1.22.0
ts1-gcp-wenvdk7-gcp-compute-2   Ready    control-plane,master   54m   v1.22.0
ts1-gcp-wenvdk7-gcp-control-1   Ready    control-plane,master   54m   v1.22.0
ts1-gcp-wenvdk7-gcp-control-2   Ready    control-plane,master   53m   v1.22.0
ts1-gcp-wenvdk7-gcp-control-3   Ready    control-plane,master   52m   v1.22.0
```

MiroslavRepka:
The changes in the last commit look fine, thanks for resolving the issues. 

Regarding the pipeline marking the nodes incorrectly, I had a look in our code and I don't see a reason why they would be marked as master nodes. I saw that the kubeone version was updated in the previous PR, so I would look for an issue there. In the changelogs, they mention some changes regarding the annotations, but I haven't looked at them properly. It's just an educated guess :).

Despire:
I've actually just figured this out. Its due to this:
https://github.com/Berops/claudie/blob/5a92ab8e1feaea51cef4a91d4841e1ad82a811bc/services/terraformer/server/clusterBuilder/clusterBuilder.go#L223-L231

Its taking the NodeType and Private IP from a previous node by the IP address, And I have debugged this and it turns out the IP gets re-used and could end up taking the role of the Master

MiroslavRepka:
I see, yeah I haven't though about that. Great job!

PR #458, opened 2022-12-14
fix: CI issues

Despire:


services/builder/configProcessor.go
```diff
@@ -110,10 +110,11 @@ func configProcessor(c pb.ContextBoxServiceClient, wg *sync.WaitGroup) error {
 			config.CurrentState = tmpConfig.DesiredState
 		}
 		if toDelete != nil {
-			log.Info().Msgf("Deleting nodes for config %s", config.Name)
+			name := config.Name
```
bernardhalas, line 113: Just curious, how is this gonna work? If `config.Name` is `nil`, then `name` will become `nil` as well, won't it?

Despire:
```go
config, err = deleteNodes(config, toDelete)
```
we assign to config here (notice there is not := operator) thus in the `if err != nil` block the config will be nil and panic.

Despire: the panic happens in the `if err != nil` block when accessing `config.Name` as the function `deleteNodes` returned an error and config will be set to nil.

it only is being changed in this block because there is a return in the `if err != nil` block.

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf (approved):
Looks good to me.

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf (commented):
looks good.

is there any way to verify [this](https://github.com/Berops/claudie/blob/12895ea1e5d1e6ef0f668c0e9e70098507064168/internal/kubectl/kubectl.go#L160-L170) locally?
do we have unit tests for this?

Despire:
This is just for debugging to know where the issue is for me, I will remove this or just make it Debug only as there are lots of prints going on,

btw this PR is a work in progress I will push commit periodically

Despire:
There are still issues with the CI

Maybe the Install VPN issue is back because the test-set2 fails on ping timeout
Screenshot 2022-12-20 at 06 35 25

The 3.yaml from test-set1 fails on delete nodes (it just gets stuck there and nothing happens).
Screenshot 2022-12-20 at 06 29 58

Both of these work locally so I'm not sure where the problem might be.

MiroslavRepka (approved):
LGTM, thanks @Despire!

MiroslavRepka:
>The 3.yaml from test-set1 fails on delete nodes (it just gets stuck there and nothing happens).

Maybe memory/CPU requests can not keep up?

Despire:
There is a new issues in ansibler when changing the ApiEndpoint in test-set2
```
EP - ts2-lb-1   TASK [replace endpoint] ********************************************************
EP - ts2-lb-1   fatal: [ts2-c-1-joinbiw-hetzner-control-1]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: list object has no element 0\n\nThe error appears to be in '/bin/services/ansibler/server/ansible-playbooks/apiEndpointChange.yml': line 28, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n    - name: replace endpoint\n      ^ here\n"}
EP - ts2-lb-1   fatal: [ts2-c-1-joinbiw-hetzner-control-2]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: list object has no element 0\n\nThe error appears to be in '/bin/services/ansibler/server/ansible-playbooks/apiEndpointChange.yml': line 28, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n    - name: replace endpoint\n      ^ here\n"}
```

The installVPN bug can be bypassed by allocated the IPs which can be reached before running the CI
Screenshot 2022-12-28 at 12 23 13

As for the issue where kuber gets stuck on deleting the nodes this is still present.

Issue #448, opened 2022-11-30
Complete docs/input-manifest/providers/azure.md


MarioUhrik:
Azure was missing from https://github.com/Berops/claudie/pull/429, now it can be added.

Issue #439, opened 2022-11-28
Example inputManifest lacks Azure machines


MarioUhrik:
Let's try to ensure that the example inputManifest showcases all features Claudie has to offer, to the extent that makes reasonable sense for one inputManifest.

PR #471, opened 2022-12-27
Feat/update azure docs

### LINKED ISSUE #424: Update Storage docs
I suspect there's outdated information about storage classes.
We now also support Azure, OCI, and AWS.

I suggest to future-proof it somewhat, so that we don't need to keep coming back to this so often

https://github.com/Berops/claudie/blob/master/docs/storage/storage-solution.md

MiroslavRepka:
Clean up and clarification in docs related mostly to Azure. 

Closes #424 #439 #448

### DIFF
--- a/docs/input-manifest/example.yaml
+++ b/docs/input-manifest/example.yaml
@@ -149,9 +149,27 @@ nodePools:
         region: eu-central-1
         zone: eu-central-1c
       count: 2
-      server_type: t3.medium 
+      server_type: t3.medium
       # https://cloud-images.ubuntu.com/query/focal/server/released.current.txt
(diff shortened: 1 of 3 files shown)


Despire:
```yaml
image: Canonical:0001-com-ubuntu-minimal-focal:minimal-20_04-lts:20.04.202004230
```
can we use 22.04 image ? once https://github.com/Berops/claudie/pull/458 is merged the images will be 22.04

MiroslavRepka:
>can we use 22.04 image ? once https://github.com/Berops/claudie/pull/458 is merged the images will be 22.04

Definitely, thanks for noticing!

Despire (approved):

PR #473, opened 2022-12-28
Feat/cleanup docs

### LINKED ISSUE #389: Delete original proposals in docs
I think they just confuse people and don't bring much value anymore.
If they still bring value, then that value should be imported into the "current" doc.

MiroslavRepka:
This PR removes original proposals as described in issue #389. The information which provided some value is now included in the current solution docs. I also updated the LB diagram.

Closes #389


wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf (approved):
Looks good, the presence of those proposal documents was confusing to me.

PR #474, opened 2022-12-28
Update kubernetes version validation & docs

### LINKED ISSUE #416: Input validation & docs: supported K8s version
One particular version of KubeOne presumably doesn't support all versions of K8s.
See https://docs.kubermatic.com/kubeone/v1.5/architecture/compatibility/supported-versions/

We should:
- guard versions of K8s at the input validation level
- maintain a doc that states which versions of K8s are valid. Like here https://github.com/Berops/claudie/blob/master/docs/input-manifest/input-manifest.md#cluster-k8s

MiroslavRepka:
This PR updates the validation and docs regarding supported Kubernetes versions in Claudie.

Closes #416


Despire (approved):

internal/manifest/validation_test.go
```diff
@@ -0,0 +1,23 @@
+package manifest
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 6: technically, this looks good to me.

I have a couple of questions, though:
* why are we using `testify`? do we use it in other packages?
* why is the name of this file `validation_test.go`? since it's being added as a test file for `validate_kubernetes.go`, wouldn't it be more appropriate to name it `validate_kubernetes_test.go`?

what confuses me is that there is also another file named `validate_test.go` (as a complement test file to `validate.go`) and `validation_test.go` oddly sticks out. 

[`internal/manifest` folder structure](https://github.com/Berops/claudie/tree/ea216b780c08af642e7a99540f97c2dada992f8b/internal/manifest)

internal/manifest/validation_test.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+var (
+	testManifest                = &Manifest{NodePools: NodePool{Dynamic: []DynamicNodePool{{Name: "np1"}}}}
+	testClusterVersionPass      = &Kubernetes{Clusters: []Cluster{{Name: "cluster1", Network: "10.0.0.0/8", Version: "v1.22.0", Pools: Pool{Control: []string{"np1"}}}}}
+	testClusterVersionFailMinor = &Kubernetes{Clusters: []Cluster{{Name: "cluster1", Network: "10.0.0.0/8", Version: "v1.21.0", Pools: Pool{Control: []string{"np1"}}}}}
+	testClusterVersionFailMajor = &Kubernetes{Clusters: []Cluster{{Name: "cluster1", Network: "10.0.0.0/8", Version: "v2.22.0", Pools: Pool{Control: []string{"np1"}}}}}
+)
+
+func TestKubernetes(t *testing.T) {
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 16: also, could this function get a short comment? :)

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf (changes requested):
overall it looks good.
got a couple of questions, though, left some comments there.

internal/manifest/validation_test.go
MiroslavRepka: >why are we using testify? do we use it in other packages?

yes, in most of the test function.

>why is the name of this file validation_test.go? since it's being added as a test file for validate_kubernetes.go, wouldn't it be more appropriate to name it validate_kubernetes_test.go?

mb, have not noticed the `validate_test.go`

internal/manifest/validate_test.go
```diff
@@ -42,9 +48,21 @@ var (
 	}
 )
 
+// TestDomain tests the domain which will be formed from node name
 func TestDomain(t *testing.T) {
 	err := checkLengthOfFutureDomain(testDomainSuccess)
 	require.NoError(t, err)
 	err = checkLengthOfFutureDomain(testDomainFail)
 	require.Error(t, err)
+	fmt.Println(err)
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 57: what is the purpose of this print statement? do we further validate that the error is what we expect?

also, could this call be replaced by `testing`'s own `Log` or `Logf` so that we don't import the whole of `fmt` for this one line?

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf (changes requested):

internal/manifest/validation_test.go
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf: > yes, in most of the test functions.

cool, right.
thanks.

> ..have not noticed..

no worries.

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf (approved):

Despire:
Should we maybe have a table of supported k8s versions? 

| Kubernetes Version | Supported |
| ---------- | ----------- |
| 1.21 | ✕ |
| 1.22 | ✓ |

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
> Should we maybe have a table of supported k8s versions?
> 
> Kubernetes Version	Supported
> 1.21	✕
> 1.22	✓

absolutely!

MiroslavRepka:
Instead of listing it ourselves, I linked official kubeone docs as they can add or remove support for the current version (`1.5.`). So instead of us constantly checking and updating, users are one click away from knowing this information from the official source. thus I don't think it's necessary to have a table there.
 
 From docs:
 >The current `kubeone` version used in Claudie is `1.5`. To see the list of supported versions, please refer to `kubeone` [documentation](https://docs.kubermatic.com/kubeone/v1.5/architecture/compatibility/supported-versions/#supported-kubernetes-versions).

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
> Instead of listing it ourselves, I linked official kubeone docs as they can add or remove support for the current version (`1.5.`). So instead of us constantly checking and updating, users are one click away from knowing this information from the official source. thus I don't think it's necessary to have a table there.
> 
> From docs:
> 
> > The current `kubeone` version used in Claudie is `1.5`. To see the list of supported versions, please refer to `kubeone` [documentation](https://docs.kubermatic.com/kubeone/v1.5/architecture/compatibility/supported-versions/#supported-kubernetes-versions).

I like that now there at least *is* a link.

fair points @MiroslavRepka
IMO we could have both for posterity, I'm not going to push it, though. :)

MarioUhrik:
This looks good to me as it is, well done, thanks :)

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf (approved):
LGTM, good job @MiroslavRepka

MiroslavRepka:
@MarioUhrik pipeline was giving me an [error](https://github.com/Berops/claudie/actions/runs/3909391858/jobs/6683015350) that some tools were not installed, and while fixing it I decided to solve the warnings which we had in pipelines as well. Could you please review and approve the changes? :)

MarioUhrik (approved):
LGTM with https://github.com/Berops/claudie/pull/474/commits/368f3d36486957d0f8d951130f22a38f6f5e9f00 , but I'm wondering why is there "Get PR labels" in CI E2E, especially when its absent from CD E2E

MiroslavRepka:
`Get PR labels` is used to check whether the developer appended  `Disable Clean Up` label to the PR, which will set auto clean up to false, as described in PR #490 

For the CD, there is no need to disable clean-up as any issues should be debugged before merging.

MarioUhrik:
Got it, thanks @MiroslavRepka :)

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
> pipeline was giving me an [error](https://github.com/Berops/claudie/actions/runs/3909391858/jobs/6683015350) ...

just out of curiosity, do we know since when was this happening? is that related to the `20.04` -> `22.04` bump?

MiroslavRepka:
>just out of curiosity, do we know since when was this happening? is that related to the 20.04 -> 22.04 bump?

Which bump are you referring to? I am not aware of any upgrades done to our runners.

Afaik, this error is popping up only in this PR.

PR #475, opened 2022-12-29
fix:trim prefix for node names

Despire:
Fixes the issue where when the changeApiEndpoint playbook would reference the old node names with the prefix clusterName + clusterHash.

### DIFF
--- a/services/ansibler/server/server.go
+++ b/services/ansibler/server/server.go
@@ -78,12 +78,14 @@ func (*server) TeardownLoadBalancers(ctx context.Context, req *pb.TeardownLBRequ
 		attached        = make(map[string]bool)           // [k8sClusterName]Bool
 		k8sNodepools    = make(map[string][]*pb.NodePool) //[k8sClusterName][]nodepools
 		k8sNodepoolsKey = make(map[string]string)         //[k8sClusterName]keys
+		clusterIDs      = make(map[string]string)         //[k8sClusterName]prefix (clusterName + hash)
 	)
 
 	// Collect all NodePools that will exist after the deletion
 	for _, k8s := range req.CurrentState.Clusters {
 		k8sNodepools[k8s.ClusterInfo.Name] = k8s.ClusterInfo.NodePools
 		k8sNodepoolsKey[k8s.ClusterInfo.Name] = k8s.ClusterInfo.PrivateKey
+		clusterIDs[k8s.ClusterInfo.Name] = fmt.Sprintf("%s-%s", k8s.ClusterInfo.Name, k8s.ClusterInfo.Hash)
 	}
 
 	// Look for newly attached LoadBalancers.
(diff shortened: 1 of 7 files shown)


services/ansibler/server/server.go
```diff
@@ -97,7 +99,7 @@ func (*server) TeardownLoadBalancers(ctx context.Context, req *pb.TeardownLBRequ
 	for _, lb := range req.DeletedState.LoadBalancerClusters {
 		np, ok := k8sNodepools[lb.TargetedK8S]
 		if !ok {
-			log.Error().Msgf("LoadBalancer %s has not found a target k8s cluster %s", lb.ClusterInfo.Name, lb.TargetedK8S)
+			log.Info().Msgf("LoadBalancer %s has not found a target k8s cluster %s", lb.ClusterInfo.Name, lb.TargetedK8S)
```
MiroslavRepka, line 102: Isn't this considered an error/warning? If we have a Loadbalancer in a state which does not target any Kubernetes cluster, it should be caught by the frontend, thus if we detect it here, something went wrong.

services/ansibler/server/server.go
```diff
@@ -156,7 +161,7 @@ func (*server) SetUpLoadbalancers(_ context.Context, req *pb.SetUpLBRequest) (*p
 	for _, lb := range req.DesiredState.LoadBalancerClusters {
 		np, ok := k8sNodepools[lb.TargetedK8S]
 		if !ok {
-			log.Error().Msgf("Loadbalancer %s has not found a target k8s cluster (%s)", lb.ClusterInfo.Name, lb.TargetedK8S)
+			log.Info().Msgf("Loadbalancer %s has not found a target k8s cluster (%s)", lb.ClusterInfo.Name, lb.TargetedK8S)
```
MiroslavRepka, line 164: Isn't this considered an error/warning? If we have a Loadbalancer in a state which does not target any Kubernetes cluster, it should be caught by the frontend, thus if we detect it here, something went wrong.

services/ansibler/server/server.go
```diff
@@ -78,12 +78,14 @@ func (*server) TeardownLoadBalancers(ctx context.Context, req *pb.TeardownLBRequ
 		attached        = make(map[string]bool)           // [k8sClusterName]Bool
 		k8sNodepools    = make(map[string][]*pb.NodePool) //[k8sClusterName][]nodepools
 		k8sNodepoolsKey = make(map[string]string)         //[k8sClusterName]keys
+		clusterIDs      = make(map[string]string)         //[k8sClusterName]prefix (clusterName + hash)
```
MiroslavRepka, line 81: Just for the consistency
`//[k8sClusterName]prefix (clusterName + hash)`  -> `//[k8sClusterName]clusterID (<clusterName>-<hash>)`

services/ansibler/server/server.go
```diff
@@ -139,13 +142,15 @@ func (*server) SetUpLoadbalancers(_ context.Context, req *pb.SetUpLBRequest) (*p
 		lbInfos         = make(map[string]*LBInfo)        //[k8sClusterName]lbInfo
 		k8sNodepools    = make(map[string][]*pb.NodePool) //[k8sClusterName][]nodepools
 		k8sNodepoolsKey = make(map[string]string)         //[k8sClusterName]keys
+		clusterIDs      = make(map[string]string)         //[k8sClusterName]prefix (clusterName + hash)
```
MiroslavRepka, line 145: Just for the consistency
`//[k8sClusterName]prefix (clusterName + hash)`  -> `//[k8sClusterName]clusterID (<clusterName>-<hash>)`

MiroslavRepka (changes requested):
LGTM, please see my comments. Thanks, @Despire for spotting the issue!

Despire:
Changes implemented, also changed to `inventory_hostname` which actually will use the name from the inventory file.

MiroslavRepka (approved):

PR #490, opened 2023-01-02
Feat/auto clean up

MiroslavRepka:
This PR implements an option for automatic infrastructure clean-up even if the test manifest fails. This option defaults to true, meaning all infra should be deleted by Claudie, even if it fails. To override this option, use the label `Disable clean up` on a PR, where the pipeline will run. 

The PR was tested by 
- adding the label, which resulted in the environment variable being false (as expected)
- removing the label, triggering the CI with a new commit, which resulted in the environment variable being true (as expected)

The changes in the behaviour based on the flag were tested locally.


bernardhalas:
Hi @MiroslavRepka , while I agree that this was a good issue to tackle, I'm not convinced the best implementation path has been chosen. I was expecting a conclusion on the relevant discussion first. Ref: https://github.com/Berops/infra/issues/142#issuecomment-1320028787

services/testing-framework/claudie_test.go
```diff
@@ -149,55 +200,54 @@ func applyTestSet(setName, namespace string, c pb.ContextBoxServiceClient) error
...
 				}
-				//manifest is done
-				return nil
 			}
+			if elapsedSec >= maxTimeout {
+				return fmt.Errorf("Test took too long... Aborting after %d seconds", maxTimeout)
+			}
+			time.Sleep(time.Duration(sleepSec) * time.Second)
+			counter++
+			log.Info().Msgf("Waiting for %s to from %s finish... [ %ds elapsed ]", manifestName, testSetName, elapsedSec)
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 249: I believe there's a typo in this log message.

services/testing-framework/claudie_test.go
```diff
@@ -279,3 +329,21 @@ func checkIfManifestSaved(configID string, idType pb.IdType, c pb.ContextBoxServ
 		}
 	}
 }
+
+// cleanUp will delete manifest from claudie which will trigger infra deletion
+// it deletes a secret if claudie is deployed in k8s cluster
+// it calls for a deletion from database directly if claudie is deployed locally
+func cleanUp(id string, c pb.ContextBoxServiceClient) error {
+	if envs.Namespace != "" {
+		//delete secret from namespace
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 338:
```go
//abc
```
vs
```go
// abc
```
It is also advised that comments are written as full sentences ending in full stops.

services/testing-framework/claudie_test.go
```diff
@@ -149,55 +200,54 @@ func applyTestSet(setName, namespace string, c pb.ContextBoxServiceClient) error
...
-	if namespace != "" {
-		if err = deleteSecret(setName, namespace); err != nil {
-			return fmt.Errorf("error while deleting the secret %s from %s : %w", pathToTestSet, namespace, err)
-		}
-	} else {
-		// delete config from database
-		if err = cbox.DeleteConfig(c, idInfo.id, pb.IdType_HASH); err != nil {
-			return fmt.Errorf("error while deleting the manifest from test set %s : %w", pathToTestSet, err)
-		}
+	//delete manifest from DB to clean up the infra after configChecker is done without error
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 203:
```go
//abc
```
vs
```go
/// abc
```

services/testing-framework/claudie_test.go
```diff
@@ -140,7 +179,19 @@ func applyTestSet(setName, namespace string, c pb.ContextBoxServiceClient) error
 			}
 		}
 		// wait until test config has been processed
-		if err := configChecker(c, pathToTestSet, manifest.Name(), idInfo); err != nil {
+		if err := configChecker(ctx, c, pathToTestSet, manifest.Name(), idInfo); err != nil {
+			if cleanUpFlag == "TRUE" {
+				log.Info().Msgf("Deleting infra even after error due to flag \"-auto-clean-up\" set to %v : %v", cleanUpFlag, err)
+				//delete manifest from DB to clean up the infra
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 185:
```
//abc
```
vs
```
// abc
```

services/testing-framework/claudie_test.go
```diff
@@ -32,9 +35,49 @@ const (
...
+
+		return err
+	})
+
+	//start E2E tests in separate goroutines
+	group.Go(func() error {
+		return testClaudie(ctx)
+	})
+
+	//wait for either test to finish or interrupt signal to occur
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 74: `//abc` -> `// abc`

services/testing-framework/claudie_test.go
```diff
@@ -32,9 +35,49 @@ const (
...
+			err = ctx.Err()
+		case sig := <-ch:
+			log.Warn().Msgf("Received signal %v", sig)
+			err = errors.New("testing-framework received interrupt signal")
+		}
+
+		return err
+	})
+
+	//start E2E tests in separate goroutines
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 69: `//abc` -> `// abc`

services/testing-framework/claudie_test.go
```diff
@@ -32,9 +35,49 @@ const (
 	maxTimeoutSave = 60 * 12 // max allowed time for config to be found in the database
 )
 
+var (
+	//get env var from runtime directly so we do not pollute original envs package by unnecessary variables
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 39: `//abc` -> `// abc`

services/testing-framework/claudie_test.go
```diff
@@ -32,9 +35,49 @@ const (
...
 // TestClaudie will start all the test cases specified in tests directory
 func TestClaudie(t *testing.T) {
 	utils.InitLog("testing-framework")
+	group, ctx := errgroup.WithContext(context.Background())
+
+	// start goroutine to check for SIGTERM in case pod is being deleted
+	// or interrupt in case local testing is aborted
+	group.Go(func() error {
+		ch := make(chan os.Signal, 1)
+		signal.Notify(ch, os.Interrupt, syscall.SIGTERM)
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 54: this here thing makes sure we're only notified on {SIGINT|SIGTERM}. would a comment be too redundant?

services/testing-framework/claudie_test.go
```diff
@@ -32,9 +35,49 @@ const (
...
+		defer signal.Stop(ch)
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 55: what happens when we stop being notified about signals sent to our program? I can see that we return an error a couple of lines below (after the "select-case"), where is it being handled afterwards, please?

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf (changes requested):
I like this implementation, it's quite straight-forward, I left some comments, though.

also, as @bernardhalas has pointed out, I suppose the discussion about whichever implementation to go forward with ought to be concluded first (even though I said liked what you've done here, @MiroslavRepka).

services/testing-framework/claudie_test.go
bernardhalas: There's little to no value in the comment formatting change. Isn't there?

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf: I don't agree, the formatting is inconsistent.
also, have a look at literally all comments in e.g. https://tip.golang.org/doc/comment, they're all `// abc` style.
I agree that it's inconsequential to how the program runs, however, people are going to read the code, which means there's value in formatting.

services/testing-framework/claudie_test.go
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf: awesome :rocket:
cheers!

MarioUhrik:
Hi guys, I think that this PR is fine, and I think we don't have to go far too deep into discussions on how to tackle https://github.com/Berops/infra/issues/142.
This PR is a massive improvement over status quo. If it turns out not to be perfect, we'll manage.
I personally think this implementation will suffice for the time being, with only very low risk of things going wrong.

manifests/testing-framework/testing-framework.yaml
```diff
@@ -30,6 +30,11 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
+            - name: AUTO_CLEAN_UP
```
MarioUhrik, line 33: Hi @MiroslavRepka , I'm just wondering whether this works when AUTO_CLEAN_UP is missing from the `env` file, e.g. in cases where you try to deploy it manually.
But I guess that's not a big deal anyway.

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf: hey @MarioUhrik, if the question is whether the infra would get auto-cleaned up when `AUTO_CLEAN_UP` variable is not present in the environment, the answer is no.

the current implementation [performs a lookup](https://github.com/Berops/claudie/blob/5ef343dcc2f735ecb76db78d64334ded11e1b055/services/testing-framework/claudie_test.go#L40) using [`os.Getenv`](https://pkg.go.dev/os#Getenv), which doesn't really mind if the ENV var is not present; it simply returns an empty string in that case, which subsequently fails our check `if ... == "TRUE"` [here](https://github.com/Berops/claudie/blob/5ef343dcc2f735ecb76db78d64334ded11e1b055/services/testing-framework/claudie_test.go#L182), i.e. the infrastructure would not get automatically cleaned up.

Somebody correct me, please, if I'm wrong.

services/testing-framework/claudie_test.go
```diff
@@ -32,9 +35,48 @@ const (
 	maxTimeoutSave = 60 * 12 // max allowed time for config to be found in the database
 )
 
+var (
+	// get env var from runtime directly so we do not pollute original envs package by unnecessary variables
+	cleanUpFlag = os.Getenv("AUTO_CLEAN_UP")
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 40: my only suggestion here would be to name the thing `autoCleanUpFlag`, so that it's clearer.

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf (changes requested):

manifests/testing-framework/testing-framework.yaml
MarioUhrik: I was specifically wondering whether Kubernetes can accept a Deployment with an env variable sourced from a configmap where such a variable isn't present

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf: ah, I see, that's a good question.

the [Optional ConfigMaps](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#optional-configmaps) section ([added](https://github.com/kubernetes/website/pull/35160) not so long ago) should answer the original question: "If the ConfigMap exists, but the referenced key is non-existent the data is also empty.".

that is, the `AUTO_CLEAN_UP` key needs to be marked as **optional**.

manifests/testing-framework/testing-framework.yaml
```diff
@@ -30,6 +30,11 @@ spec:
...
+              valueFrom:
+                configMapKeyRef:
+                  name: env
+                  key: AUTO_CLEAN_UP
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 37: this key needs to be marked as optional, otherwise the pods might fail to start in case it's missing (manual triggers).

refs:
* https://github.com/kubernetes/website/pull/35160/files#diff-20d118b7f6f20eeff74747a63494a459bf067d00cb3fe543b858ec8222b754c8R702
* https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#optional-configmaps

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf (changes requested):

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf (approved):
looks good :rocket:
good job @MiroslavRepka

MiroslavRepka:
>I see you're still bumping the deps.

That's from master #496, this PR is done. Thx for review

PR #507, opened 2023-01-10
feat: use fully qualified registry names

### LINKED ISSUE #503: Feature: Use fully qualified registry names in Dockerfiles
### Motivation:
Currently, the service Dockerfiles use unqualified image names, which is generally not an issue with Docker. However, I recently tried to build an image with Podman/Buildah and that failed the first time I tried to build it (without me changing the Dockerfile or `/etc/containers/registries.conf`) with "image unqualified".

### Description:
I expect to be able to build the images with Podman (`libbuildah`).

### Exit criteria:
- [ ] service image Dockerfiles contain fully qualified registry names
- [ ] consequently, service images build with Podman

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
as title, use fully qualified registry names in service Dockerfiles.
fixes #503.

### DIFF
--- a/services/kube-eleven/Dockerfile
+++ b/services/kube-eleven/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.19.4 AS build
+FROM docker.io/library/golang:1.19.4 AS build
 
 # download and unzip kube-one binary
 RUN apt-get -qq update && apt-get -qq install unzip
@@ -39,7 +39,7 @@ WORKDIR /go/services/kube-eleven/server
 #Compile the golang code, CGO_ENABLE=0 removes cross compile dependencies
 RUN CGO_ENABLED=0 go build
 
-FROM alpine
+FROM docker.io/library/alpine
 
 RUN apk update
 RUN apk add -q bash
--- a/services/kube-eleven/Dockerfile
+++ b/services/kube-eleven/Dockerfile
@@ -39,7 +39,7 @@ WORKDIR /go/services/kube-eleven/server
 #Compile the golang code, CGO_ENABLE=0 removes cross compile dependencies
 RUN CGO_ENABLED=0 go build
 
-FROM docker.io/library/alpine
+FROM docker.io/library/alpine:3.17.1
 
 RUN apk update
 RUN apk add -q bash
--- a/services/terraformer/Dockerfile
+++ b/services/terraformer/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.19.4 AS build
+FROM docker.io/library/golang:1.19.4 AS build
 
 # download and unzip kube-one binary
 RUN apt-get -qq update && apt-get -qq install unzip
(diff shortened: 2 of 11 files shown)


MarioUhrik:
I think we might as well wait for @MiroslavRepka to finish his work, which will deprecate this PR - correct me if I'm wrong

MiroslavRepka:
I don't think this regards a registry we use, rather than a tool people will use to build an image. It is explained in the linked issue (#503).

 It is nothing we have to worry about for now since we use `docker build`, however, if we switch to other tools like a `podman` the issue becomes relevant.

services/kube-eleven/Dockerfile
```diff
@@ -39,7 +39,7 @@ WORKDIR /go/services/kube-eleven/server
 #Compile the golang code, CGO_ENABLE=0 removes cross compile dependencies
 RUN CGO_ENABLED=0 go build
 
-FROM alpine
+FROM docker.io/library/alpine
```
MiroslavRepka, line 42: Can we pin the version of the `alpine` image to some reasonable version while we are at it?

services/terraformer/Dockerfile
```diff
@@ -40,7 +40,7 @@ WORKDIR /go/services/terraformer/server
 #Compile the golang code to /out, CGO_ENABLE=0 removes cross compile dependencies
 RUN CGO_ENABLED=0 go build
 
-FROM alpine
+FROM docker.io/library/alpine
```
MiroslavRepka, line 43: ditto

MiroslavRepka (changes requested):
LGTM, please see my comments.

MarioUhrik:
Sorry guys, I completely misunderstood this PR/issue.
I agree we can merge this.

I mistakenly thought this is about the `image:` field inside the Deployment manifest.

FYI @wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
no worries there :)) @MarioUhrik 


> I don't think this regards a registry we use, rather than a tool people will use to build an image. It is explained in the linked issue (#503).
> 
> It is nothing we have to worry about for now since we use `docker build`, however, if we switch to other tools like a `podman` the issue becomes relevant.
> 
> Afaik, we can merge this, it will not break anything, since the pipeline built images successfully.

correct, thanks, yes, this concerns the paths of the (base) images used inside the `service/x` Dockerfiles.


> I agree we can proceed with this.

awesome :rocket:

services/kube-eleven/Dockerfile
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf: absolutely, we could do `3.17` or `3.17.1`, depending on for how long we want to call it a day (minor version tag is more precise, although it still isn't a hash).

PR #509, opened 2023-01-10
feat: add Code of Conduct

### LINKED ISSUE #508: Feature: add Code of Conduct
### Motivation:
The project could benefit from having a Code of Conduct.

### Description:
A Code of Conduct document is added.

### Exit criteria:
- [x] Code of Conduct exists in the repo

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
as title, add Code of Conduct file.
fixes #508.

the email address listed is of course subject to discussion.


bernardhalas:
Listing the email address publicly here might result in elevated levels of spam. Shall we try it out nevertheless?

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
> Listing the email address publicly here might result in elevated levels of spam. Shall we try it out nevertheless? Alternatively this can be reported on the Claudie Slack workspace.

fair point.
while I can imagine not everyone is going to be thrilled by the use of Slack, I think it's a good idea.

one thing to consider is that if the issue concerned a rogue internal employee, they would have a harder time manipulating email than Slack IMO.

also, we should then probably have a think about retention periods of possible reports (Slack vs. email)..

bernardhalas:
I still prefer Slack here. In there the person affected can decide which maintainer to approach. That is not the case with the email address (and maintainers can be the ones breaking their own Code of Conduct. So let's rephrase this accordingly and I think this is good to go.

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
> I still prefer Slack here. In there the person affected can decide which maintainer to approach. That is not the case with the email address (and maintainers can be the ones breaking their own Code of Conduct. So let's rephrase this accordingly and I think this is good to go.

similarly to how the security policy is handled, the email could be forwarded to all maintainers easily.

I also believe that email is preferable to Slack from accessibility point of view.

what would you say to doing both Slack and an `claudie-abuse` email address? @bernardhalas

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Scope
+
+This Code of Conduct applies both within project spaces and in public spaces when an individual is representing the project or its community. Examples of representing a project or community include using an official project e-mail address, posting via an official social media account, or acting as an appointed representative at an online or offline event. Representation of a project may be further defined and clarified by project maintainers.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be reported by contacting the project team either at [claudie-abuse at berops dot com](mailto:claudie-abuse@berops.com) email address or via [project's Slack workspace](https://join.slack.com/t/claudie-workspace/shared_invite/zt-1imfso8r4-xwrpZjL9kt61FT1LjvWD5w). The project team will review and investigate all complaints, and will respond in a way that it deems appropriate to the circumstances. The project team is obligated to maintain confidentiality with regard to the reporter of an incident. Further details of specific enforcement policies may be posted separately.
```
MarioUhrik, line 37: Please note that the slack invitation link expires relatively fast.
Instead of duplicating it, I suggest referring to the one in the primary README.

bernardhalas:
I see the reasoning. We can go with `abuse@domain`, set it as an alias and let's see how it turns out. In that case, we can even leave Slack out of the picture for now.

CODE_OF_CONDUCT.md
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf: I didn't know that, thanks @MarioUhrik

MiroslavRepka (approved):

MarioUhrik:
I am OK with merging this, let's just make sure that the email address is valid though

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
> `abuse@` is not allowed. I've set up `claudie-abuse` instead.

nm, this is already present in the document.

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
:rocket:

Issue #369, opened 2022-11-02
Revisit CPU and Memory resource limits and requests 


jaskeerat789:
### Current Behavior:
A lot of beta-testers have been facing issues with crash looping pods due to liveness probes not responding(specially for context-box). We believe that these issues might be cause by low cpu and memory limits set for pods. After taking a brief look at the `metric explorer` on GCP, we could see a  lot of services hitting their requested limits for cpu and memory. We certainly want to look at this and tune the limits in order to make sure that our services are getting enough resource required for a healthy functioning.

@MarioUhrik and I had a discussion regarding `context-box` limits since it was giving problems to most of the testers. We decided that the limits on `context-box` pod were very low. In response to this #368 PR was opened, bumping up resource limits and requests to much higher relative values.

PR #515, opened 2023-01-11
feat/increase-requests

Despire:
#369 

Increase the limits for terraformer, ansibler, context-box and builder, also increase the interval in which the healthcheck is called to 30 seconds (instead of 10s the default value)


MarioUhrik (approved):
LGTM

MiroslavRepka (approved):
LGTM, thanks @Despire!

PR #517, opened 2023-01-11
feat(dockerfiles): use $TARGETARCH everywhere

### LINKED ISSUE #468: Feature: Arm binaries inside docker images
### Description:

When building multi-arch docker images the binaries downloaded inside the images should be of the same arch as the docker image

For example here we always download amd64 binary
https://github.com/Berops/claudie/blob/b7af0962c73bf57d694d521f21422ef688ca18ef/services/kuber/Dockerfile#L4
### Exit criteria:
- [ ] Download binaries based on the arch type

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
`$TARGETARCH` variable is automatically passed into images during the multi-arch build process, and all that is then left on our side is to consume it.
`$TARGETARCH` contains, as the name suggests, the target architecture (arm64, amd64, ppc64le...) of the image currently being built (as requested via `--platform x`).

* also use fully qualified image registry paths in Dockerfiles

fixes #468
EDIT: replaces #507
closes #507


MarioUhrik:
This looks pretty sweet, thanks !

I'm just wondering, is there a way to set some sort of a default TARGETARCH so that the Dockerfiles remain usable locally, outside buildx ?
Something like "if TARGETARCH isn't defined, then TARGETARCH=amd64" ?
If not, no worries.

MiroslavRepka:
>also use fully qualified image registry paths in Dockerfiles

Does it make sense to have #507 open then? Should we not close it in favour of this PR? @wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf

MiroslavRepka (approved):

MarioUhrik:
> > This looks pretty sweet, thanks !
> > I'm just wondering, is there a way to set some sort of a default TARGETARCH so that the Dockerfiles remain usable locally, outside buildx ? Something like "if TARGETARCH isn't defined, then TARGETARCH=amd64" ? If not, no worries.
> 
> thanks for reminding me, I wanted to have a look into the build hooks that could provide what you're asking about (they're basically shellscripts that run at `docker build` time..).

Could you also please check if this is possible from within a Dockerfile somehow?
I'd like to avoid the necessity of local configurations on the developers' machines, if that's what you're thinking

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
> > > This looks pretty sweet, thanks !
> > > I'm just wondering, is there a way to set some sort of a default TARGETARCH so that the Dockerfiles remain usable locally, outside buildx ? Something like "if TARGETARCH isn't defined, then TARGETARCH=amd64" ? If not, no worries.
> > 
> > 
> > thanks for reminding me, I wanted to have a look into the build hooks that could provide what you're asking about (they're basically shellscripts that run at `docker build` time..).
> 
> Could you also please check if this is possible from within a Dockerfile somehow? I'd like to avoid the necessity of local configurations on the developers' machines, if that's what you're thinking

right, apart from having a if-then-fi shell block after each `RUN` instruction, that would check if the variable is set, I don't think so.
this does not require any set-up from anybody, it's already there when you clone the repo (or check out the changes) and should work with `docker` seamlessly.
I can try out the Dockerfile approach though, but it's not going to look very nice, I expect.

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
will let you guys know when it's ready then..

MarioUhrik:
OK @wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf , I don't want to push us into something that isn't nice.
Please disregard my comments, I'm leaving it up to you :)

MarioUhrik:
Hi @wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf,

the next steps for this PR are likely going to involve the following:
- resolving git conflicts
- thoroughly testing the PR
  - ensuring that `docker build` works both locally and in E2E (with and without `buildx`)
  - checking whether ARM support is now fully ready
- asking Miro and Matus for reviews

I have mixed feelings about the hooks, I'm not sure they're worth it, so I'd rely on opinions of Miro and Matus here.
If we conclude that the hooks aren't worth it, I'd be quite fine with it, the cost being that Dockerfiles will only be usable from within the pipeline.

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
> Hi @wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf,
> 
> the next steps for this PR are likely going to involve the following:
> 
> * resolving git conflicts
> * thoroughly testing the PR
>   
>   * ensuring that `docker build` works both locally and in E2E (with and without `buildx`)
>   * checking whether ARM support is now fully ready
> * asking Miro and Matus for reviews
> 
> I have mixed feelings about the hooks, I'm not sure they're worth it, so I'd rely on opinions of Miro and Matus here. If we conclude that the hooks aren't worth it, I'd be quite fine with it, the cost being that Dockerfiles will only be usable from within the pipeline.

thanks @MarioUhrik.
- [x] conflicts resolved
- [x] local builds work
- [ ] E2E tests pending pipeline completion
- [ ] once images are built and pushed to GHCR, test in an ARM cluster

the hooks are (IMO) the nicer alternative to directly polluting the Dockerfiles with branching logic to achieve the same thing. of course, there's a lot of duplication, but I don't know how to solve that, given there's quite a lot of Dockerfiles...

also, I am not convinced this is good enough of a reason to break local container builds, even though at the moment the Dockerfiles are mostly being used in the CI; especially if we could agree on the hooks.
@MarioUhrik could you perhaps elaborate on why you have mixed feelings about the hooks?


cc @MiroslavRepka @Despire

MarioUhrik:
> @MarioUhrik could you perhaps elaborate on why you have mixed feelings about the hooks?

Miro said it best - small value, large footprint

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
> If we miss any dependency, it will show up during testing.

one last thing I am not quite sure about is whether the images get pushed to GHCR after being built with these build hooks, since a build-and-push has been replaced by build-only docker command. if this does not get handled "automagically", another condition will need to be added to the hook, checking whether a "docker login" is available (not yet sure how to check for that just yet) and we'll need to run docker push at the end of the build ourselves (in the hook)...

> In terms of the hooks, I had a brief look at the documentation, it is a shame that we cannot specify the path to the `hooks` directory and we have to have duplicates of them.

exactly, but that's what we're working with. another option would probably be separate Dockerfiles...

> The value they provide is rather small (the ability to build images locally), yet the footprint is large. If it would be up to me, I would probably keep them as we do not need to build images locally, but when we do, it would suck to wait for CI to build them.

right..

Despire:
It think this look great!

I'm also not a fan of using hooks, if it was just a single file I'd be fine with it but duplicating it for each service seems to bloated for me.

I was wondering if we could use `uname -m` when building the docker images locally and just add the script to the makefile

```
docker build --build-arg TARGETARCH=$(uname -m)  ...
```

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
thanks for the feedback @Despire.

----

> I'm also not a fan of using hooks, if it was just a single file I'd be fine with it but duplicating it for each service seems to bloated for me.

I know what you mean. unfortunately, {pre,-post-,}build hooks need to be present near each Dockerfile, it's a limitation of the program...another way to look at it could be that just as each service has its Dockerfile, it also has its own build hook...even when they all do the same..I know..

----

> I was wondering if we could use `uname -m` (or fetching the arch from `go env GOHOSTARCH`) when building the docker images locally and just add the logic to the makefile
> 
> ```
> docker build --build-arg TARGETARCH=$(uname -m)  ...
> ```

this is very much viable but would need to be tweaked slightly with `uname -m` - for example it would fail on x86_64, since what we want there is `amd64` but `uname -m` yields `x86_64`. however, the build hook logic *could* at least be moved to the Makefile, removing the duplication.

using `go env GOHOSTARCH` would be even better, I honestly haven't thought of this before.
the only prerequisite for this would be Go, which needs to be installed anyway, and it would obviate the decision-making logic, both in the hook or Makefile, whichever we choose in the end.
I like this one - Makefile+`go env GOHOSTARCH` - the most. :rocket:

also, CI would *not* be negatively affected at all, since `$TARGETARCH` gets passed in during `docker buildx ...`.

cc @MiroslavRepka

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
I am doing a try build of the container images on an RPI4 (a 64 bit ARM machine)..

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
alright, looks like we hit the thing again:
image
> 9 nodes had an untolerated taint (kubernetes.io/arch: arm64)

this is weird.

cc @Despire @MarioUhrik

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
I have found out that (for some reason) the nodes all have this `kubernetes.io/arch=arm64:NoSchedule` taint (last line):
image

removing the taint schedules the pods just fine...
image

MarioUhrik:
Hi @wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf , did you confirm that Claudie now works perfectly when deployed on ARM? How about the E2E test? Do you think we can merge this?

Despire (approved):
LGTM,

The e2e tests failed on the installVPN issue, but these changes doesn't affect it in any way.

MarioUhrik (approved):

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
> did you confirm that Claudie now works perfectly when deployed on ARM?

yes, Claudie was running just fine in the ARM cluster, once the `NoSchedule` taint was removed.

PR #522, opened 2023-01-12
Replace the diagram with Bernard's new one

MarioUhrik:
The README was still showing the old one.
Instead, we want to ditch the old one and show the new one.


bernardhalas:
That was really lame from me. Many thanks!

bernardhalas (approved):

Issue #462, opened 2022-12-15
Feature: Exponential timeout


Despire:
### Motivation:

Networking issues are common, traffic getting lost, slow connections etc... We should switch to exponential timeout that cancels at 5 min.

Make the timeout exponential with a hard limit between ~5-10 min

### Exit criteria:
- [ ] Implement exponential backoff with a hard limit between 5-10 min

bernardhalas:
Very good topic @Despire . I'd like to throw in an idea - any other than exponential back-off algorithm all the way? This way our concern would be just the initial delay + retry count.

Despire:
I was thinking maybe we could use this library https://github.com/cenkalti/backoff
already has constant, exponential backoff implemented less work on our side.

PR #523, opened 2023-01-13
Feat/expon timeout

MiroslavRepka:
This PR implements an exponential backoff as explained in #462 using the function `5 * 2^n`, with the hard limit set as 5min. 

It also adds a command cancellation after a set amount of time to prevent commands from dangling indefinitely, as seen in Kuber or Kube-eleven. The time after which the command is cancelled is set individually for each service. I chose some default values, let me know if you think it's too long or too short.


internal/command/cmd.go
```diff
@@ -36,67 +37,89 @@ const (
 	colorOkay  = "\x1b[32m"
 	colorFail  = "\x1b[31m"
 	colorReset = "\x1b[0m"
+	maxBackoff = 5 * 60 // max backoff time [5 min]
 )
 
-// RetryCommand will retry the given command, every 5 sec until either successful or reached numOfRetries
+// RetryCommand will retry the given command, with exponential backoff, maxing at 5 min, for numOfRetries times
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 43: nit:
```diff
-...will retry...
+...retries...
```
nit2: dots at the end of sentences.

internal/command/cmd.go
```diff
@@ -36,67 +37,89 @@ const (
...
-		backoff := 10 * i
-		log.Info().Msgf("Next retry in %ds...", backoff)
-		time.Sleep(time.Duration(backoff) * time.Second)
 	}
 	log.Error().Msgf("Command %s was not successful after %d retries", c.Command, numOfRetries)
 	return err
 }
 
-// RetryCommandWithOutput will retry the given command, every 5 sec until either successful or reached numOfRetries
+// RetryCommandWithOutput will retry the given command,  with exponential backoff, maxing at 5 min, for numOfRetries times
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 64: 1. `will retry` --> `retries`.
2. sentences ending in `.`.

internal/command/cmd.go
```diff
@@ -36,67 +37,89 @@ const (
...
 		log.Warn().Msgf("Error encountered while executing %s : %v", c.Command, err)
-		backoff := 10 * i
-		log.Info().Msgf("Next retry in %ds...", backoff)
-		time.Sleep(time.Duration(backoff) * time.Second)
 	}
 	log.Error().Msgf("Command %s was not successful after %d retries", c.Command, numOfRetries)
 	return nil, err
 }
 
+// execute executes the cmd with context canceled after commandTimeout seconds
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 86: dots (`.`) to end the sentences.

internal/command/cmd.go
```diff
@@ -36,67 +37,89 @@ const (
...
+// buildCmd prepares a exec.Cmd datastructure with context
+func (c *Cmd) buildCmd(ctx context.Context) *exec.Cmd {
+	cmd := exec.CommandContext(ctx, "bash", "-c", c.Command)
+	cmd.Dir = c.Dir
+	cmd.Stdout = c.Stdout
+	cmd.Stderr = c.Stderr
+	return cmd
+}
+
+// getNewBackoff returns a new backoff 5 * (2 ^ iteration), with the hard limit set at maxBackoff
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 113: nit: `.` at the end of the sentence.

internal/command/cmd_test.go
```diff
@@ -0,0 +1,20 @@
+package command
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestCmd will test a command retry and cancellation
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 9: nit: `.` at the end of the sentence.
nit2: `tests` instead of `will test`.

internal/command/cmd.go
```diff
@@ -36,67 +37,89 @@ const (
...
+// execute executes the cmd with context canceled after commandTimeout seconds
+// returns error if unsuccessful, nil otherwise
+func (c *Cmd) execute(i, numOfRetries, commandTimeout int) error {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(commandTimeout)*time.Second)
+	defer cancel()
+	log.Warn().Msgf("Retrying command %s... (%d/%d)", c.Command, i, numOfRetries)
+	return c.buildCmd(ctx).Run()
+}
+
+// executeWithOutput executes the cmd with context canceled after commandTimeout seconds
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 95: nit: `.` at the end of the sentence.

internal/command/cmd.go
```diff
@@ -36,67 +37,89 @@ const (
...
+// returns error, nil if unsuccessful, nil, output otherwise
+func (c *Cmd) executeWithOutput(i, numOfRetries, commandTimeout int) ([]byte, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(commandTimeout)*time.Second)
+	defer cancel()
+	log.Warn().Msgf("Retrying command %s... (%d/%d)", c.Command, i, numOfRetries)
+	return c.buildCmd(ctx).CombinedOutput()
+}
+
+// buildCmd prepares a exec.Cmd datastructure with context
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 104: nit: `.` at the end of the sentence.

internal/command/cmd_test.go
```diff
@@ -0,0 +1,20 @@
...
+func TestCmd(t *testing.T) {
+	cmd := Cmd{"sleep 2 && ls", "", nil, nil}
+	err := cmd.RetryCommand(1, 1)
+	require.Error(t, err)
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 13: we expect an error here because the command times out before it could be retried? adding a couple of comments here could make it clearer.

services/kube-eleven/server/kubeone/kubeone.go
```diff
@@ -10,7 +10,8 @@ import (
 )
 
 const (
-	maxNumOfRetries = 5 //max number of retries for kubeone apply
+	maxNumOfRetries = 5   //max number of retries for kubeone apply
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 13: nit: `.` at the end of the sentence.

services/terraformer/server/terraform/terraform.go
```diff
@@ -14,6 +14,8 @@ const (
 	// it succeeds. If after 10 retries the commands still fails an error should be
 	// returned containing the reason.
 	maxTerraformRetries = 10
+	// cancel terraform command after terraformTimeout seconds
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 17: nit: `.` at the end of the sentence.

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf (changes requested):
I like this, left a couple of comments (nits + questions). @MiroslavRepka

internal/command/cmd_test.go
MiroslavRepka: >we expect an error here because the command times out before it could be retried

Yes, an error occurs because the command is cancelled before it ends the execution, for every retry intentionally.

internal/command/cmd.go
```diff
@@ -36,67 +37,89 @@ const (
...
+	cmd := exec.CommandContext(ctx, "bash", "-c", c.Command)
+	cmd.Dir = c.Dir
+	cmd.Stdout = c.Stdout
+	cmd.Stderr = c.Stderr
+	return cmd
+}
+
+// getNewBackoff returns a new backoff 5 * (2 ^ iteration), with the hard limit set at maxBackoff
+func getNewBackoff(iteration int) int {
+	backoff := 5 * (2 ^ iteration)
```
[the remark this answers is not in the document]
MiroslavRepka: > 0 is not passed in anyway? 

As you said, 0 is fine since 2^0 = 1

>why are we iterating from 1 then? 

For easier logging

> in that case, we'd need to make sure that iteration >= 0

Don't necessarily have to, since the function works with negative numbers as well. i.e. `5 * 2^(-1) = 2.5`. The backoff function is clearly stated in the comment, thus if developer wants to start with a backoff smaller than 5, they can.

internal/command/cmd.go
```diff
@@ -36,75 +37,97 @@ const (
...
 	log.Error().Msgf("Command %s was not successful after %d retries", c.Command, numOfRetries)
 	return nil, err
 }
 
-// GetStdOut returns an io.Writer for exec with the defined prefix
-// Cannot be used with cmd.CombinedOutput()
+// execute executes the cmd with context canceled after commandTimeout seconds.
+// Returns error if unsuccessful, nil otherwise.
+func (c *Cmd) execute(i, numOfRetries, commandTimeout int) error {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(commandTimeout)*time.Second)
```
Despire, line 89: I'm not sure how do I feel about the timeout, the time that the command will take to execute depend on the size of the cluster and we can't really make any assumptions about that.

I've seen that you've set the timeout to 600 seconds (10 mins) if I request to create a large enough cluster there is a certainty that it won't be build due to exceeding the timeout threashold

MiroslavRepka: That's a good point. I think we have two options here

1. Disregard the `commandTimeout` for calls which are affected by the size/complexity of a cluster
2. Somehow scale the `commandTimeout` based on the size/complexity of a cluster

I lean toward the second option since we can scale that number by node pool size/number of node pools, however, I want to hear other opinions too.

Despire: if we wanted to keep the timeout, then we could benchmark clusters with different sizes to get the idea of how long does it take to execute each command and maybe base the number off of that, but I'm not sure how accurate would this be

Despire: This look much better, I think its fine for now 👍 , I agree with the timeout on kubectl commands which we can increase in the future if it proves to be a problem

Despire (approved):

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf (approved):

Issue #529, opened 2023-01-20
Secrets visible in logs


cloudziu:
When context-box is starting, mongoldb connection string has plain text visible login and password:
```
8:33AM INF ../go/services/context-box/server/claudieDB/claudieMongo.go:60 > Trying to ping the DB at mongodb://HERE-LOGIN:HERE-PASSWORD@mongodb:27017... module=context-box
8:34AM INF ../go/services/context-box/server/claudieDB/claudieMongo.go:60 > Trying to ping the DB at mongodb://HERE-LOGIN:HERE-PASSWORD@mongodb:27017... module=context-box
8:34AM INF ../go/services/context-box/server/claudieDB/claudieMongo.go:60 > Trying to ping the DB at mongodb://HERE-LOGIN:HERE-PASSWORD@mongodb:27017... module=context-box
8:34AM INF ../go/services/context-box/server/server.go:191 > Connected to database at mongodb://HERE-LOGIN:HERE-PASSWORD@mongodb:27017 module=context-box
```
I think that this should be hidden. ( or visible in DEBUG mode only )

MiroslavRepka:
We should review other logs in other services too, I have a feeling that there might be one or two more cases of this.

PR #535, opened 2023-01-24
bugfix: sanitise log/console secrets

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
fixes https://github.com/Berops/claudie/issues/529

list of issues fixed by this PR:
* [x] logging connection string URIs with exposed secrets
* [x] command (`kubectl`) logging that contains `kubeconfig` files with certificates


MiroslavRepka (commented):
LGTM, thanks @wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf. 

In the description for #529 I mention that there is a possibility that other credentials are in logs, more concretely in `command` package (i.e. when retrying kubectl command with kubeconfig inside). Could you have a look and resolved it as well?

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
> LGTM, thanks...

:rocket: 

> ...there is a possibility that other credentials are in logs ... when retrying kubectl command with kubeconfig inside...

as suggested, I tried fixing it at the source (the `command` package) as that's what all the `kubectl` methods are calling anyway.
let me know if I missed anything. @MiroslavRepka

internal/command/cmd.go
```diff
@@ -45,18 +46,26 @@ const (
 // Returns error if all retries fail, nil otherwise.
 func (c *Cmd) RetryCommand(numOfRetries int) error {
 	var err error
+
+	// Have a cmd that is safe for printing.
+	printSafeCmd := utils.SanitiseKubeconfig(c.Command)
```
MiroslavRepka, line 51: I am actually not sure about this. The command package is used for any command, not just kubectl. Therefore, I do not see a reason to have this function call here. 

What I would do instead, is maybe something like 
```go
func (c *Cmd) RetryCommand(numOfRetries int) error {
    printSafeCmd := c.sanitiseCommand()
    //.....
}

// sanitiseCommand returns c.Command with all secrets removed by ***
func (c *Command)sanitiseCommand() string {
    var s string
    //in case c.Command is kubectl, remove any possible kubeconfig.
    s = utils.SanitiseKubeconfig(c.Command)
    //in case c.Command is XXX, remove YYY
    s = utils.SanitiseXXX(s)
    //.....
    return s
}
```

internal/command/cmd.go
```diff
@@ -65,18 +74,26 @@ func (c *Cmd) RetryCommand(numOfRetries int) error {
 func (c *Cmd) RetryCommandWithOutput(numOfRetries int) ([]byte, error) {
 	var err error
 	var out []byte
+
+	// Have a cmd that is safe for printing.
+	printSafeCmd := utils.SanitiseKubeconfig(c.Command)
```
MiroslavRepka, line 79: ditto

MiroslavRepka (commented):
Good job :) 

I like the implementation, from a functional point of view, although, I left a comment, which we can discuss if needed.

internal/command/cmd.go
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf: got it, I see what you're saying.
I'll rework it, then.
to check for the actual cmd, we can just query arg[0] of `c.Args` and switch on it.

MiroslavRepka: >to check for the actual cmd, we can just query arg[0] of c.Args and switch on it.
yes, for the actual command, we will call `c.Command`, while sanitised will be returned from the function.. we can even add it to the `Command` struct, but I will leave this up to you.

internal/command/cmd.go
```diff
@@ -119,6 +146,34 @@ func (c *Cmd) buildCmd() (*exec.Cmd, context.CancelFunc) {
 	return cmd, cancelFun
 }
 
+func (c *Cmd) sanitisedCmd() string {
+	if c.Command == "" {
+		return c.Command
+	}
+
+	// split the command string on " ".
+	cmdStrings := strings.Split(c.Command, " ")
```
MiroslavRepka, line 155: I am not a fan of this approach. We are assuming that `kubectl`, or any other command we would want to sanitise in future would have to be first. I think we could do better.

What I had in mind originally would be to call `SanitiseXXX()` on every command, regardless if the command is `kubectl`, `terraform` or whatnot. If the secret regex would match, we would replace it, if not, we would just return the original command. Maybe we will be calling `SanitiseXXX()` unnecessarily, however, we would be sure that no secret would be leaked in the logs.

internal/command/cmd.go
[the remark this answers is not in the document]
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf: I really have, thanks.

internal/command/cmd.go
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf: > We are assuming that `kubectl`, or any other command we would want to sanitise in future would have to be first.

well..as long as this assumed behaviour is covered by tests..which it is.

> What I had in mind originally would be to call `SanitiseXXX()` on every command, regardless if the command is `kubectl`, `terraform` or whatnot. If the secret regex would match, we would replace it, if not, we would just return the original command. Maybe we will be calling `SanitiseXXX()` unnecessarily, however, we would be sure that no secret would be leaked in the logs.

I see. this approach has its own risks, however, for instance mangling the logs potentially by replacing inputs that were not meant to be passed in..
I can give it a rework, though.

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf: the way it is now we know what we're after and where and apply precise methods to achieve it.
with generic solutions it will definitely be easier to miss stuff.

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf: so to clarify, did you mean to run *any* command through all available sanitisation functions? I originaly got the idea that we were going to selectively choose what to sanitise and how.. @MiroslavRepka

MiroslavRepka: >did you mean to run any command

yes, I did. We could selectively choose which one to sanitise, however, the algorithm you have for it now is not enough. We can add commands like `kustomize build | kubectl --kubeconfig ..` later, which would not be caught by your solution. Instead of processing the commands to select which functions to call for sanitising, we can just sanitise it right away, even if we end up with the same command

internal/command/cmd.go
MiroslavRepka, line 51: ~`printSafeCmd := utils.SanitiseKubeconfig(c.Command)`~ -> `printSafeCmd := c.sanitisedCmd()`

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf: :see_no_evil:

MiroslavRepka (approved):
LGTM

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
thanks for the reviews @MiroslavRepka :rocket:

PR #536, opened 2023-01-24
Fix inequality of states after Claudie workflow

MiroslavRepka:
This PR fixes the `SaveConfigBuilder()` function. The issue was flipped condition for updating the desired state.


wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf (approved):
LGTM..

PR #538, opened 2023-01-25
bugfix: sort out ansible failing to install WG due to unattended upgrades

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
fixes https://github.com/Berops/claudie/issues/524

Issue #524, opened 2023-01-13
Ansible apt occasionally fails 


cloudziu:
```
8:11AM INF ../go/services/builder/serviceCaller.go:141 > Calling BuildInfrastructure on terraformer for project Scenario-1 module=builder
8:13AM INF ../go/services/builder/serviceCaller.go:163 > Calling InstallVPN on ansibler for project Scenario-1 module=builder
8:42AM INF ../go/services/builder/serviceCaller.go:168 > Calling InstallNodeRequirements on ansibler for project Scenario-1 module=builder
```
One of the playbooks was retrying for about 30 min until it get into working state. The cause of this was ansible `apt` module witch tends to occasionally fail while getting the lock file and/or while connecting to repositories. I've encountered a 30min ansible play:
 - On first run three nodes failed
 - On second run two node failed
 - On third and fourth a single node
The 5th run executed without issues.  

Example logs:
```
fatal: [w-oci-2]: FAILED! => {"ansible_facts": {"discovered_interpreter_python": "/usr/bin/python3.8"}, "changed": false, "msg": "Failed to lock apt for exclusive operation: Failed to lock directory /var/lib/apt/lists/: E:Could not get lock /var/lib/apt/lists/lock. It is held by process 1246 (apt)"}
```

```
[...]

PR #542, opened 2023-01-27
Disable apt fix

MiroslavRepka:
This PR disables the playbook added in PR #538. The playbook does not work due to errors like `'ansible_distribution' is undefined`. I tried to simply enable `gather_facts`, however, the errors were still present. 

 cc @wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf

### DIFF
--- a/manifests/claudie/kustomization.yaml
+++ b/manifests/claudie/kustomization.yaml
@@ -21,7 +21,7 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: ghcr.io/berops/claudie/ansibler
-  newTag: 69fb5b3-1339
+  newTag: af6519f-1341
 - name: ghcr.io/berops/claudie/builder
   newTag: 69fb5b3-1339
 - name: ghcr.io/berops/claudie/context-box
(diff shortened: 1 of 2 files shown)


MiroslavRepka:
Example of the full log
```
VPN - services/ansibler/server/clusters/ts2-c-2-8cyx    fatal: [azure-lb-1]: FAILED! => {"msg": "The conditional check 'ansible_distribution == 'Ubuntu' and ansible_distribution_version == '22.04'' failed. The error was: error while evaluating conditional (ansible_distribution == 'Ubuntu' and ansible_distribution_version == '22.04'): 'ansible_distribution' is undefined\n\nThe error appears to be in '/bin/services/ansibler/server/ansible-playbooks/wireguard/tasks/kill_unattended_upgrades.yml': line 2, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n---\n- name: \"Disable unattended (unintended) upgrades\"\n  ^ here\n"}
```

MarioUhrik (approved):

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
oh no, my bad, sorry guys..

perhaps the conditional could just have been disabled temporarily for a quick fix, since right now we only officially support Ubuntu 22.04 anyway.

I'll rework the conditional, I think I have an idea what exactly the issue was...

MarioUhrik:
> oh no, my bad, sorry guys..
> 
> perhaps the conditional could just have been disabled temporarily for a quick fix, since right now we only officially support Ubuntu 22.04 anyway.
> 
> I'll rework the conditional, I think I have an idea what exactly the issue was...

No worries. Please open a new PR as per the comment in https://github.com/Berops/claudie/issues/524

Issue #472, opened 2022-12-27
Feature: Storage classes based on user defined providers


MiroslavRepka:
### Motivation:

While writing some docs, it occurred to me that the storage classes we define for longhorn are based on used providers, not user-defined providers. What it means is that if the user defines the `gcp1` and `gcp2` providers, Claudie will only create the `longhorn-gcp-zone` storage class. However, it would make more sense to create two storage classes, like `longhorn-gcp1-zone` and `longhorn-gcp2-zone`.

### Description:

Storage classes should be created based on user defined providers.

### Exit criteria:
- [ ] Make appropriate changes in `Kuber`
- [ ] Update the storage docs to clarify how are storage classes defined.

### Note 

This should be very simple fix. It should be enough to change the way we get sorted nodepools [here](https://github.com/Berops/claudie/blob/master/services/kuber/server/longhorn/longhorn.go#L62) from calling `utils.GroupNodepoolsByProvider()` to `utils.GroupNodepoolsByProviderSpecName()`.

Issue #425, opened 2022-11-16
Improve HA of Longhorn PVs via node labels


cloudziu:
### Motivation:

Currently, Longhorn PV data is replicated with complete disregard to zone/region/provider.
In order to have more "HA" storage, we can provide Longhorn with the necessary information to schedule replica placements better.

### Description:

Longhorn when scheduling PVC replication, lookups [Kubernetes well-known labels](https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesioregion) `topology.kubernetes.io/zone`. Nodes that don’t belong to any zone will be treated as if they belong to the same zone. More info on [Longhorn scheduling policy](https://longhorn.io/docs/archives/1.0.1/volumes-and-nodes/scheduling/#scheduling-policy). To leverage its capabilities - of maintaining replicas in different zones, Kubernetes Nodes should have labels with their respective zone. 
We propose the following labels.

Hetzner Node:
```
topology.kubernetes.io/zone: fcs1
topology.kubernetes.io/region: eu-central
claudie.io/cloudprovider: hetzner
```

AWS Node:
```
topology.kubernetes.io/zone: us-east-1c
topology.kubernetes.io/region: us-east-1
claudie.io/cloudprovider: aws
```

### Exit criteria:
- [x] Nodes are now labeled as suggested
- [ ] We've experimentally ensured that Longhorn tries to spread replicas across zones

cloudziu:
Hi @MarioUhrik, thank for your input. 
The way I thought it would look like: 
Labels have only their atomic values. If we want to add additional information, it would be added in different labels. In case there are already described in [well-known labels](https://kubernetes.io/docs/reference/labels-annotations-taints/,) we should use them. Otherwise, I suggest to extend with Claudie specific labels. 
Example: if we want to add cloud provider/zone/region information:
Hetzner Node:
`topology.kubernetes.io/zone: fcs1`
`topology.kubernetes.io/region: eu-central`
`claudie.io/cloudprovider: hetzner`

AWS Node:
`topology.kubernetes.io/zone: us-east-1c`
`topology.kubernetes.io/region: us-east-1 `
`claudie.io/cloudprovider: aws`

cloudziu:
[...]

PR #543, opened 2023-01-27
Feat/node labels

MiroslavRepka:
This PR adds default node labels as discussed. All labels are added via `kubeone` so the changes in our code are minimal.

All nodes will have labels 
```yaml
claudie.io/worker-node: # true or false
claudie.io/provider: # cloud provider name [hetzner, gcp, oci, ...]
claudie.io/provider-instance: # cloud provider instance [hetzner-1, gcp-1, gcp-2, ...]
claudie.io/nodepool: # nodepool name

topology.kubernetes.io/region: # region name
topology.kubernetes.io/zone: # zone name
```

The storage classes in longhorn are now configured based on cloud provider instance, instead of cloud provider as described in #472. 
The topology labels also improve Longhorn scheduling as described in #425.


Despire (approved):
I see no issues, LGTM.

PR #546, opened 2023-01-31
remove arm image building from e2e

### LINKED ISSUE #497: Feature: Shorten the time to build multi-arch images
### Motivation:

Waiting ~1h 20 min for the images to build is a bit long, with the CI pipeline taking more than 2 hours to complete,  shortening the building phase of the CI could decrease the time for the CI to complete substantially while also given faster feedback to the changes made in the codebase.

### Description:

When making changes in only a single service the build times are reasonable, around ~10 mins, however, if any source file is changed that is common to all services they all need to be rebuild and the build times are at ~1h. When the E2E tests also fail making a fix for the issue repeats the cycle where the images need to be build waiting another ~1hour before the E2E tests run.
### Exit criteria:
- [ ] Shorten the time for building multi-arch images

Despire:
closes #497


MiroslavRepka (approved):
LGTM, thanks @Despire!

PR #557, opened 2023-02-02
Fix/oci iptables

### LINKED ISSUE #466: Bug: `iptables` reset after reboot and block all trafic on OCI node
### Current Behavior:

While debugging, I had to reboot an OCI node. While doing so, I noticed I lost connectivity to all other VMs and when I investigated, I noticed the `iptables` have reset after reboot, blocking the traffic.

### Expected Behavior:

[Changes we made](https://github.com/Berops/claudie/blob/master/services/terraformer/templates/oci.tpl#L128) to `iptables` during VM creation should be persistent.

### Steps To Reproduce:
1. Create an OCI node
2. Reboot the VM after creation
3. Check `iptables` and connectivity of the node

Despire:
closes #466 

The real issue was with the IP for wg0 not persistent across reboots, but also made the iptables persistent


MiroslavRepka (approved):
LGTM, thanks @Despire!

PR #559, opened 2023-02-03
Update go version to `1.20` and some dependencies

MiroslavRepka:


Despire:
missed https://github.com/Berops/claudie/blob/master/go.mod#L3

Despire (approved):

PR #569, opened 2023-02-07
no error returned when deletion of secrets fail

### LINKED ISSUE #527: Bug: Deletion of failed input manifests sometimes errors out
### Current Behavior:

When an input manifest has a mistake, i.e. a wrong VM shape, the terraformer fails with an error. Later, when the user deletes a secret, the input manifest will go through the whole deletion process, meaning it will try to remove secrets about cluster metadata and kubeconfig, which does not exist as the error occurred before they could be created. This will result in an error, even though it should be successful.

### Expected Behavior:

When metadata/kubeconfig secrets were not created, deletion should not throw an error when it will try to delete them.

### Steps To Reproduce:

1. Deploy Claudie to the cluster
2. Create an input manifest with a mistake
3. Insert it into the Claudie
4. After it errors out, triggers deletion
5. See the error about failing to delete metadata/kubeconfig secret

### Anything else to note:

Please, keep the fix simple. IMO, if Kuber fails to delete secrets about kubeconfig/metadata, it should not result in an error, max a warning about it.

Despire:
closes #527

### DIFF
--- a/manifests/claudie/kustomization.yaml
+++ b/manifests/claudie/kustomization.yaml
@@ -31,7 +31,7 @@ images:
 - name: ghcr.io/berops/claudie/kube-eleven
   newTag: c7f8a92-1392
 - name: ghcr.io/berops/claudie/kuber
-  newTag: c7f8a92-1392
+  newTag: 381614f-1400
 - name: ghcr.io/berops/claudie/scheduler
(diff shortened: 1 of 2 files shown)


MiroslavRepka (approved):
LGTM

PR #571, opened 2023-02-07
Fix/cidr

### LINKED ISSUE #558: Bug: Respect CIDR format
### Current Behavior:

The manifest allows to specify network using CIDR format `<ip-address>/<mask>` however we are not respecting the mask in code and always use `/24`

more occurrences can be found in the following links:
https://github.com/Berops/claudie/blob/master/services/ansibler/server/ansible-playbooks/wireguard/tasks/configure.yml#L32
https://github.com/Berops/claudie/blob/master/services/ansibler/server/vpn.go#L78-L118

### Expected Behavior:

We should either 1. respect the supplied mask that was specified in the input manifest or 2. always use `/24`

Despire:
closes #558

### DIFF
--- a/services/ansibler/server/ansible-playbooks/wireguard/templates/wg.conf.j2
+++ b/services/ansibler/server/ansible-playbooks/wireguard/templates/wg.conf.j2
@@ -7,7 +7,7 @@ ListenPort = {{ wg_listen_port }}
 [Peer]
 PublicKey = {{ hostvars[host].publickey.stdout }}
 Endpoint = {{ hostvars[host].ansible_host }}:{{ wg_listen_port }}
-AllowedIps = {{ hostvars[host].private_ip }}/32
+AllowedIps = {{ hostvars[host].private_ip }}/{{ netmask}}
 PersistentKeepalive = 60
 {% endif %}
 {% endfor %}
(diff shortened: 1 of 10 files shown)


services/ansibler/server/ansible-playbooks/wireguard/templates/wg.conf.j2
```diff
@@ -7,7 +7,7 @@ ListenPort = {{ wg_listen_port }}
 [Peer]
 PublicKey = {{ hostvars[host].publickey.stdout }}
 Endpoint = {{ hostvars[host].ansible_host }}:{{ wg_listen_port }}
-AllowedIps = {{ hostvars[host].private_ip }}/32
+AllowedIps = {{ hostvars[host].private_ip }}/{{ netmask}}
```
bernardhalas, line 10: When discussed during the last grooming, I was wondering how could've `/32` ever worked.

> The keyword allowed-ips is a list of addresses that will get routed to the peer.

Ref: https://wiki.archlinux.org/title/WireGuard

MiroslavRepka (approved):
LGTM, thanks @Despire

services/ansibler/server/ansible-playbooks/wireguard/templates/wg.conf.j2
Despire: I was thinking the same but turns out the `/32` is needed (I misunderstood. the CIDR notation)

The `/32` means only a single IP address whereas anything lower than that means a range of IP addresses,
And if we have multiple clients with the same network `/32` is needed as otherwise wireguard couldn't resolve to which client to send the packets

more: 
https://www.reddit.com/r/WireGuard/comments/enh9xa/help_on_24_and_32_when_using_as_a_vpn_server/
https://www.wireguard.com/#cryptokey-routing

Issue #352, opened 2022-10-28
Use Cloud tags on Claudie-made infrastructure


MarioUhrik:
It would be convenient if the Claudie-made infrastructure in the Cloud had some tags.
This would make it distinguishable from non-Claudie infrastructure.

The goal of this task is to modify the TF templates for all providers, such that all resources are tagged.

Example tags:
- `claudie`
- `claudie-${PROJECTNAME}`

AWS Examples:
- how to tag a resource in TF https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/resource-tagging
- "mass" tagging https://registry.terraform.io/providers/hashicorp/aws/latest/docs#default_tags

PR #579, opened 2023-02-13
Feat/cloud labels

MiroslavRepka:
This PR implements #352. While doing so, I also formatted/unified the templates to be more readable as templates and terraform files after generation. 

Other chores:
- Update of the provider versions (only minor version changed)
- Small bugfix in the `command` package

All of the templates were tested locally to ensure they work as expected.


services/terraformer/templates/gcp-lb.tpl
```diff
@@ -50,39 +51,45 @@ resource "google_compute_firewall" "firewall-{{ $region }}" {
...
+    subnetwork = google_compute_subnetwork.{{ $nodepool.Name }}_subnet.self_link
     access_config {}
   }
   metadata = {
     ssh-keys = "root:${file("./public.pem")}"
   }
   metadata_startup_script = "echo 'PermitRootLogin without-password' >> /etc/ssh/sshd_config && echo 'PubkeyAuthentication yes' >> /etc/ssh/sshd_config && service sshd restart"
+
+  labels = {
+    environment = "claudie"
```
Despire, line 86: "Managed by claudie" wouldn't work here ?

Despire (approved):
LGTM, thanks @MiroslavRepka

MarioUhrik (commented):
Hi Miro, nice work

I think we should leave the `environment` tag unused.
It's probably the most commonly used tag, people may want to control it themselves.
Instead, something like `managedBy: claudie` may work better.

Also, how about project/cluster specific cloud tags?
e.g. having both:
- `managedBy: claudie`
-  `claudieCluster: $CLUSTERNAME`
-  `claudieProject: $PROJECTNAME`

services/terraformer/templates/gcp-lb.tpl
MiroslavRepka: The GCP labels have a strict requirements, like no capitals, no white spaces. We can do `managed-by-claudie` though.

MiroslavRepka:
Hi @MarioUhrik, thanks for the feedback. I added `cluster name` to the tags/labels/descriptions, however, I did not add `project name`, as that would require some changes in the code. If you think we want to have it there, it can be done. :slightly_smiling_face:

MarioUhrik (approved):
Thanks Miro, I'm happy with this :)

PR #584, opened 2023-02-16
Feat/cluster processing

### LINKED ISSUE #477: Feature: Process input manifest on per cluster basis for the whole Claudie process, not just for a single service
### Motivation:

Currently, when the user inputs a manifest, claudie will process it as a single unit, even if there are N clusters defined in it. Although there is per-cluster processing deployed on almost every service, the issue arises when at least one of the clusters encounters an error. What happens is, that error in a single cluster will stop the processing of the whole input manifest.

Example: 
- The user defines a single input manifest, with three clusters.
- Terraformer will build the first two clusters, but the third one contains an error, i.e. a wrong VM shape.

The whole workflow is now stopped, as there was an error, however, the first two clusters could be good to continue to the next stage.

Furthermore, I saw a few times some clusters just idling in services, while waiting for one, which was still being processed. The time they spend idling could have been used to continue processing.

### Description:

Claudie should process the input manifest on a per-cluster basis.

### Exit criteria:
- [ ] Claudie is able to process input manifest on a per-cluster basis. 

### Note

I had some time to think this through, so here is my high-level concept of handling it.

Instead of passing the whole desired state to the Terraformer, Ansibler, ... we could change their API to accept only a "single cluster" (or any other data regarding a single cluster), which they would process. 

It would be up to the Builder to call those services and handle the concurrent execution of the clusters. Once all clusters would finish processing, either successfully or not, Builder would return. This would assure that Claudie would try to build every cluster regardless of errors in different clusters and it could save time as some clusters can be built faster than others.

Despire:
closes #477 

This PR implements a per-cluster processing of the config. Since while building the cluster we also need the loadbalancers at some points I didn't seperate the workflow for Lbs from the kubernetes clusters.

The idea is that everything a cluster needs during a workflow execution is stored in the [BuildContext](https://github.com/berops/claudie/blob/50758ba6bbd19b71088cbd1717fecf4e7c8ef34c/services/builder/serviceCaller.go#L15). After a successfull workflow the changes made within that context are then propagated [back](https://github.com/berops/claudie/blob/50758ba6bbd19b71088cbd1717fecf4e7c8ef34c/services/builder/configProcessor.go#L134)

To be able to process the cluster individually I had to create another "view" that would just aggregate the data per cluster thus the [ClusterView struct](https://github.com/berops/claudie/blob/50758ba6bbd19b71088cbd1717fecf4e7c8ef34c/services/builder/cluster_view.go#L11)

Any changes made are than propagated back to the [config](https://github.com/berops/claudie/blob/50758ba6bbd19b71088cbd1717fecf4e7c8ef34c/services/builder/configProcessor.go#L147) and saved to the DB


proto/ansibler.proto
```diff
@@ -4,32 +4,41 @@ import "proto/config.proto";
...
+  repeated LBcluster desiredLbs = 2;
 }
 
-message SetUpLBRequest { 
-  Project desiredState = 1;
-  Project currentState = 2;
-  map<string, string> oldApiEndpoints = 3;
+message SetUpLBRequest {
+  K8scluster desired = 1;
+  repeated LBcluster lbs = 2;
```
MiroslavRepka, line 19: Are these current state LBs? If so, can we name them like that?

proto/ansibler.proto
```diff
@@ -4,32 +4,41 @@ import "proto/config.proto";
...
+  repeated LBcluster desiredLbs = 3;
+  string previousAPIEndpoint = 4;
+  string name = 5;
 }
 message SetUpLBResponse {
-  Project desiredState = 1;
+  K8scluster desired = 1;
+  repeated LBcluster lbs = 2;
```
MiroslavRepka, line 26: ditto

proto/terraformer.proto
```diff
@@ -5,18 +5,34 @@ import "proto/config.proto";
 option go_package = "proto/pb";
 
 message BuildInfrastructureRequest {
-  Project currentState = 1;
-  Project desiredState = 2;
+  K8scluster current = 1;
+  K8scluster desired = 2;
+
+  repeated LBcluster lbs = 3;
```
MiroslavRepka, line 11: ditto

proto/terraformer.proto
```diff
@@ -5,18 +5,34 @@ import "proto/config.proto";
...
 }
 
 message BuildInfrastructureResponse {
-  Project currentState = 1;
-  Project desiredState = 2;
-  string errorMessage = 3;
+  K8scluster current = 1;
+  K8scluster desired = 2;
+
+  repeated LBcluster lbs = 3;
```
MiroslavRepka, line 21: ditto

proto/ansibler.proto
```diff
@@ -4,32 +4,41 @@ import "proto/config.proto";
 
 option go_package = "proto/pb";
 
-message InstallRequest { 
-  Project desiredState = 1;
-  Project currentState = 2; 
+message InstallRequest {
+  K8scluster desired = 1;
+  repeated LBcluster desiredLbs = 2;
+  string name = 3;
```
MiroslavRepka, line 10: Can we rename it to `ProjectName`, so it's more noticeable?

services/builder/cluster_view.go
```diff
@@ -0,0 +1,160 @@
...
+
+import (
+	"github.com/Berops/claudie/proto/pb"
+	"google.golang.org/protobuf/proto"
+)
+
+// ClusterView contains the per-cluster view on a given config.
+// No mutex is needed when processing concurrently as long as each cluster only
+// works with related values.
+type ClusterView struct {
```
MiroslavRepka, line 11: I really like this approach, great job!

services/builder/utils.go
```diff
@@ -0,0 +1,66 @@
...
+		})
+	}); err != nil {
+		return err
+	}
+
+	return cbox.DeleteConfigFromDB(c, config.Id, pb.IdType_HASH)
+}
+
+// destroy destroys any Loadbalancers or the cluster itself.
+func destroy(projectName, clusterName string, clusterView *ClusterView) (bool, error) {
```
MiroslavRepka, line 26: What is the returned boolean value for?

services/ansibler/server/nodeRequirments.go
```diff
@@ -13,35 +13,29 @@ const (
...
-		if _, err := os.Stat(directory); os.IsNotExist(err) {
-			if err := os.MkdirAll(directory, os.ModePerm); err != nil {
-				return fmt.Errorf("failed to create directory %s : %w", directory, err)
-			}
-		}
-		if err := utils.CreateKeyFile(nodepoolInfo.PrivateKey, directory, fmt.Sprintf("%s.%s", nodepoolInfo.ID, privateKeyExt)); err != nil {
-			return fmt.Errorf("failed to create key file for %s : %w", nodepoolInfo.ID, err)
+	directory := filepath.Join(baseDirectory, outputDirectory, utils.CreateHash(utils.HashLength))
+
+	if _, err := os.Stat(directory); os.IsNotExist(err) {
```
MiroslavRepka, line 20: I see these functions used in many places, would it make sense to move them to `utils`, i.e. `utils.CreateDir()`?

MiroslavRepka (commented):
LGTM, great job @Despire! I left some comments, just minor changes if anything.

MiroslavRepka (approved):

Issue #574, opened 2023-02-09
Chore: Remove CD-pipeline-dev.yml


Despire:
### Description

Running the pipeline for E2E tests twice, once when the Pull Request is opened and then again after it's merged into master is unnecessary, and given that we have different quotas across cloud providers may block another PR's e2e tests
### Exit criteria

- [ ] Delete `workflows/CD-pipeline-dev.yml`

Despire:
The badge need to be adjusted

PR #585, opened 2023-02-17
Remove the CD pipeline

MarioUhrik:
Implements https://github.com/berops/claudie/issues/574

We have nothing left that would be sensible to put into the badge, so we're removing the badge


Despire (approved):
LGTM

PR #594, opened 2023-02-20
reduce hetzner usage in e2e tests

### LINKED ISSUE #575: Chore: Seperate Hetzner from tests-sets
### Description

We've alot of issues with hetzner when the e2e pipeline receives IP's that were misused in some form. To name a few, docker images may fail to download on the build cluster, IP may be banned from certain endpoints that are part of the building workflow and thus the workflow fails etc...

Seperate hetzner from the test-sets into its own "job",  so that if we receive such IP in our e2e pipeline we can still determine if the changes made were breaking or not by looking at the test-set that didn't include hetzner

### Exit criteria

- [ ] Seperate Hetzner (possibly by including a seperate test-set or creating a new stage in the e2e pipeline)

Despire:
Closes #575 by reducing the counts and spreading across multiple hetzner datacenters


MiroslavRepka (approved):
LGTM, thanks @Despire!

PR #597, opened 2023-02-20
utils: extend the kubeconfig-sanitising regex+test

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
based on a short conversation with @MiroslavRepka in Slack, it became clear that another form of the kubeconfig passed as an argument to `kubectl`'s `--kubeconfig` flag needed accomodating, specifically the following:
```sh
--kubeconfig <(echo 'the real kubeconfig')
```

this PR does exactly that, without breaking the old behaviour - i.e. the regex has just been extended, so that it matches more cases, e.g. double quoted `("blah")` strings are now also handled.

the tests have also been extended to cover the new paths.

ref pic:
image

### DIFF
--- a/.github/workflows/CI-pipeline.yml
+++ b/.github/workflows/CI-pipeline.yml
@@ -264,28 +264,28 @@ jobs:
             for file in $directory/*
             do
             #replace hetznerdns token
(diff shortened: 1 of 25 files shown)


MiroslavRepka (approved):
LGTM

PR #601, opened 2023-02-21
unify manifest naming

### LINKED ISSUE #563: Chore: Unify the naming scheme in input manifest
### Description

This issue aims to unify the naming scheme in the input manifest. Currently, we use camelCase (`nodePool`, `providerSpec`), snake_case (`server_type`, `disk_size`) and kebab-case (`targeted-k8s`). We should really choose one and stick with it. 

I propose using camelCase everywhere.

### Exit criteria

- [ ] New naming scheme was chosen and implemented to input manifest
- [ ] Docs have been updated accordingly

Despire:
closes #563

### DIFF
--- a/.github/workflows/CI-pipeline.yml
+++ b/.github/workflows/CI-pipeline.yml
@@ -264,28 +264,28 @@ jobs:
             for file in $directory/*
             do
             #replace hetznerdns token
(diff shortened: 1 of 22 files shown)


MiroslavRepka (approved):
LGTM, although I have not verified if there is anything else that we need to change (other docs, etc...). But from the top of my head, that should be it. Thanks, @Despire.

PR #607, opened 2023-02-24
add lb tests

### LINKED ISSUE #419: Feature: Add E2E tests for LB create/update/delete
### Description:

Add E2E tests for changes introduced in #371 
### Exit criteria:
- [ ] Add E2E tests

Despire:
closes #419

### DIFF
--- a/manifests/testing-framework/kustomization.yaml
+++ b/manifests/testing-framework/kustomization.yaml
@@ -15,6 +15,8 @@ secretGenerator:
 - files:
   - test-sets/test-set3/1.yaml
   - test-sets/test-set3/2.yaml
+  - test-sets/test-set3/3.yaml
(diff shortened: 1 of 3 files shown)


MiroslavRepka (approved):
LGTM, thank you @Despire! 

Also, as you had some issues with a typo in roles, can we add some checks to validation, which would save users and us the trouble next time this happens? (It can be just logged issue)

Despire:
It handled by the validation I just missed it (because I wasn't looking at the frontend logs but at the k8s-sidecar logs) so just a mistake from my side

MiroslavRepka:
>It handled by the validation I just missed it (because I wasn't looking at the frontend logs but at the k8s-sidecar logs) so just a mistake from my side

I see, never mind then :slightly_smiling_face:

PR #648, opened 2023-03-20
Feat/longhorn replicas

### LINKED ISSUE #564: Feature: Automatically manage number of longhorn replicas upon node deletion.
### Motivation

Currently, when a node contains a longhorn volume and we delete that node, a new replica of the volume is not created. The longhorn devs [advise](https://github.com/longhorn/longhorn/issues/1496#issuecomment-644729797) to temporarily increase the replica count, so a new one is scheduled and then decrease it back to the desired number. By doing this, we assure the volume will be accessible as long as there are any nodes assigned to that storage class. There will be a need for some additional logic, like checking if a new replica can be even created, or maybe deleting the old replica (from the deleted node).

### Description

Claudie should try to manage longhorn volume replicas when deleting a node from the cluster.

### Exit criteria

- [ ] The replicas are now created whenever possible on a node deletion.
- [ ] The docs will document how node deletion work in regard to the storage.

MiroslavRepka:
This PR fixes #564 by implementing the algorithm, run upon node deletion, which tries to create a new replica for the longhorn volume, if any replicas will be lost due to node deletion.

It also fixes a small bug, when deleting master nodes (case when the API endpoint was not defined).

An additional feature is the introduction of variadic functions for the `kubectl` package.


MarioUhrik (approved):
LGTM, if you've sufficiently tested that the new behavior is as you expect

services/kuber/server/nodes/nodes.go
```diff
@@ -158,27 +167,79 @@ func (d *Deleter) updateClusterData() {
...
-			return fmt.Errorf("error while deleting node %s from nodes.longhorn.io : %w", nodeName, err)
+func deleteFromLonghorn(kc kubectl.Kubectl, worker string) error {
+	log.Info().Msgf("Deleting node %s from nodes.longhorn.io", worker)
+	if err := kc.KubectlDeleteResource("nodes.longhorn.io", worker, "-n", longhornNamespace); err != nil {
+		return fmt.Errorf("error while deleting node %s from nodes.longhorn.io : %w", worker, err)
+	}
+	return nil
+}
+
+// assureReplication tries to assure, that replicas for each longhorn volume are migrated to nodes, which will remain in the
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 178: I cannot figure out the ending of this sentence..
regardless, it should probably be added.

services/kuber/server/nodes/nodes.go
```diff
@@ -158,27 +167,79 @@ func (d *Deleter) updateClusterData() {
...
+					return fmt.Errorf("error while increasing number of replicas in volume %s from cluster : %w", v.Metadata.Name, err)
+				}
+				// Wait pvcReplicationTimeout for Longhorn to create new replica.
+				log.Info().Msgf("Waiting %d seconds for new replicas to be scheduled if possible for node %s", pvcReplicationTimeout, worker)
+				time.Sleep(pvcReplicationTimeout * time.Second)
+				// Decrease number of replicas in volume -> original state.
+				if err := revertReplicaCount(v, kc); err != nil {
+					return fmt.Errorf("error while increasing number of replicas in volume %s : %w", v.Metadata.Name, err)
+				}
+				// Delete old replica, on to be deleted node.
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 204: `Delete old replica, on to be deleted node.` -> `Delete old replica on a to-be-deleted node.`

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf (changes requested):
had a look, looks great, had some questions/suggestions.

internal/kubectl/kubectl.go
```diff
@@ -27,55 +28,39 @@ const (
...
+// example: kubectl apply -f test.yaml -> k.KubectlApply("test.yaml")
+// example: kubectl apply -f test.yaml -n test -> k.KubectlApply("test.yaml", "-n", "test")
+func (k *Kubectl) KubectlApply(manifest string, options ...string) error {
 	kubeconfig := k.getKubeconfig()
-	if namespace != "" {
-		namespace = fmt.Sprintf("-n %s", namespace)
-	}
-	command := fmt.Sprintf("kubectl apply -f %s %s %s", manifest, kubeconfig, namespace)
-	return k.run(command)
+	command := fmt.Sprintf("kubectl apply -f %s %s", manifest, kubeconfig)
```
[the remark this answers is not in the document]
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf: that's right, my bad :)

cloudziu (approved):

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf (approved):
great stuff, LGTM, thanks @MiroslavRepka! :rocket:

PR #657, opened 2023-03-23
Log rework

### LINKED ISSUE #622: Feature: Supress logs from tools other than Claudie in `info` level, but show them in `debug` level.
### Motivation

The number of logs Claudie is producing can be quite high. All of the tools which are used internally are producing them in different styles, quantities and levels. It would be nice if the `info` level of logging in Claudie would suppress the logging from other tools. It would look cleaner, would produce significantly fewer logs and would still keep the basic level of information, as Claudie is also logging the success/failure of the steps while building the clusters. 

### Description

Logs from tools like `Ansible`, `Terraform`, or `KubeOne` should be suppressed in the `info` level, but enabled in the `debug` level of logging.

### Exit criteria

- [ ] Logs, other than Claudie ones are disabled in the `info` level and enabled in the `debug` level.

MiroslavRepka:
This PR implements a feature described in #622. It also clarifies/adds some additional logs, as in the `info` level, a lot of information is no longer visible.

Moreover, in the `debug` level, the `kubectl` logs are now visible, with cluster ID prefix.
Example:
```
2:48PM INF services/kuber/server/server.go:60 > Setting up the longhorn on the cluster miro-cluster-s87luix module=kuber
miro-cluster-s87luix     namespace/longhorn-system created
miro-cluster-s87luix     podsecuritypolicy.policy/longhorn-psp created
miro-cluster-s87luix     serviceaccount/longhorn-service-account created
miro-cluster-s87luix     configmap/longhorn-default-setting created
miro-cluster-s87luix     configmap/longhorn-storageclass created
```

closes #622


services/builder/configProcessor.go
```diff
@@ -68,7 +70,7 @@ func configProcessor(c pb.ContextBoxServiceClient, wg *sync.WaitGroup) error {
 			// Check if we need to destroy the cluster or any Loadbalancers
 			done, err := destroy(config.Name, clusterName, clusterView)
 			if err != nil {
-				log.Error().Msgf("failed to destroy cluster %s project %s: %s", clusterName, config.Name, err)
+				log.Error().Msgf("Error while to destroy cluster %s project %s : %v", clusterName, config.Name, err)
```
cloudziu, line 73: I'm not sure if this is a proper sentence 😅

services/builder/configProcessor.go
```diff
@@ -92,20 +94,20 @@ func configProcessor(c pb.ContextBoxServiceClient, wg *sync.WaitGroup) error {
 				}
 
 				if ctx, err = buildCluster(ctx); err != nil {
-					log.Error().Msgf("failed to build cluster %s project %s: %s", clusterName, config.Name, err)
+					log.Error().Msgf("Failed to build cluster %s project %s : %v", clusterName, config.Name, err)
```
cloudziu, line 97: You kept `failed to destroy cluster` here and changed  the previous one to `Error while to destroy(ing)` . Might be wrong but I think we logs should keep the logs in a consistent in form.

services/builder/serviceCaller.go
```diff
@@ -210,50 +221,55 @@ func callKuber(ctx *BuilderContext) error {
...
+	log.Info().Msgf("StoreNodeMetadata on Kuber for cluster %s project %s finished successfully", ctx.GetClusterName(), ctx.projectName)
 
 	return nil
 }
 
 // destroyConfig destroys existing clusters infra for a config by calling Terraformer and Kuber
 func destroyCluster(ctx *BuilderContext) error {
 	if err := destroyConfigTerraformer(ctx); err != nil {
-		return fmt.Errorf("error in destroy config terraformer for config %s project %s : %w", ctx.GetClusterName(), ctx.projectName, err)
+		return fmt.Errorf("error in destroy config Terraformer for config %s project %s : %w", ctx.GetClusterName(), ctx.projectName, err)
```
cloudziu, line 272: Same as previous, `error in destroy...`

services/kuber/server/longhorn/longhorn.go
```diff
@@ -175,7 +182,7 @@ func (l *Longhorn) deleteOldStorageClasses(existing, applied []string, kc kubect
 		//if not found in applied, delete the sc
 		if !found {
 			err := kc.KubectlDeleteResource("sc", ex, "")
-			log.Info().Msgf("Deleting storage class %s", ex)
+			log.Debug().Msgf("Deleting storage class %s", ex)
```
cloudziu, line 185: I think that log about removing a StorageClass would be usefull for user, not only in Debug mode. WDYT?

services/kuber/server/nodes/nodes.go
```diff
@@ -85,7 +92,7 @@ func deleteNodesByName(kc kubectl.Kubectl, nodesToDelete, realNodeNames []string
 	for _, nodeName := range nodesToDelete {
 		realNodeName := utils.FindName(realNodeNames, nodeName)
 		if realNodeName != "" {
-			log.Info().Msgf("Deleting node %s from k8s cluster", realNodeName)
+			log.Debug().Msgf("Deleting node %s from k8s cluster", realNodeName)
```
cloudziu, line 95: SImillary here, I would keep it `Info`

services/kuber/server/nodes/nodes.go
```diff
@@ -129,11 +134,10 @@ func (d *Deleter) deleteFromEtcd(kc kubectl.Kubectl) error {
 	for _, nodeName := range d.masterNodes {
 		for _, etcdPodInfo := range etcdPodInfos {
 			if nodeName == etcdPodInfo.nodeName {
-				log.Info().Msgf("Deleting etcd member %s, with hash %s ", etcdPodInfo.nodeName, etcdPodInfo.memberHash)
+				log.Debug().Msgf("Deleting etcd member %s, with hash %s ", etcdPodInfo.nodeName, etcdPodInfo.memberHash)
```
cloudziu, line 137: Same here - `Info` level

cloudziu (commented):
Great job with the PR. I left few comments, some of them are regarding the log `linguistic correctness` 😅 - but I think this can be a completly separate task, I will leave the decision to you. Others are about log level in some point that I'm not sure about. Once again awesome job! 😄

services/kuber/server/nodes/nodes.go
MiroslavRepka: This log is more useful when debugging than for a user IMO. It shows quite an "under the hood" message, and it also misses the information about which cluster are those nodes from. Thus I used the `debug` level while keeping the info about deleting a master node in the `info` level. :)

cloudziu: > Thus I used the debug level while keeping the info about deleting a master node in the info level. :)
Great thank you!
I think on this you might be right. You convinced me 👍

cloudziu (approved):
LGTM 👍

PR #660, opened 2023-03-27
Feat/terminate ci

### LINKED ISSUE #656: Chore: Terminate CI on testing-framework error
### Description

Currently, CI only terminates on two conditions
- When testing-framework is successfully completed
- On timeout

This means, that if testing-framework fails, CI has to wait, until the timeout occurs. This could be sped up, if CI would terminate on testing-framework failure as well (with error). 

### Exit criteria

- [ ] CI now terminates, when the testing-framework encounters an error.

MiroslavRepka:
Fixes #656

### DIFF
--- a/.github/workflows/CI-pipeline.yml
+++ b/.github/workflows/CI-pipeline.yml
@@ -308,7 +308,25 @@ jobs:
 
       - name: Monitor E2E test
         run: |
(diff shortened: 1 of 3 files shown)


MarioUhrik (approved):
LGTM as long as the E2E test can confirm it works as intended

MiroslavRepka:
When dynamo DB was not deployed into the claudie namespace, the testing-framework failed, which also failed CI after [~37min](https://github.com/berops/claudie/actions/runs/4531020779/jobs/7980692076) (the same time it took testing-framework to show an error). I would consider this a sufficient test.

PR #661, opened 2023-03-28
Feat/cidr caching

### LINKED ISSUE #647: Bug: Change in CIDR forces replacement
### Current Behaviour

Currently, the CIDR for subnets created in VPCs are calculated dynamically. For instance, in the case of AWS, each node pool in the same region has a subnet created based of their position in node pool slice.
```
{{- range $i, $nodepool := .NodePools }}
resource "aws_subnet" "{{ $nodepool.Name }}_subnet" {
  provider                = aws.k8s_nodepool_{{ $nodepool.Region  }}
  vpc_id                  = aws_vpc.claudie_vpc_{{ $nodepool.Region }}.id
  cidr_block              = "{{ getCIDR "10.0.0.0/24" 2 $i}}"        # This line uses index $i to define the CIDR for the subnet
...
```

The issue arises, when you change that slice of node pools, i.e. you remove a certain node pool from the input manifest. This then assigns a new CIDR to the subnet, which in the case of AWS forces replacement.
```tf
~ cidr_block = "10.0.1.0/24" -> "10.0.0.0/24" # forces replacement
```

As the subnet is being replaced, all resources that are using subnet ID are being replaced as well, which includes VMs themselves.
```tf
~ subnet_id      = "subnet-0ff64c9885134791c" # forces replacement -> (known after apply)
```

### Expected Behaviour

The deletion of a one node pool should not trigger the replacement of another.

### Steps To Reproduce

1. Start Claudie
2. Define two-step test case

`1.yaml`
```yaml
kubernetes:
  clusters:
    - name: test-cluster
      version: v1.24.0
      network: 192.168.2.0/24
      pools:
        control:
          - oci-control
          - aws-control      # Same providerSpec as aws-compute
        compute:
          - oci-compute
          - aws-compute  # Same providerSpec as aws-control
```
`2.yaml`
```yaml
kubernetes:
  clusters:
    - name: test-cluster
      version: v1.24.0
      network: 192.168.2.0/24
      pools:
        control:
          - oci-control
        compute:
          - oci-compute
          - aws-compute
```
3. Run the test cases
4. See the replacement of `aws-compute` in `terraformer` logs

MiroslavRepka:
Fixes #647


cloudziu (approved):
LGTM 👍

services/terraformer/templates/azure.tpl
```diff
@@ -103,7 +103,7 @@ resource "azurerm_subnet" "{{ $nodepool.Name }}_{{ $clusterHash }}_subnet" {
   name                 = "{{ $nodepool.Name }}_{{ $clusterHash }}_subnet"
   resource_group_name  = azurerm_resource_group.rg_{{ replaceAll $nodepool.Region " " "_"  }}.name
   virtual_network_name = azurerm_virtual_network.claudie_vn_{{ replaceAll $nodepool.Region " " "_" }}.name
-  address_prefixes     = ["{{getCIDR "10.0.0.0/24" 2 $i}}"]
+  address_prefixes     = ["{{index $.Metadata (printf "%s-subnet-cidr" $nodepool.Name) }}"]
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 106: this was failing for me when I tried to test it, I actually had a comment about it, I see you fixed it. :rocket: 
when I run it now it succeeds.

```diff
-address_prefixes     = ["{{index $.Metadata(printf "%s-subnet-cidr" $nodepool.Name) }}"]
+address_prefixes     = ["{{index $.Metadata (printf "%s-subnet-cidr" $nodepool.Name) }}"]
```

services/terraformer/server/clusterBuilder/clusterBuilder_test.go
```diff
@@ -42,3 +43,72 @@ func TestGenerateTf(t *testing.T) {
...
+			position: 2,
+			existing: map[string]struct{}{
+				"10.0.0.0/24": {},
+			},
+			out: "10.0.10.0/24",
+		},
+	}
+	for _, test := range testDataFail {
+		if _, err := getCIDR(test.baseCIDR, test.position, test.existing); err == nil {
+			t.Error(test.desc, fmt.Errorf("test should have failed, but was successful"))
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 109: why are we using `fmt.Errorf` here? wouldn't a plain `string` do?

services/terraformer/server/clusterBuilder/clusterBuilder.go
```diff
@@ -277,3 +323,39 @@ func getTplFile(clusterType pb.ClusterType) string {
...
+			// CIDR does not exist yet, break.
+			return fmt.Sprintf("%s/%d", ip.String(), ones), nil
+		}
+	}
+}
+
+// copyCIDRsToMetadata copies CIDRs for subnets in VPCs for every nodepool.
+func copyCIDRsToMetadata(data *NodepoolsData) {
+	if data.Metadata == nil {
+		data.Metadata = make(map[string]any)
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 356: why are we mapping to `any`? is there any structure that would be better suited to describe the intent of the map?

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf (commented):
overall looks good, had a couple of questions though.

services/terraformer/server/clusterBuilder/clusterBuilder.go
MiroslavRepka: No, as this map holds any data, we might need in the templates. Currently, it holds LB roles for LB templates and CIDRs for VPC subnets.

services/terraformer/server/clusterBuilder/clusterBuilder_test.go
MiroslavRepka: I suppose it can be just string.

proto/config.proto
```diff
@@ -82,6 +82,13 @@ message NodePool {
   repeated Node nodes = 10;
   Provider provider = 11 ;
   bool isControl = 12;
+  map<string, MetaValue> metadata = 13;
+}
+
+message MetaValue {
```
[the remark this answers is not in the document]
MiroslavRepka: You can think about this message as a `union` in C. [Ref](https://protobuf.dev/programming-guides/proto3/#oneof)

services/terraformer/server/clusterBuilder/clusterBuilder.go
```diff
@@ -277,3 +323,39 @@ func getTplFile(clusterType pb.ClusterType) string {
...
+func getCIDR(baseCIDR string, position int, existing map[string]struct{}) (string, error) {
+	_, ipNet, err := net.ParseCIDR(baseCIDR)
+	if err != nil {
+		return "", fmt.Errorf("cannot parse a CIDR with base %s, position %d", baseCIDR, position)
+	}
+	ip := ipNet.IP
+	ones, _ := ipNet.Mask.Size()
+	var i int
+	for {
+		if i > 255 {
```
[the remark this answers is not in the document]
MiroslavRepka: Yeah, I am not 100% sure here. The CIDR which we need is based on the hardcoded value of the VPC CIDR (which is always `10.0.0.0/16`). Previously, we calculated all CIDR for the subnet in a way, that they only had `/24` (i.e. `10.0.1.0/24`,...). Thus, I kept the assumptions and only allowed `/24`.

Bear in mind that this is only used in a cloud provider. The WireGuard VPN respects the CIDR user defines, with netmask.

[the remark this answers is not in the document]
MiroslavRepka: > so do I get it right that with the existing settings, you should theoretically be able to create 256-1 nodepools?

The current constrain is
- 256 nodes per node pool 
- 256 node pools per single region (same provider project/resource group, etc.)

So theoretically, you can have 65536 nodes under the same VPC based on the IP constraints alone.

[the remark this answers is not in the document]
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf: right, thanks.
those are some impressive numbers, worth flaunting in the docs. :))

one thing still not clear to me: how come we get 256 and not 255 nodes per nodepool? I might be missing some basic stuff here so go easy on me please.

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf (approved):
LGTM, thanks @MiroslavRepka.

Issue #641, opened 2023-03-16
Longhorn hardening


cloudziu:
At this point [Longhorn](https://longhorn.io/docs/1.4.0/) driver is backed up by Claudie node OS disks - this creates issues when for example: upgrading OS version can result in data loss. 
The desired state would be to decouple the OS disk and Storage Disk. 
This link can be helpful when designing the change: https://longhorn.io/docs/1.4.0/best-practices/#configuring-default-disks-before-and-after-installation

Tasks:
- [ ] get rid of the field that controls the OS disk size, and use a hardcoded sensible default size
- [ ] replace that field by the "storage-disk" disk size, which is used exclusively for longhorn
- [ ] reconfigure longhorn to always use the "storage disk" only

cloudziu:
I think we should check what are the constraints on the cloud provider's site. For example, Hetzner minimum disk size is 10GB : [ref](https://www.hetzner.com/news/12-18-volumes/)

MarioUhrik:
I would choose to define a global minimum value, and probably specify it to be the minimum of all supported providers.
I think 10GB is OK.
I would go with this as the simplest solution, despite it not necessarily being the best.

PR #698, opened 2023-04-06
Feat/storage disk

MiroslavRepka:
This PR implements #641 by creating a separate disk for Longhorn replication on all compute nodes.

Notable changes:
- `storageDiskSize` field in the input manifest is now optional and is only used for compute node pools. (The default value is 50)
- OS disk for all k8s nodes is now `100GB` by default
- OS disk for all LB nodes is now `30GB` by default


services/terraformer/templates/oci.tpl
```diff
@@ -161,14 +161,21 @@ resource "oci_core_instance" "{{ $node.Name }}" {
         - iptables -Z
         # Make changes persistent
         - netfilter-persistent save
+        {{- if not $nodepool.IsControl }}
+        # Mount volume
+        - mkdir -p /opt/claudie/data
+        - mkfs.xfs /dev/sdb
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 167: can we be sure that the "storage" disk will always be recognised as `sdb`?
would there be an option to get a UUID from GCP or `/dev/disk/by-id` or sth?

services/terraformer/templates/hetzner.tpl
```diff
@@ -85,7 +85,30 @@ resource "hcloud_server" "{{ $node.Name }}" {
     "managed-by"      : "Claudie"
     "claudie-cluster" : "{{ $clusterName }}-{{ $clusterHash }}"
   }
+
+{{- if not $nodepool.IsControl }}
+  user_data = <<-EOF
+#!/bin/bash
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 91: should we add `set -euxo pipefail`?

we don't want the script to end even with errors, do we?
not sure what difference a script failure would make here tho, we'd need to find out how errors in `user_data` are handled by Hetzner here...

but I definitely wouldn't want the script to modify fstab if it could not mount the disk in the previous step, much less format it..that modification would in fact result in at least the machine not being able to boot again (unable to mount disk).

also, shoudln't we run `systemctl daemon-reload` after modifying fstab to let the system know there's a new mount?

services/terraformer/templates/gcp.tpl
```diff
@@ -80,13 +80,48 @@ resource "google_compute_instance" "{{ $node.Name }}" {
...
     ssh-keys = "root:${file("./public.pem")}"
   }
-  metadata_startup_script = "echo 'PermitRootLogin without-password' >> /etc/ssh/sshd_config && echo 'PubkeyAuthentication yes' >> /etc/ssh/sshd_config && service sshd restart"
+  metadata_startup_script = <<EOF
+# Allow ssh as root
+echo 'PermitRootLogin without-password' >> /etc/ssh/sshd_config && echo 'PubkeyAuthentication yes' >> /etc/ssh/sshd_config && service sshd restart
+
+{{- if not $nodepool.IsControl }}
+# Mount managed disk only when not mounted yet
+if ! grep -qs "/dev/sdb" /proc/mounts; then
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 89: same comment as in hetzner, with one difference being *why is there not shebang here?*

services/terraformer/templates/azure.tpl
```diff
@@ -197,10 +199,21 @@ resource "azurerm_virtual_machine_extension" "{{ $node.Name }}_{{ $clusterHash }
...
       "script": "${base64encode(<<EOF
+      # Allow ssh as root
       sudo sed -n 's/^.*ssh-rsa/ssh-rsa/p' /root/.ssh/authorized_keys > /root/.ssh/temp
       sudo cat /root/.ssh/temp > /root/.ssh/authorized_keys
       sudo rm /root/.ssh/temp
       sudo echo 'PermitRootLogin without-password' >> /etc/ssh/sshd_config && echo 'PubkeyAuthentication yes' >> /etc/ssh/sshd_config && echo "PubkeyAcceptedKeyTypes=+ssh-rsa" >> sshd_config && service sshd restart
+      
+      {{- if not $nodepool.IsControl }}
+      # Mount managed disk only when not mounted yet
+      if ! grep -qs "/dev/sdc" /proc/mounts; then
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 210: same comment as in Hetzner tpl..with one difference - how do we know the disk gets recognised as `sdc` every time? this mapping is unreliable and should not be used.

`sdc` might be another drive next reboot and we should be getting it's UUID/PARTID instead...

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf (changes requested):
hey @MiroslavRepka, had a look, left some comments..

docs/input-manifest/example.yaml
```diff
@@ -131,7 +131,7 @@ nodePools:
   #     count:          # Static number of nodes in this nodepool.
   #     serverType:     # Machine type of the nodes in this nodepool.
   #     image:          # OS image of the nodes in the nodepool.
-  #     diskSize:       # Disk size, nodes in the nodepool will have.
+  #     storageDiskSize:       # Disk size, of the storage disk for compute nodepool.
```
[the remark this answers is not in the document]
MarioUhrik: Could we also perhaps clarify that exclusively Longhorn PVs will be stored on there?
If this seems like a good idea to you, at least

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf:
cc @MiroslavRepka 
FYI I had a look into both OCI and GCP and for what it was worth I implemented predictable mounting for GCP.
I haven't actually tested it and apparently the CI didn't run for that commit either, so bear that in mind please. but I'm quite confident it's going to be alright, the docs on it were quite clear :rocket: 

as for OCI, they have their own special thing ("consistent device paths") that allows you to have the attached disk available under a certain path in `/dev/oracleoci/<whatever>` but it's not exactly as simple and configurable as the GCP solution: the `<whatever>` part does not *really* allow arbitrary input, you have to choose from a list of predefined available options, which wouldn't be such a problem, if it were at least available in the TF provider, which, to the best of my knowledge, it is not. I saw it in the GUI though... :see_no_evil: 

so, OCI remains unresolved, I tried to give it a shot but the above mentioned was not the way to go.

refs:
* GCP: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_attached_disk#device_name
* OCI: https://docs.oracle.com/en-us/iaas/Content/Block/References/consistentdevicepaths.htm

MarioUhrik (approved):
Hi Miro, still looks fine to me.

I think @Despire should have a look for a more in-depth review

services/terraformer/templates/oci.tpl
```diff
@@ -163,10 +163,18 @@ resource "oci_core_instance" "{{ $node.Name }}" {
         - netfilter-persistent save
         {{- if not $nodepool.IsControl }}
         # Mount volume
-        - mkdir -p /opt/claudie/data
-        - mkfs.xfs /dev/sdb
-        - mount /dev/sdb /opt/claudie/data
-        - echo "/dev/sdb /opt/claudie/data xfs defaults 0 0" >> /etc/fstab
+        - |
+          sleep 50
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 167: could this be replaced by a `while` loop waiting e.g. `1s` before each check?
also, which is better: to stand up the machine without the requested disk or have it stuck in this stage?

services/terraformer/templates/oci.tpl
```diff
@@ -205,11 +218,12 @@ resource "oci_core_volume" "{{ $node.Name }}_volume" {
 }
 
 resource "oci_core_volume_attachment" "{{ $node.Name }}_volume_att" {
-  provider         = oci.k8s_nodepool_{{ $nodepool.Region }}
+  provider        = oci.k8s_nodepool_{{ $nodepool.Region }}
   attachment_type = "paravirtualized"
   instance_id     = oci_core_instance.{{ $node.Name }}.id
   volume_id       = oci_core_volume.{{ $node.Name }}_volume.id
   display_name    = "{{ $node.Name }}-volume-att"
+  device          = "/dev/oracleoci/${var.storage_disk_name}"
```
wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf, line 226: hmm this is how it's done! good stuff.
it'd be awesome if OCI allowed defining truly custom device names but hey, we can't have it all..

wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf (approved):
left some comments, otherwise LGTM :)

Makefile
```diff
@@ -50,7 +50,7 @@ minio:
 
 # Start DynamoDB backend used for state file locks
 dynamodb:
-	docker run --rm -p 8000:8000 --name dynamodb-local -v ~/dynamodb:/home/dynamodblocal/data amazon/dynamodb-local:latest -jar DynamoDBLocal.jar -sharedDb -dbPath ./data
```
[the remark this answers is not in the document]
MiroslavRepka: Yeah, I had some issues with dynamo and while investigating, I found that we were using the `latest` tag. I pinned it down to the current version, looks like I pushed it with the changes. Can remove them if necessary.

services/terraformer/templates/oci.tpl
MiroslavRepka: Yeah, similar to other providers but just enough different to cause me headache :)

MarioUhrik (approved):
LGTM, but I'm wondering why it was necessary to increase the LB disk size for OCI, and if we shouldn't do that everywhere

MiroslavRepka:
The OCI had some limitations, where the volumes (boot/storage) could not be `30GB` as in other providers (`50GB` min). Should we change the other LB boot disk + change the docs?

MarioUhrik:
> The OCI had some limitations, where the volumes (boot/storage) could not be `30GB` as in other providers (`50GB` min). Should we change the other LB boot disk + change the docs?

Up to you Miro, I think either way is fine (secretive anomaly or changing all the others to match)

PR #701, opened 2023-04-11
Implementing clean architecture for the frontend microservice

Archisman-Mridha:
Implemented clean architecture for the frontend microservice.


services/frontend/adapters/outbound/context_box_connector.adapter.go
```diff
@@ -0,0 +1,101 @@
...
+	connectionUri string
+
+	// grpcConnection is the underlying gRPC connection to context-box microservice.
+	grpcConnection *grpc.ClientConn
+
+	// GrpcClient is a gRPC client connection to context-box microservice.
+	GrpcClient pb.ContextBoxServiceClient
+}
+
+func NewContextBoxConnector(connectionUri string) *ContextBoxConnector {
```
MiroslavRepka, line 28: Ditto, please add some comments above functions

services/frontend/domain/usecases/process_manifest_files.usecase.go
```diff
@@ -0,0 +1,154 @@
...
+			defer func() {
+				manifestProcessingResultsChan <- &ManifestProcessingResult{
+					manifestName:     manifest.Name,
+					rawManifestData:  rawManifestData,
+					manifestFilepath: manifestFilepath,
+					processingError:  processingError,
+				}
+			}()
+
+			if rawManifestData, processingError = os.ReadFile(manifestFilepath); err != nil {
```
MiroslavRepka, line 65: `err != nil` the err variable is not one we are after here. I think you meant `processingError != nil`

services/frontend/domain/usecases/process_manifest_files.usecase.go
```diff
@@ -0,0 +1,154 @@
...
+				return
+			}
+
+			if processingError = yaml.Unmarshal(rawManifestData, &manifest); processingError != nil {
+				return
+			}
+
+			manifest.Validate()
```
MiroslavRepka, line 73: No error check here

services/frontend/domain/usecases/process_manifest_files.usecase.go
```diff
@@ -0,0 +1,154 @@
...
+
+	go func() {
+		waitGroup.Wait()
+
+		close(manifestProcessingResultsChan)
+	}()
+
+	// Collect processing results of manifest files which were processed successfully
+	for manifestProcessingResult := range manifestProcessingResultsChan {
+		// TODO: review
```
MiroslavRepka, line 85: I think you can remove this :)

services/frontend/domain/usecases/process_manifest_files.usecase.go
```diff
@@ -0,0 +1,154 @@
...
+		if manifestProcessingResult.processingError != nil {
+			log.Error().Msgf("Skipping over file %v due to processing error : %v", manifestProcessingResult.manifestFilepath, manifestProcessingResult.processingError)
+			continue
+		}
+
+		configs = removeConfig(configs, manifestProcessingResult.manifestName)
```
MiroslavRepka, line 91: This function has to be called first in the loop, as any configs left over will be deleted.

If we leave this line here, after checking for errors in processing, the configs which had an error will get deleted, which we do not want.

services/frontend/domain/usecases/process_manifest_files.usecase.go
```diff
@@ -0,0 +1,154 @@
...
+
+		err = u.ContextBox.SaveConfig(
+			&pb.Config{
+				Name:     manifestProcessingResult.manifestName,
+				Manifest: string(manifestProcessingResult.rawManifestData),
+			},
+		)
+		if err != nil {
+			log.Error().Msgf("Skipped saving config %v due to error : %v", manifestProcessingResult.manifestName, err)
```
MiroslavRepka, line 100: `"Skipped saving config %v due to error : %v"` -> `"Failed to save config %v due to error`

services/frontend/domain/usecases/process_manifest_files.usecase.go
```diff
@@ -0,0 +1,154 @@
...
+
+		log.Info().Msgf("Details of manifest file %s has been saved to context-box database", manifestProcessingResult.manifestFilepath)
+	}
+
+	// threadSafeMap is a go-routine safe map that stores id of configs that are being currently deleted
+	// to avoid having multiple go-routines deleting the same configs from MongoDB (of contextBox microservice).
+	var threadSafeMap sync.Map
+
+	for _, config := range configs {
+		if _, isConfigBeingDeleted := threadSafeMap.Load(config.Id); isConfigBeingDeleted {
```
MiroslavRepka, line 112: `if _,  ok := ...` is a more go idiomatic way IMO, but I am okay if you leave it like this :).

services/frontend/domain/usecases/process_manifest_files.usecase.go
```diff
@@ -0,0 +1,154 @@
...
+			log.Error().Msgf("Skipped saving config %v due to error : %v", manifestProcessingResult.manifestName, err)
+			continue
+		}
+
+		log.Info().Msgf("Details of manifest file %s has been saved to context-box database", manifestProcessingResult.manifestFilepath)
+	}
+
+	// threadSafeMap is a go-routine safe map that stores id of configs that are being currently deleted
+	// to avoid having multiple go-routines deleting the same configs from MongoDB (of contextBox microservice).
+	var threadSafeMap sync.Map
```
MiroslavRepka, line 109: Do we really need this map? If we iterate over `configs`, how can get a single config get picked up twice?

services/frontend/main.go
```diff
@@ -0,0 +1,113 @@
...
+			err = waitGroupContext.Err()
+
+		case shutdownSignal := <-shutdownSignalChan:
+			log.Info().Msgf("Received program shutdown signal %v", shutdownSignal)
+			err = errors.New("Program interruption signal")
+		}
+
+		// First shutdown the HTTP server to block any incoming connections.
+		// And wait for all the go-routines to finish their work.
+		performGracefulShutdown := func() {
```
MiroslavRepka, line 95: Any reason to save the function into the variable, instead of calling it straight?

MiroslavRepka (changes requested):
Overall, really nice work @Archisman-Mridha! I left some comments, we can discuss them even further if needed. :)

services/frontend/main.go
Archisman-Mridha: Opted to calling the function directly. Added a meaningful comment in the beginning.

services/frontend/domain/usecases/process_manifest_files.usecase.go
Archisman-Mridha: Let it be like this please :)

services/frontend/adapters/inbound/k8s_sidecar_notifications_receiver.adapter.go
```diff
@@ -0,0 +1,106 @@
...
+
+	return k8sSidecarNotificationsReceiver, k8sSidecarNotificationsReceiver.PerformHealthCheck()
+}
+
+func (k *K8sSidecarNotificationsReceiver) registerNotificationHandlers() {
+
+	var router *http.ServeMux = http.NewServeMux()
+
+	// TODO: rename the route "/reload" to "/process-manifest-files"
+	router.HandleFunc("/reload", k.processManifestFilesHandler)
```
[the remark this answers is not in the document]
Archisman-Mridha: No, this PR is regarding the implementation of clean architecture :). So let's stick to that. I wrote that comment so that we can discuss about it (and implement it if we agree) after this PR gets merged.

services/frontend/domain/usecases/process_manifest_files.usecase.go
Archisman-Mridha: In the original codebase, the **server** struct contains a property called **deletingConfigs** which is a `go-routine safe map` -
```go
type server struct {
    ...
    // deletingConfigs is a go-routine safe map that
    // stores configs that are being currently deleted
    // to avoid having multiple go-routines deleting the
    // same config from the database.
    deletingConfigs sync.Map
    ...
}
```

Those manifest files which were deleted from the directory, needs to be deleted from the context-box DB as well. Here is the implementation in  the original codebase -
```go
for _, config := range configs.Configs {
    if _, ok := s.deletingConfigs.Load(config.Id); ok {
        continue
    }
    s.deletingConfigs.Store(config.Id, nil)
    go func(config *pb.Config) {
        ...
	s.deletingConfigs.Delete(config.Id)
    }(config)
}
```

You can see the **go-routine safe map** is used here!

Despire: In the current version there is a `sync.Map` because if you would create 2 or 3 secrets one behind the other multiple go-routines would be spawned that would process the stored manifests.

https://github.com/berops/claudie/blob/master/services/frontend/server.go#L152-L159

The `sync.Map` is just a guard to be sure that the deletion process is started only once and not multiple times.

Whereas in this refactored version https://github.com/berops/claudie/blob/f8a04437c0b15fd582b4aa9d097ee848cfb024bd/services/frontend/domain/usecases/process_manifest_files.usecase.go#L129-L131

There is no guard againts that since the map is defined in the same function where the configs are being processed

services/frontend/adapters/inbound/k8s_sidecar_notifications_receiver.adapter.go
```diff
@@ -0,0 +1,131 @@
...
+	return k8sSidecarNotificationsReceiver, k8sSidecarNotificationsReceiver.PerformHealthCheck()
+}
+
+// registerNotificationHandlers registers a router to the underlying HTTP server.
+// The router contains a route ("/reload") that handles incoming notifications from the K8s-sidecar.
+func (k *K8sSidecarNotificationsReceiver) registerNotificationHandlers() {
+
+	var router *http.ServeMux = http.NewServeMux()
+
+	// TODO: rename the route "/reload" to "/process-manifest-files"
```
Despire, line 66: Is this planned ? if not remove the `TODO`

If yes, rename it in the manifest as well https://github.com/berops/claudie/blob/065c07e714e392be35a7afe9b8887fdc3566b508/manifests/claudie/frontend.yaml#L82

services/frontend/adapters/outbound/context_box_connector.adapter.go
MiroslavRepka: Not a big deal, but comments above functions usually start with the function name, i.e. how you did it in here.

(This applies to more places, not only here)
```go
// PerformHealthCheck checks health of the underlying gRPC connection to context-box microservice
func (c *ContextBoxConnector) PerformHealthCheck() error {
```

MiroslavRepka (commented):
Thanks @Archisman-Mridha for implementing the feedback looks better now :). I left some more comments but nothing serious. Please @Despire also have a look.

services/frontend/adapters/outbound/context_box_connector.adapter.go
```diff
@@ -0,0 +1,109 @@
...
+
+// Communicates with the gRPC server of context-box microservice
+type ContextBoxConnector struct {
+	connectionUri string
+
+	// grpcConnection is the underlying gRPC connection to context-box microservice.
+	grpcConnection *grpc.ClientConn
+
+	// GrpcClient is a gRPC client connection to context-box microservice.
+	GrpcClient pb.ContextBoxServiceClient
```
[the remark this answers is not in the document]
Archisman-Mridha: No. switching to lowercase.

services/frontend/main.go
```diff
@@ -0,0 +1,109 @@
...
+func run() error {
+	contextBoxConnector := outboundAdapters.NewContextBoxConnector(envs.ContextBoxURL)
+	err := contextBoxConnector.Connect()
+	if err != nil {
+		return err
+	}
+
+	usecases := &usecases.Usecases{
+		ContextBox: contextBoxConnector,
+	}
```
Despire, line 49: Missing  initialization of the `done` channel,

services/frontend/adapters/inbound/k8s_sidecar_notifications_receiver.adapter.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// NewK8sSidecarNotificationsReceiver creates an instance of the K8sSidecarNotificationsReceiver struct.
+// It registers the notification handling route to the underlying HTTP server.
+// Then performs an initial healthcheck of K8sSidecarNotificationsReceiver.
+func NewK8sSidecarNotificationsReceiver(usecases *usecases.Usecases) (*K8sSidecarNotificationsReceiver, error) {
+	manifestDir, isEnvFound := os.LookupEnv("MANIFEST_DIR")
+	if !isEnvFound {
+		return nil, fmt.Errorf("env MANIFES_DIR not found")
+	}
```
Despire, line 42: If we're passing the `usecases` here shouldn't we take the `manifest` dir from it instead of calling `os.LookupEnv` ?

move the 

```go
manifestDir, isEnvFound := os.LookupEnv("MANIFEST_DIR")
	if !isEnvFound {
		return nil, fmt.Errorf("env MANIFES_DIR not found")
	}
```

into the `run` function in the `main.go` function

services/frontend/domain/usecases/usecases.go
```diff
@@ -0,0 +1,22 @@
+package usecases
+
+import (
+	"sync"
+
+	"github.com/berops/claudie/services/frontend/domain/ports"
+)
+
+type Usecases struct {
```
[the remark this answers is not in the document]
Archisman-Mridha: Yes these need to be exported.

In `run` function of main.go file -
```go
usecases := &usecases.Usecases{
    ContextBox: contextBoxConnector,
}
```

And in the `Stop` function of k8s_sidecar_notificatins_receiver.go file -
```go
// Stop stops receiving notifications sent by the K8s sidecar.
// The underlying HTTP server is stopped.
func (k *K8sSidecarNotificationsReceiver) Stop() error {
	close(k.usecases.Done)
```

services/frontend/adapters/inbound/k8s_sidecar_notifications_receiver.adapter.go
Archisman-Mridha: **usecases** contains only the business logic. It will be better if we `don't couple the manifestDir with the business logic layer`.

MiroslavRepka (approved):

Issue #606, opened 2023-02-24
Bug: Workflow fails to build


Despire:
EDIT:

This issue is about figuring out which changes should be applied first. Given a change in the desired state where we change the kubernetes version add/remove LBs or nodes or change worker/control planes we should figure out which steps (running kube-eleven, or Terraformer, ansibler etc...) should be executed first.

An example is given below where the current workflow fails

### Current Behaviour

When you change the `control nodes` and attach a `Loadbalancer` in a config, the workflow fails on setting up the loadbalancers when trying to execute the `change api endpoints` ansible playbook as the new control node just completed the terraformer stage so these components are not installed yet (until the KUBE-ELEVEN stage which won't be reached).

### Expected Behaviour

I would expect to skip setting up the loadbalancers in the first stage of the processing and then set them up in the second stage.

### Steps To Reproduce

1. apply a very basic config
```yaml
kubernetes:
  clusters:
    - name: matus-ts3
      version: v1.22.0
      network: 192.168.2.0/24
      pools:
        control:
          - hetzner-control
        compute:
          - hetzner-compute
```

2. change the control node and attach a LB

```yaml
kubernetes:
  clusters:
    - name: test
      version: v1.22.0
      network: 192.168.2.0/24
      pools:
        control:
          - gcp-control
        compute:
          - hetzner-compute
loadBalancers:
  roles:
    - name: apiserver-lb
      protocol: tcp
      port: 6443
      targetPort: 6443
      target: k8sControlPlane
  clusters:
    - name: test-lb
      roles:
        - apiserver-lb
      dns:
        dnsZone: claudie.dev
        provider: cloudflare-1
      targetedK8s: test
      pools:
        - gcp-lb
```

3. the wofklow fails in ANSIBLE stage when setting up the load balancers

cloudziu:
I have encountered kinda similar error. So to reproduce:
1. Create a cluster with a LB Cluster
2. Update manifest with removing the LB at all

I'm getting an ansibler error on step `Changing API Endpoint`
```
2:41PM INF services/ansibler/server/server.go:154 > Loadbalancers for cluster test-cluster project cloudziu were successfully set up module=ansibler
2:44PM INF services/ansibler/server/loadbalancers.go:340 > Changing the API endpoint for the cluster test-cluster module=ansibler
2:44PM INF services/ansibler/server/loadbalancers.go:401 > New endpoint is 5.75.237.5 module=ansibler
EP - test-cluster
[...]

Issue #588, opened 2023-02-17
Bug: Deletion of config fails if builder crashes after deleting nodes.


Despire:
### Current Behaviour

If the builder service crashes after deleting [nodes](https://github.com/berops/claudie/blob/2b1b803f2a5193c9742db7148b54cb9779b87f60/services/builder/configProcessor.go#L102) when It comes back online and starts restarts the workflow for deletion it wail fail.
### Expected Behaviour

The actions should be idempotent, if the nodes were successfully deleted, subsequent deletion shouldn't end with an error.

### Steps To Reproduce

1. Delete a config to trigger the deletion process
2. immediately after deleting the nodes kill the builder process/image
3. Repeat the deletion workflow
4. the builder now fails on deleting the nodes

Despire:
fixed by #728

PR #728, opened 2023-04-27
Fix/deletion

Despire:
Fixes https://github.com/berops/claudie/issues/606

**Why changes in terraformer server ?**

If you would have the following config: 

```yaml
kubernetes:
  clusters:
    - name: matus-t1
      version: v1.24.0
      network: 192.168.2.0/24
      pools:
        control:
          - hetzner-control
        compute:
          - gcp-compute
          - hetzner-compute
```

and then remove the `gcp-compute` compute plane terraformer would fail because when it would enter the building stage it wouldn't find the `.tf` files for the  gcp provider and thus not be able to delete it.

The changes now generate the `.tf` files for the `provider` both from the current and desired state.
Related `.tf` files have been moved into its own directory e.g. (gcp, aws ....)

**Why changes in the kuber?**
The current api endpoint change playbook: https://github.com/berops/claudie/blob/master/services/ansibler/server/ansible-playbooks/apiEndpointChange.yml

Corretly changes the endpoint, however 1 "config" was missed. Kubeadm uses the `kubectl get cm cluster-info -oyaml -n kube-public` and this was left unchanged and as a result wrong endpoints were used for joining new nodes.

The new kuber handler updates the config in the `kube-public` namespace if a change in kubeconfig is detected.
ref: https://kubernetes.io/docs/reference/setup-tools/kubeadm/implementation-details/#shared-token-discovery

**Why new ansibler endpoint?**

When we remove a control plane and add a new one at the same time, if there was an endpoint in the removed control plane we need to move it to the control plane.

**Why changes in the Builder?**

The original `tmpConfig` https://github.com/berops/claudie/blob/7b305222b3ac22480f526a226bcd3fc704ea2d73/services/builder/configProcessor.go#L86-L111

Was adjusted, basically here is the current flow of building a cluster.

1. Find the dirrefence between the current and desired state. If we're `adding and deleting` or `adding and addingLb with API` we then need to first run a tmp config called `IntermediateRepresentation`. This config should be put in the workflow without any *loadbalancers*, this should only prepare the addition of nodes if any.

2. If there was detected a control plane deletion in which and api endpoint was it will be moved to the new one.

3. Now we can delete the nodes missing in the desired state and everything should work

4. Next the workflow for the desired state will run including the Lbs if any.

5. As the last step delete the Lbs or the cluster missing from the desired state.

To test this a new test-set4 was created (locally tested and works without isssues)

**Why the changes in kube-eleven?**

The master taint is deprecated and the `control-plane` taint should be used instead.

ref: https://kubernetes.io/docs/reference/labels-annotations-taints/#node-role-kubernetes-io-master-taint

I've also updated the kubeone version to the latest 1.6.2 I had some troubles with the current version which were resolved in the new one.


services/terraformer/server/loadbalancer/dns.go
```diff
@@ -162,14 +163,17 @@ func (d DNS) generateFiles(dnsID, dnsDir string, dns *pb.DNS, nodeIPs []string)
 		return fmt.Errorf("error creating provider credential key file for provider %s in %s : %w", dns.Provider.SpecName, dnsDir, err)
 	}
 
-	templateLoader := templateUtils.TemplateLoader{Directory: templateUtils.TerraformerTemplates}
-	tpl, err := templateLoader.LoadTemplate(fmt.Sprintf("%s-dns.tpl", dns.Provider.CloudProviderName))
+	sourceDirectory := templateUtils.TemplateLoader{
+		Directory: path.Join(templateUtils.TerraformerTemplates, dns.GetProvider().GetCloudProviderName(), "dns"),
```
MiroslavRepka, line 167: I think this line will break if the provider is `hetznerdns`, as it will form `<templates>/hetznerdns/dns` instead of `<templates>/hetzner/dns`.

services/terraformer/server/clusterBuilder/clusterBuilder.go
```diff
@@ -407,3 +415,58 @@ func copyCIDRsToMetadata(data *NodepoolsData) {
 		data.Metadata[fmt.Sprintf(subnetCidrKeyTemplate, np.Name)] = np.Metadata[subnetCidrKey].GetCidr()
 	}
 }
+
+// generateProviderTemplates generates only the `provider.tpl` templates so terraform can
+// destroy the infra if needed.
+func generateProviderTemplates(current, desired *pb.ClusterInfo, clusterID, directory, suffix string) error {
+	currentNodepools := utils.GroupNodepoolsByProvider(current)
```
MiroslavRepka, line 422: Do we need to check whether `current` or `desired` is not `nil` before calling this function?

services/terraformer/server/clusterBuilder/clusterBuilder.go
```diff
@@ -158,34 +158,35 @@ func (c ClusterBuilder) DestroyNodepools() error {
...
-	template := templateUtils.Templates{Directory: clusterDir}
-	templateLoader := templateUtils.TemplateLoader{Directory: templateUtils.TerraformerTemplates}
+	// generate Providers terraform configuration
+	providers := provider.Provider{
+		ProjectName: c.ProjectName,
+		ClusterName: clusterID,
+		Directory:   clusterDir,
+	}
 
+	if err := providers.CreateProvider(c.CurrentInfo, c.DesiredInfo); err != nil {
```
MiroslavRepka, line 179: Do we need to check whether `CurrentInfo` or `DesiredInfo` is not `nil` before calling this function?

services/builder/configProcessor.go
```diff
@@ -68,66 +49,115 @@ func configProcessor(c pb.ContextBoxServiceClient, wg *sync.WaitGroup) error {
...
 				}
 			}
 
 			message := fmt.Sprintf("Processing cluster %s config %s", clusterName, config.Name)
-			if tmpDesired != nil {
-				clusterView.ClusterWorkflows[clusterName].Description = "Processing stage [2/2]"
-				message = fmt.Sprintf("Processing stage [2/2] for cluster %s config %s", clusterName, config.Name)
+			if diff.Stages() > 0 {
+				currentStage++
+				clusterView.ClusterWorkflows[clusterName].Description = fmt.Sprintf("Processing stage [%d/%d]", currentStage, stages)
```
MiroslavRepka, line 159: Just a clarification, now, there can be up to 3 stages?

1. Adding new nodes (just k8s, LB is ignored)
2. Deleting nodes
3. Changing API EP and adding LBs?

services/builder/configProcessor.go
```diff
@@ -68,66 +49,115 @@ func configProcessor(c pb.ContextBoxServiceClient, wg *sync.WaitGroup) error {
...
-					loadbalancers:        clusterView.Loadbalancers[clusterName],
-					desiredLoadbalancers: clusterView.DesiredLoadbalancers[clusterName],
-					deletedLoadBalancers: clusterView.DeletedLoadbalancers[clusterName],
+					desiredCluster:       clusterView.CurrentClusters[clusterName],
+					desiredLoadbalancers: clusterView.Loadbalancers[clusterName],
 					Workflow:             clusterView.ClusterWorkflows[clusterName],
 				}
 
-				if ctx, err = buildCluster(ctx, c); err != nil {
+				if err := callKubeEleven(ctx, c); err != nil {
```
MiroslavRepka, line 122: Why are we skipping terraformer/ansibler/kuber here?

MiroslavRepka (commented):
LGTM, great work @Despire! 

I left some comments, mostly just a question.

services/terraformer/server/loadbalancer/dns.go
Despire: Yes, good catch!

services/kube-eleven/Dockerfile
```diff
@@ -4,7 +4,7 @@ ARG TARGETARCH
 
 # download and unzip kube-one binary
 RUN apt-get -qq update && apt-get -qq install unzip
-RUN KUBEONE_V=1.5.3 && \
+RUN KUBEONE_V=1.6.2 && \
```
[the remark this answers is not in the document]
Despire: Yes, definitely 👍

services/terraformer/server/clusterBuilder/clusterBuilder.go
Despire: No, it will be checked in the getProvidersUsed func.

https://github.com/berops/claudie/blob/93402dd20ecfec0335716b3bb7e865f8e013a004/services/terraformer/server/provider/provider.go#L62-L65

services/terraformer/server/clusterBuilder/clusterBuilder.go
Despire: No, it will be checked in the `getProvidersUsed` func

https://github.com/berops/claudie/blob/93402dd20ecfec0335716b3bb7e865f8e013a004/services/terraformer/server/provider/provider.go#L62-L65

services/builder/configProcessor.go
Despire: Yes.

1. is the `adding && deleting` stage or now also `adding & adding LB api endpoint` with the tmp config.
2. the api endpoint change if needed
3. the deletion of nodes if needed

then the original workflow.

services/builder/configProcessor.go
Despire: The idea is a follows.

if you have the following config:

```yaml
kubernetes:
  clusters:
    - name: ts4-c-1
      version: v1.24.0
      network: 192.168.2.0/24
      pools:
        control:
          - hetzner-control
        compute:
          - hetzner-compute

```

and then change it to (as an example): 

```yaml
kubernetes:
  clusters:
    - name: ts4-c-1
      version: v1.24.0
      network: 192.168.2.0/24
      pools:
        control:
          - gcp-control
        compute:
          - hetzner-compute
          - gcp-compute
loadBalancers:
  roles:
    - name: apiserver-lb
      protocol: tcp
      port: 6443
      targetPort: 6443
      target: k8sControlPlane
  clusters:
    - name: ts4-lb-1
      roles:
        - apiserver-lb
      dns:
        dnsZone: claudie.dev
        provider: cloudflare-1
      targetedK8s: ts4-c-1
      pools:
        - gcp-lb
```


In the first stage here:
https://github.com/berops/claudie/blob/f9fbe0ed03f641356c01283da3638b8363d79ef2/services/builder/configProcessor.go#L63-L90

It will run the changes but without LBs as there is a clear dependency issue (you can attach a LB and change the api endpoint as kubelet and other binaries aren't even installed until the kube-eleven stage)


After that point you'll have a state that resembles something like this (note no lbs build yet):
```yaml
kubernetes:
  clusters:
    - name: ts4-c-1
      version: v1.24.0
      network: 192.168.2.0/24
      pools:
        control:
          - hetzner-control
          - gcp-control
        compute:
          - hetzner-compute
          - gcp-compute
```

So all the nodes have been build and went through the workflow and when we enter the stage
https://github.com/berops/claudie/blob/f9fbe0ed03f641356c01283da3638b8363d79ef2/services/builder/configProcessor.go#L93

we basically work with the in-between state so we just transfer the api endpoint to the new control plane, run kube-eleven to get the kubeconfig and patch it in kuber. No need to run ansibler terraformer etc... (has been done)

Then the deletion of nodes follows and building the state now with also the Lbs.

services/kuber/server/server.go
```diff
@@ -54,6 +57,68 @@ type server struct {
 	pb.UnimplementedKuberServiceServer
 }
 
+func (s *server) PatchClusterInfoConfigMap(_ context.Context, req *pb.PatchClusterInfoConfigMapRequest) (*pb.PatchClusterInfoConfigMapResponse, error) {
```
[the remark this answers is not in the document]
Despire: hmm... I'm not really familiar with the patch command.

The config map looks like this:
```json
{
    "apiVersion": "v1",
    "data": {
        "jws-kubeconfig-6e9vpx": "..token...",
        "kubeconfig": "apiVersion: v1\nclusters:\n- cluster:\n    certificate-authority-data: ...certifiacte_authority_data...\n    server: https://0.0.0.0:6443\n  name: \"\"\ncontexts: null\ncurrent-context: \"\"\nkind: Config\npreferences: {}\nusers: null\n"
    },
    "kind": "ConfigMap",
    "metadata": {
        "creationTimestamp": "2030-09-27T09:49:01Z",
        "name": "cluster-info",
        "namespace": "kube-public",
        "resourceVersion": "360",
        "uid": "....uid...."
    }
}
```

The kubeconfig is stored in the "data" however its' not a json its yaml so to change the certificate-authority-data and server I'll need to parse it again.

basically working with 2 yamls where one is nested and only parts of the nested one needs to be changed. I tried replacing the whole kubeconfig but that didn't work not sure why.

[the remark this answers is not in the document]
MiroslavRepka: I am not so sure it can patch the nested value. However, I am surprised that replacing the kubeconfig did not work. Is there any other diff between them, other than EP and certs?

Despire:
With https://github.com/berops/claudie/pull/728/commits/22d224263b49b3a67b1e171b4fa691cdb60874c0 the current state and desired state of the manifest is updated even if the workflow fails. This is done as otherwise if it would fail mid processing without saving changes it would fail on deletion as we couldn't generate the providers for the already build infra.

this also closes issue https://github.com/berops/claudie/issues/588

MiroslavRepka (approved):

MiroslavRepka:
>this also closes issue https://github.com/berops/claudie/issues/588

Are we sure, the issue can be closed, i.e. what if the builder deletes 2 out of 3 nodes successfully? Will the subsequent run delete just the third one?

Despire:
>Are we sure, the issue can be closed, i.e. what if the builder deletes 2 out of 3 nodes successfully? 

Yes, you are right, I've cleared the link to that issues

Despire:
https://github.com/berops/claudie/pull/728/commits/fc7e009c527956c4ea3b15ab054f753c060c0dbc fixes https://github.com/berops/claudie/issues/588

Tested with a simple config 
```yaml
kubernetes:
  clusters:
    - name: matus-t1
      version: v1.24.0
      network: 192.168.2.0/24
      pools:
        control:
          - gcp-control
          - hetzner-control
        compute:
          - hetzner-compute
          - gcp-compute
```

Where I would delete control node and worker node, kill Builder service and let the deletion repeat,

MiroslavRepka (approved):

Issue #721, opened 2023-04-26
Bug: Frontend triggers deletion if updated manifest contains error


MiroslavRepka:
### Current Behaviour

After successfully applying the input manifest, once you update it with some error (in my case, it was missing `"`), the frontend triggers the deletion process.

Logs:
```
4:41PM INF workflow finished for cluster workflow-migration-cluster module=frontend
5:18PM ERR Skipping over file /input-manifests/input-manifest.yaml due to error : yaml: line 6: found unexpected end of stream module=frontend
5:18PM INF Deleting config: WorkflowMigration module=frontend
5:18PM INF Config: WorkflowMigration - cluster workflow-migration-cluster currently in stage DESTROY_TERRAFORMER with status IN_PROGRESS destroying infrastructure module=frontend
```

### Expected Behaviour

Frontend should ignore the manifest if it contains an error.

### Steps To Reproduce

1. Run Claudie
2. Apply manifest with the correct configuration
3. Wait until it finishes building
4. Update the current manifest, so it contains an error
5. See how Frontend finds an error and triggers the deletion of a manifest.
### Anything else to note

This bug is present in the release `v0.2.2`

PR #744, opened 2023-05-02
Fix/frontend bug

MiroslavRepka:
Fixes issue described in #721.

The issue was that the field used to filter manifests which need to be deleted was a part of the manifest. However, if the manifest contained an invalid yaml, i.e. unclosed quote, the yaml was not parsed and the field was undefined. 

This has been fixed by replacing that field with the filename of the manifest.


services/frontend/domain/usecases/process_manifest_files.go
```diff
@@ -71,39 +69,38 @@ func (u *Usecases) ProcessManifestFiles(manifestDir string) error {
 			if rawManifestData, processingError = os.ReadFile(manifestFilepath); processingError != nil {
 				return
 			}
-
 			if processingError = yaml.Unmarshal(rawManifestData, &unmarshalledManifest); processingError != nil {
 				return
 			}
```
Despire, line 74: I've noticed that `yaml.Unmarshal(rawManifestData, &unmarshalledManifest)` doesn't quite make sense. Since `unmarshalledManifest` is already a pointer `*manifest.Manifest`, can you change it to just `yaml.Unmarshal(rawManifestData, unmarshalledManifest)`

Despire (approved):
LGTM @MiroslavRepka , Just one thing I noticed from the previous refactor of the frontend service.

Archisman-Mridha (approved):

PR #747, opened 2023-05-02
Feat/add logging docs

### LINKED ISSUE #684: Chore: Add docs about logging in Claudie
### Description

Currently, there is no mention of how logs work in Claudie, in particular
- What is displayed in info logs
- What is displayed in debug logs
- How to enable debug logs

This issue aims to resolve it. Additionally, it can be expanded to include various customisations possible in Claudie.

### Exit criteria

- [ ] Docs about logging are created

MiroslavRepka:
This PR adds docs about customisation in Claudie on the manifest level.

Closes #684


Despire:
Shouldn't we also make the Builder, Scheduler and frontend port customisable ?

MiroslavRepka:
Builder, Scheduler and Frontend are services that act as a client. I am not sure why the port variables were in `.env` but AFAIK the only place where these ports are used is in health check probes and I think its okay if these will not be customisable.

MarioUhrik (approved):

PR #765, opened 2023-05-04
Implementing clean architecture for the scheduler microservice

Archisman-Mridha:


services/scheduler/utils/lb_cluster.go
```diff
@@ -129,3 +143,90 @@ func getMatchingRoles(roles []manifest.Role, roleNames []string) ([]*pb.Role, er
...
+				}
+				continue desired
+			}
+		}
+	}
+}
+
+// createKeys will create a RSA key-pair and save it into the clusterInfo provided
+// return error if key creation fails
+func createKeys(desiredInfo *pb.ClusterInfo) error {
```
MiroslavRepka, line 193: Ditto, not exactly related to LB only.

services/scheduler/domain/usecases/config_processor.go
```diff
@@ -0,0 +1,91 @@
...
+		return fmt.Errorf("error while saving the config: %w", err)
+	}
+
+	return nil
+}
+
+// saveErrorMessageToConfig saves error message to the config
+// Returns error if not successful, nil otherwise
+func (u *Usecases) saveErrorMessageToConfig(config *pb.Config, contextBoxGrpcClient pb.ContextBoxServiceClient, err error) error {
+	// TODO: Investigate this line - @Miro
```
MiroslavRepka, line 77: `@Miro` -> `@MiroslavRepka` :)

services/scheduler/adapters/outbound/context_box_connector.go
```diff
@@ -0,0 +1,46 @@
...
+func (c *ContextBoxConnector) Connect() error {
+	connection, err := utils.GrpcDialWithInsecure("context-box", envs.ContextBoxURL)
+	if err != nil {
+		return err
+	}
+	c.Connection = connection
+
+	log.Info().Msgf("Initiated connection Context-box: %s, waiting for connection to be in state ready", envs.ContextBoxURL)
+
+	return err
```
MiroslavRepka, line 27: `return nil`

services/scheduler/scheduler.go
```diff
@@ -157,43 +118,14 @@ func main() {
...
-		log.Debug().Msgf("All spawned go-routines finished")
-
-		return nil
-	})
+	log.Info().Msgf("Stopping Scheduler: %v", errGroup.Wait())
+}
 
-	log.Info().Msgf("Stopping Scheduler: %v", g.Wait())
+// healthCheck function is used for querying readiness of the pod running this microservice
+func healthCheck(usecases *usecases.Usecases) error {
```
MiroslavRepka, line 125: Would it make sense to move cbox connection check into the health check function @Despire?

Despire: It would makes sense, though there would be a question what to do about the worker. It pulls a config every few seconds and it would wouldn't be able to connect to cbox that it would just flood the logs with these error messages instead of printing cbox not available.

PR #773, opened 2023-05-05
Feat/deletion error

### LINKED ISSUE #712: Bug: Infra not deleted
### Current Behaviour

In a kubernetes cluster If you start a workflow by creating a secret for a config and delete the secret to delete the infra, it at some point while its still being worked on the and worfklow ends with an error the secret was deleted however the infra won't.

### Expected Behaviour

The infra should be deleted.

### Steps To Reproduce

1. Deploy claude to a kubernetes cluster and create a create the secret for the requested config.
2. delete the secret at some point while its still being processed
3. make sure the workflow fails (by deleting the pods etc...)
4. see that the infra wont get deleted

MiroslavRepka:
This PR fixes #712 by ignoring errors from the build stage, if the config should be triggered for deletion.

This PR also contains a few QoL improvements, like adjusting the number of retries to more reasonable values.


services/context-box/server/server.go
```diff
@@ -55,13 +55,25 @@ func (*server) SaveConfigScheduler(ctx context.Context, req *pb.SaveConfigReques
...
+			if err := database.UpdateWorkflowState(config.Name, cluster.ClusterInfo.Name, config.State[cluster.ClusterInfo.Name]); err != nil {
+				return nil, fmt.Errorf("error while updating workflow state for k8s cluster %s in config %s : %w", cluster.ClusterInfo.Name, config.Name, err)
+			}
+		}
+		// Update workflow state for LB clusters.
+		for _, cluster := range config.DesiredState.LoadBalancerClusters {
+			if err := database.UpdateWorkflowState(config.Name, cluster.ClusterInfo.Name, config.State[cluster.ClusterInfo.Name]); err != nil {
+				return nil, fmt.Errorf("error while updating workflow state for LB cluster %s in config %s : %w", cluster.ClusterInfo.Name, config.Name, err)
+			}
+		}
```
Despire, line 74: LB state isn't tracked, as of now, in builder when the config enters the workflow. The state of the cluster represents both the LB's attached to it and the cluster itself, thus this section is redundant.

services/context-box/server/server.go
```diff
@@ -109,26 +121,42 @@ func (*server) SaveConfigBuilder(ctx context.Context, req *pb.SaveConfigRequest)
...
+		if err := database.UpdateWorkflowState(config.Name, cluster.ClusterInfo.Name, config.State[cluster.ClusterInfo.Name]); err != nil {
+			return nil, fmt.Errorf("error while updating workflow state for k8s cluster %s in config %s : %w", cluster.ClusterInfo.Name, config.Name, err)
+		}
+	}
+	// Update workflow state for LB clusters.
+	for _, cluster := range config.CurrentState.LoadBalancerClusters {
+		if err := database.UpdateWorkflowState(config.Name, cluster.ClusterInfo.Name, config.State[cluster.ClusterInfo.Name]); err != nil {
+			return nil, fmt.Errorf("error while updating workflow state for LB cluster %s in config %s : %w", cluster.ClusterInfo.Name, config.Name, err)
+		}
+	}
```
Despire, line 158: ditto

services/context-box/server/cbox.go
```diff
@@ -66,6 +67,36 @@ func (ci *ConfigInfo) HasError() bool {
...
+		// Scheduler queue
+		return true
+	}
+	if ci.MsChecksum == nil && ci.DsChecksum == nil && ci.CsChecksum != nil {
+		// Builder queue
+		return true
+	}
+	// Not scheduled for deletion
+	return false
+}
```
Despire, line 98: Do I understand the deletion correctly? Scheduler will enqueue configs scheduled for deletion and builder will try to execute the destroy of the config 3x before failure.

services/context-box/server/server.go
MiroslavRepka: My bad did not notice that, thanks for pointing this out!

services/context-box/server/cbox.go
MiroslavRepka: Yes, the current workflow is 

secret gets deleted -> Frontend sets msChecksum to nil, -> Context-box pushes config to SchedulerQueue ->  Scheduler sets dsChecksum to nil -> Context-box pushes config to BuilderQueue -> Builder will try to remove it 3 times before aborting with an error if any.

Despire (approved):

Issue #746, opened 2023-05-02
Bug: Unreliable Volume replication 


MiroslavRepka:
### Current Behaviour

While working on #651, I noticed that replication of the volume replicas is somewhat unreliable. It looks like PR #648 I did a few months ago has some shortcomings. This issue aims to investigate why the behaviour is unreliable and hopefully implement a fix.

### Expected Behaviour

The replication of longhorn volume replicas should work without any issues. See #648 and #564.

### Steps To Reproduce

1. Deploy Claudie and build a simple cluster
2. Deploy stateful workload (i.e. etcd)
3. Change all worker nodes 
4. Check the PV status

MarioUhrik:
Hi Miro, for my curiosity, I'm wondering what exact bad-case scenario you've observed

MiroslavRepka:
There were two cases I observed.

1. No replicas were created, even though Kuber thought they were.
2. Too many replicas were present. (Ones from the removed nodes were not cleaned up)

PR #782, opened 2023-05-11
Feat/volume replication

MiroslavRepka:
This PR fixes the issue described in #746 by implementing more checks and cordoning the nodes before Claudie will move Replicas of Volumes. This was tested by migrating the ETDC cluster to multiple new node pools, where each migration resulted in the same state of the database.

There might occasionally be still one issue, which is not severe, as all replicas will migrate with the data correctly. The issue is described as a note in [one comment](https://github.com/berops/claudie/compare/feat/volume-replication?expand=1#diff-83286613a1db2b0823d24bedfdef1758c39d2ac7f08a39a1a8b491fc0930d0bbR111-R116), with the link to the discussion I started with Longhorn maintainers.

### DIFF
--- a/services/kuber/server/nodes/delete.go
+++ b/services/kuber/server/nodes/delete.go
@@ -108,6 +108,12 @@ func (d *Deleter) DeleteNodes() (*pb.K8Scluster, error) {
 		if err := d.deleteNodesByName(kubectl, worker, realNodeNames); err != nil {
 			return nil, fmt.Errorf("error while deleting nodes from worker nodes for %s : %w", d.clusterPrefix, err)
 		}
+		// NOTE: Might need to verify if the volume got detached. The issue is still being discussed in #longhorn channel in CNCF slack workspace.
+		// https://cloud-native.slack.com/archives/CNVPEL9U3/p1683744899327089
+		// Issue seen, when once the node is deleted, some volumes might be still attached to the already deleted node, even though all replicas
+		// have moved to other nodes in cluster. Once manually detaching the volume from already deleted node (via UI), it attached again to a new node
+		// with no issues. Since there is no easy way how to verify and detach volume in some CLI way (due to longhorn policy of  doing everything via UI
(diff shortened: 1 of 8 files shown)


services/kuber/server/nodes/delete.go
```diff
@@ -79,22 +81,42 @@ func (d *Deleter) DeleteNodes() (*pb.K8Scluster, error) {
...
+		// Delete worker nodes
 		if err := d.deleteNodesByName(kubectl, worker, realNodeNames); err != nil {
 			return nil, fmt.Errorf("error while deleting nodes from worker nodes for %s : %w", d.clusterPrefix, err)
 		}
+		// NOTE: Might need to verify if the volume got detached. The issue is still being discussed in #longhorn channel in CNCF slack workspace.
+		// https://cloud-native.slack.com/archives/CNVPEL9U3/p1683744899327089
+		// Issue seen, when once the node is deleted, some volumes might be still attached to the already deleted node, even though all replicas
+		// have moved to other nodes in cluster. Once manually detaching the volume from already deleted node (via UI), it attached again to a new node
+		// with no issues. Since there is no easy way how to verify and detach volume in some CLI way (due to longhorn policy of  doing everything via UI
+		// in ClickOps), leaving this note until discussion with longhorn maintainers concludes.
```
Despire, line 116: Could we open a new Issue for this so we don't forget about this ?

Despire (approved):
@MiroslavRepka LGTM 👍

services/kuber/server/nodes/delete.go
MiroslavRepka: They just asked me to open an issue for it in the Longhorn repository. Should we open an issue on our end as well?

Despire: Yes, and Link it with the Longhorn Issue

MiroslavRepka: https://github.com/berops/claudie/issues/784

PR #792, opened 2023-05-15
Feat/remove sidecar

### LINKED ISSUE #587: Bug: K8s-sidecar sometimes misses deletion of an input manifest secret
### Current Behaviour

Sometimes, the `k8s-sidecar` attached to the `frontend`, misses the deletion of a secret containing an input manifest. This causes a huge issue since if users will delete the secret, and k8s-sidecar misses it, they do not have any other way to interact with the claudie for that input manifest.

### Expected Behaviour

The `k8s-sidecar` should not miss any updates regarding the input manifest secret.

### Steps To Reproduce

1. Deploy a Claudie
2. Apply an input manifest
3. Delete the input manifest

(Some repetition of steps 2. and 3. might be needed)

MiroslavRepka:
This PR fixes #587 by removing the `k8s-sidecar`. Frontend now uses k8s API to watch for any input manifests. This simplifies the information flow in Frontend, as well as makes it simpler to maintain.


services/frontend/domain/usecases/process_manifest_files.go
Archisman-Mridha, line 1: Nice. The business logic for processing manifest files is so simplified now :).

services/frontend/adapters/inbound/secret_watcher.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/rs/zerolog/log"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+// SecretWatcher uses kubernetes API (Watch method) to scan through secrets in the specified namespace.
+// If the secret contains input manifest label, it is send through one of the channels in usecases for processing.
```
Archisman-Mridha, line 19: Just a small typo
send -> sent

services/frontend/adapters/inbound/secret_watcher.go
```diff
@@ -0,0 +1,170 @@
...
+				// Due to golangci-lint, this case has to be included.
+				log.Debug().Msgf("Got event Bookmark or Error; not supported")
+			}
+		}
+	}
+	return nil
+}
+
+// getManifest returns usecases.RawManifest from the given file.
+func (sw *SecretWatcher) getManifest(file []byte, secretName, fileName string) (*usecases.RawManifest, error) {
```
Archisman-Mridha, line 145: I think it makes more sense if this function is renamed to `getRawManifest`.

services/frontend/main.go
```diff
@@ -44,41 +40,48 @@ func run() error {
...
 	errGroup.Go(func() error {
-		log.Info().Msgf("Listening for notifications from K8s-sidecar at port: %v", k8sSidecarNotificationsReceiverPort)
-		return k8sSidecarNotificationsReceiver.Start("0.0.0.0", k8sSidecarNotificationsReceiverPort)
+		log.Info().Msgf("Frontend is ready to process input manifests")
+		go usecases.ProcessManifestFiles()
+
+		log.Info().Msgf("Frontend is ready to watch input manifest statuses")
+		go usecases.WatchConfigs()
+
+		log.Info().Msgf("Frontend is watching for any new input manifest")
```
Archisman-Mridha, line 74: I think this comment - **Frontend is watching for any Kubernetes secrets with label "claudie.io/input-manifest" in namespace "claudie"** will make more sense.

Archisman-Mridha (approved):
Good work @MiroslavRepka :+1:  . The code in frontend microservice got simpler than before.

services/frontend/main.go
MiroslavRepka: The `main.go` doesn't really have access to the variables like `label` or `namespace`. We can export it in some way, however, I don't see much value in that.

Archisman-Mridha (changes requested):
Everything else looks good to me. :+1: for the code simplification in the **frontend/main.go** file.

services/context-box/server/cbox.go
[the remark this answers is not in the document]
MiroslavRepka: This error would pop up in the linter, I advise you to pull the latest changes and check again.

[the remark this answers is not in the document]
Archisman-Mridha: Weird - I reloaded VSCode and the error went away.

Archisman-Mridha (approved):

services/frontend/adapters/inbound/secret_watcher.go
```diff
@@ -0,0 +1,216 @@
...
+// decodeContent tries to decode base64 files and returns decoded version. If decoding fails,
+// it might assume the content is not base64 based and returns original content.
+func (sw *SecretWatcher) decodeContent(content []byte) ([]byte, error) {
+	decoded := make([]byte, len(content)*(4/3))
+	if _, err := base64.StdEncoding.Decode(decoded, content); err != nil {
+		// Cant use errors.Is() as base64 package builds error dynamically in base64.CorruptInputError()
+		if strings.Contains(err.Error(), "illegal base64 data") {
+			log.Debug().Msgf("File not base64 compatible, assuming it is string data")
+			return content, nil
+		}
```
Despire, line 195: Why can't we use `errors.Is(err, base64.CorruptInputError)` ? 

The errors return from the function in the base64 package seem to always return `CorruptInputError`

services/frontend/adapters/inbound/secret_watcher.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+}
+
+// monitor continuously watches for secrets in current namespace with specified label, until usecases context will be cancelled or watch result channel gets closed.
+func (sw *SecretWatcher) monitor(w watch.Interface) string {
+	rsv := ""
+	for event := range w.ResultChan() {
+		if secret, ok := event.Object.(*v1.Secret); ok {
+			// Save resource version to continue from there once channel gets closed
```
Despire, line 100: The shortest path should be handled first, and having it like this adds unnecessary indentation

can we rewrite this into ?
```go
secret, ok := event.Object.(*v1.Secret)
if !ok {
   continue
}

.....
```

services/frontend/domain/usecases/process_manifest_files.go
```diff
@@ -14,152 +10,64 @@ import (
...
-		if _, isConfigBeingDeleted := u.configsBeingDeleted.Load(config.Id); isConfigBeingDeleted {
-			continue
-		}
-		u.configsBeingDeleted.Store(config.Id, nil)
-
-		for _, k8sCluster := range config.GetCurrentState().GetClusters() {
-			if _, ok := u.inProgress.Load(k8sCluster.ClusterInfo.Name); !ok {
-				u.inProgress.Store(k8sCluster.ClusterInfo.Name, config)
-			}
-		}
```
Despire, line 136: Could we also add this to the `deleteConfig` func ?

This code was there so that the frontend starts tracking the deletion when it got picked up.
This is so if by any chance the deletion fails immediately after the frontend service picks it up (for example builder is in a crashloop) it will be then printed in the logs that the deletion failed

Since we work with the manifest this would look like.
```go
                for _, k8sCluster := range unmarshalledManifest.Kubernetes.Clusters {
			if _, ok := u.inProgress.Load(k8sCluster.ClusterInfo.Name); !ok {
				u.inProgress.Store(k8sCluster.ClusterInfo.Name, config)
			}
		}
```

services/frontend/domain/usecases/process_manifest_files.go
```diff
@@ -14,152 +10,64 @@ import (
...
-
-		log.Info().Msgf("Details of the manifest file %s has been saved to context-box database", manifestProcessingResult.manifestFilepath)
-
-		// if the config is not in the context-box DB we start to track it.
-		if !isConfigRemoved {
-			for _, k8sCluster := range manifestProcessingResult.unmarshalledManifest.Kubernetes.Clusters {
-				if _, ok := u.inProgress.Load(k8sCluster.Name); !ok {
-					u.inProgress.Store(k8sCluster.Name, config)
-				}
-			}
```
Despire, line 120: Same as with `deleteConfig` func, can we add this code also to the `createConfig` func ? so that the frontend starts tracking the state of the manifest from the moment it was picked up

Despire (changes requested):

services/frontend/adapters/inbound/secret_watcher.go
MiroslavRepka: Oh, my bad, I have looked at `base64.CorruptInputError` and it's defined as `int64`. But yes, it implements `error` interface which I missed, that's why I used this.

Despire (approved):

PR #797, opened 2023-05-18
Remove unnecessary docs

MiroslavRepka:
This PR removes useless docs as per the [conversation in Slack](https://berops.slack.com/archives/C04JSCX9YCB/p1683628191676239?thread_ts=1683620594.330659&cid=C04JSCX9YCB). All of the information has been copied over to the relevant file. 

Changes in `.pb.go` files are due to the addition of comments from `config.proto`


MarioUhrik (approved):

proto/pb/contextBox.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.28.1
-// 	protoc        v3.21.8
+// 	protoc        v3.21.5
```
Despire, line 4: We should probably use the same versions when generating the proto files as we keep overwritting these

Despire (approved):
LGTM 👍

proto/pb/contextBox.pb.go
MiroslavRepka: Good shout, let me have a look ho we can enforce it.

MiroslavRepka: Do you think this could help us enforce it?

```sh
# Enforce same version of protoc 
PROTOC_VERSION = "3.21.8"
CURRENT_VERSION = $$(protoc --version | awk '{print $$2}')
# Generate all .proto files 
proto:
	@if [ "$(CURRENT_VERSION)" = "$(PROTOC_VERSION)" ]; then \
		protoc --go-grpc_out=. --go_out=. proto/*.proto ;\
	else \
		echo "Please update your protoc version. Current $(CURRENT_VERSION) | Required $(PROTOC_VERSION)"; \
	fi
```

Despire: should work 👍

PR #803, opened 2023-05-19
fix mike cmd not found by installing dependencies

JKBGIT1:
Install dependencies, before deploying the new version of the docs.


MiroslavRepka (approved):
LGTM, thanks for the fix @JKBGIT1!

PR #812, opened 2023-05-22
Fix metadata bug

### LINKED ISSUE #790: Bug: Subnet CIDR is not carried over from `tempState` in Builder
### Current Behaviour

Recently, we implemented (PR #661) a fix for the issue described in #647. The issue is about dynamic CIDR which if changed, forced the replacement of VMs. The PR fixed it for most use cases, however, I encountered this same issue under specific conditions.

In Builder, we use `tempState` as an intermediate state between desired and current state, which always adds nodes/node pools to the cluster, before other nodes/node pools will be deleted. However, an issue arises when `tempState` adds a node pool, which has to use a new CIDR, under the same VPC, as the node pool that will be later deleted. The newly defined CIDR is used to create the node pool, however, it is not carried over to the desired state. This causes CIDR to be defined again, however this time it is different as the node pool was deleted under the same VPC. This then again causes issue described in #647.  

### Expected Behaviour

The CIDR calculated in `tempState` should be carried over to desired state, so the nodes are not replaced due to CIDR change.

### Steps To Reproduce

1. Run Claudie
2. Apply this config
```yaml
nodePools:
  dynamic:
    - name: hetzner-control
      providerSpec:
        name: hetzner-1
        region: hel1
        zone: hel1-dc2
      count: 1
      serverType: cpx11
      image: ubuntu-22.04

    - name: gcp-compute
      providerSpec:
        name: gcp-1
        region: europe-west1
        zone: europe-west1-b
      count: 3
      serverType: e2-small
      image: ubuntu-os-cloud/ubuntu-2204-jammy-v20221206
      storageDiskSize: 50

    - name: gcp-compute-new
      providerSpec:
        name: gcp-1
        region: europe-west1
        zone: europe-west1-b
      count: 3
      serverType: e2-small
      image: ubuntu-os-cloud/ubuntu-2204-jammy-v20221206
      storageDiskSize: 50

kubernetes:
  clusters:
    - name: cidr-test
      version: v1.24.0
      network: 192.168.2.0/24
      pools:
        control:
          - hetzner-control
        compute:
          - gcp-compute
```
3. After cluster is built, apply these changes
```yaml
kubernetes:
  clusters:
    - name: cidr-test
      version: v1.24.0
      network: 192.168.2.0/24
      pools:
        control:
          - hetzner-control
        compute:
          - gcp-compute-new
```
4. Watch terraform logs, as it will force replacement of VMs, due to changed subnet CIDR

### Anything else to note

This bug should be fixed by carrying over the calculated CIDR for node pools, created in `tempState`.

MiroslavRepka:
Fixes #790 

Tested with the workflow described in the issue description.

### DIFF
--- a/manifests/claudie/kustomization.yaml
+++ b/manifests/claudie/kustomization.yaml
@@ -23,7 +23,7 @@ images:
 - name: ghcr.io/berops/claudie/ansibler
   newTag: fb1725e-1864
 - name: ghcr.io/berops/claudie/builder
-  newTag: fb1725e-1864
+  newTag: 8d3609d-1878
 - name: ghcr.io/berops/claudie/context-box
   newTag: fb1725e-1864
 - name: ghcr.io/berops/claudie/frontend
--- a/services/builder/configProcessor.go
+++ b/services/builder/configProcessor.go
@@ -130,6 +130,8 @@ func configProcessor(c pb.ContextBoxServiceClient, wg *sync.WaitGroup) error {
 
 				// make the desired state of the temporary cluster the new current state.
 				clusterView.CurrentClusters[clusterName] = ctx.desiredCluster
+				// Update nodepool metadata, as they are not carried over
+				updateNodepoolMetadata(ctx.desiredCluster.ClusterInfo.NodePools, clusterView.DesiredClusters[clusterName].ClusterInfo.NodePools)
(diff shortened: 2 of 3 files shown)


cloudziu (approved):
LGTM 👍

PR #818, opened 2023-05-24
Relocate e2e tests

katapultcloud:


MarioUhrik:
LGTM on the pipeline.

Why the `backoffLimit: 8` on the jobs?

katapultcloud:
> LGTM on the pipeline.
> 
> Why the `backoffLimit: 8` on the jobs?

@MarioUhrik We had what I can explain as an "edge case" where the cluster had 1 worker node which wasn't enough for e2e claudie deployment, so autoscaler kicks in and scales another one. The problem was that create-table job was already scheduled and dynamo was pending for new node, job kept erroring out until reaching backoff. Once the new node was up, dynamo was scheduled but the job was dead, not creating the table. So this is just to prevent this and give job more tries. But maybe there's a better fix for it?

MarioUhrik:
sounds good, thanks

LGTM for me

MiroslavRepka (approved):

PR #819, opened 2023-05-24
Feat/go embed

### LINKED ISSUE #778: Feature: Use go embed package to embed template files
### Motivation

Claudie uses a lot of templates, either for Terraform, Kubeone, Ansible and so on. Each template is its own separate file, which is then dynamically loaded multiple times at runtime. We can optimise this by using the [`embed` package](https://pkg.go.dev/embed) from Go. This package will allow us to embed those files into binary at compile time, lowering the overhead of reading the files + simplifying the code around templates in general. (Instead of using a relative path to find the template, load it and parse it, we could just parse it). 

This can be also extended to other files if beneficial (playbooks, manifests, etc.)

### Description

Claudie should start using go `embed` to embed the template files/possibly other files into its binary, to simplify the code and slightly optimise it.

### Exit criteria

- [ ] Code has been refactored to use go `embed` for templates
- [ ] Other files Claudie uses were reviewed and embedded as necessary

MiroslavRepka:
Closes #778 as it implements `go:embed` package for templates in Claudie.


cloudziu (approved):
LGTM ❤️

PR #822, opened 2023-05-29
remove deployment from mkdocs, add manual deploy guide

JKBGIT1:
When the release-docs.yml pipeline failed, Miro had to deploy the new version of the docs manually.

While he was working with the documentation he found out that the steps for dependencies installation in the [`deployment-workflow.md`](https://github.com/berops/claudie/blob/master/docs/docs-guides/deployment-workflow.md) are missing. I have added this steps and also wrote a few steps for manual deployment of the new docs version, in case the release-docs pipeline fails.

We also had a brief discussion about the the [Deployment workflow](https://docs.claudie.io/v0.3.1/docs-guides/deployment-workflow/) section. We think, that this section probably isn't suitable to be in our deployed docs, because we see this version of the documentation more of a user guide.


mkdocs.yml
```diff
@@ -73,7 +73,6 @@ nav:
     - Claudie v0.3: CHANGELOG/changelog-0.3.x.md
   - How to work with docs:
```
MiroslavRepka, line 74: I think this whole section is kind of irrelevant for Claudie users. I would expect only Claudie devs to work with docs, not Claudie users. 

IMO, the docs.claudie.io is a user guide, not developer guide.

docs/docs-guides/deployment-workflow.md
```diff
@@ -22,8 +22,21 @@ mike set-default <version>
...
\ No newline at end of file
+In case the [release-docs.yml](https://github.com/berops/claudie/blob/master/.github/workflows/release-docs.yml) fails, you can deploy the new version manually by following this steps:
+
+* `git checkout -b <branch>`
+* `python3 -m venv ./venv`
+* `source ./venv/bin/activate`
+* `pip install -r requirements.txt`
+* `mike deploy <version> latest --push`
+
+!!! Warning "Don't forget to use the `latest` tag in the last command, because otherwise the new version will not be loaded as default one, when visiting [docs.claudie.io](docs.claudie.io)"
```
MiroslavRepka, line 40: As this is not going to be rendered in GH, we can change it to `:warning:`

MiroslavRepka (approved):

MarioUhrik (approved):

PR #856, opened 2023-06-02
Add Plausible tracking to docs.claudie.io

bernardhalas:
I do not know how to test this, so any advice is much appreciated. Traffic shall be afterwards observed at https://plausible.io/claudie.io


MiroslavRepka:
I think @JKBGIT1 will now the best how to test changes in `docs.claudie.io` without making it live.

JKBGIT1:
@bernardhalas you can test the `docs.claudie.io` following [this](https://github.com/berops/claudie/blob/master/docs/docs-guides/development.md).

bernardhalas:
For some reason the data is still not flowing in. I'm using https://material-plausible-plugin.ale.sh/
@JKBGIT1 , could you have a look and see if I'm doing something wrong here?

Perhaps local testing won't work here, due to a possible domain verification?

PR #863, opened 2023-06-09
change release-docs pipeline trigger

JKBGIT1:
Change the trigger of the release-docs pipeline to the same as the release pipeline has, because the release tag in Set release tag step wasn't successfully created.


MiroslavRepka (approved):
LGTM, thanks @JKBGIT1

PR #868, opened 2023-06-13
Rework error saving in Scheduler

MiroslavRepka:
While discussing the state map with @cloudziu, we encountered a strange case where the scheduler error was saved under the `"scheduler"` key, not the cluster name key. This could be sufficient however, this PR tries to clarify it and keep it under the cluster names if defined.


services/scheduler/domain/usecases/config_processor.go
```diff
@@ -79,13 +79,20 @@ func (u *Usecases) saveErrorMessageToConfig(config *pb.Config, contextBoxGrpcCli
...
+			Status:      pb.Workflow_ERROR,
+			Description: err.Error(),
+		}
+	} else {
+		// As scheduler does not process config on per cluster basis, the error will be saved under every cluster.
+		for _, state := range config.State {
+			state.Stage = pb.Workflow_SCHEDULER
+			state.Status = pb.Workflow_ERROR
+			state.Description = fmt.Sprintf("Error encountered in Scheduler, halting processing of all clusters : %v", err)
+		}
```
Despire, line 95: I like this 👍 , originally I used `config.State["scheduler"] = &pb.Workflow{ ...` because as you said "scheduler does not process config on per cluster basis"

Despire (approved):

cloudziu (approved):

Issue #861, opened 2023-06-07
Bug: an ERROR state cluster block DONE state clusters


cloudziu:
Consider an input manifest

```yaml
  providers:
    - name: hetzner-secret
      providerType: hetzner
      secretRef:
        name: hetzner
        namespace: claudie    
  nodePools:
    dynamic:
      - name: control-hetzner-1
        providerSpec:
          name: hetzner-secret
          region: nbg1
          zone: nbg1-dc3
        count: 1
        serverType: cpx11
        image: ubuntu-22.04

      - name: compute-hetzner-1
        providerSpec:
          name: hetzner-secret
          region: nbg1
          zone: nbg1-dc3
        count: 1
        serverType: cpx11
        image: ubuntu-22.04
        storageDiskSize: 50    

      - name: control-hetzner-2
        providerSpec:
          name: hetzner-secret
          region: nbg1
          zone: nbg1-dc3
        count: 1
        serverType: cpx11
        image: ubuntu-22.04

      - name: compute-hetzner-2
        providerSpec:
          name: hetzner-secret
          region: nbg1
          zone: nbg1-dc3
        count: 1
        serverType: cpx11
        image: ubuntu-22.04
        storageDiskSize: 50          
  kubernetes:
    clusters:
      - name: first-cluster
        version: v1.24.0
        network: 192.168.2.0/24
        pools:
          control:
            - control-hetzner-1
          compute:
            - compute-hetzner-1

      - name: second-cluster
        version: v1.24.0
        network: 192.168.2.0/24
        pools:
          control:
            - control-hetzner-2
          compute:
            - compute-hetzner-2
```

and the cluster Statuses are:
```json
  "clusters": {
    "first-cluster": {
      "message": "error in Kube-eleven for cluster hetzner-test-cluster project default/hetzner-test-manifest : rpc error: c
ode = Unknown desc = error while building cluster hetzner-test-cluster for project default/hetzner-test-manifest : error whi
le running \"kubeone apply\" in services/kube-eleven/server/clusters/hetzner-test-cluster-n2fmpui : exit status 15",
      "phase": "KUBE_ELEVEN",
      "state": "ERROR"
    },
    "second-cluster": {
      "phase": "NONE",
      "state": "DONE"
    }
  },
```
Because of the ERROR in the `first-cluster`, the user won't be able to operate on the `second-cluster`. The only action he can take is to remove the input-manifest.

MiroslavRepka:
[...]

PR #875, opened 2023-06-15
Fix/partial failure

### LINKED ISSUE #795: Feature: Error handling with gRPC
### Motivation

gRPC has very basic error handling by [default](https://grpc.io/docs/guides/error/#standard-error-model) which is limiting as to what can be send as a response and what can't

### Description

Consider the following scenario, you create a secret to build the following cluster:

```yaml
kubernetes:
  clusters:
    - name: t1
      version: v1.24.0
      network: 192.168.2.0/24
      pools:
        control:
          - gcp-control
        compute:
          - gcp-compute
loadBalancers:
  roles:
    - name: apiserver-lb
      protocol: tcp
      port: 6443
      targetPort: 6443
      target: k8sControlPlane
  clusters:
    - name: t1-lb
      roles:
        - apiserver-lb
      dns:
        dnsZone: claudie.dev
        provider: cloudflare-1
      targetedK8s: t1
      pools:
        - gcp-lb
```

The k8s cluster is build successfully in the `terraformer` service but the LB cluster failed, terraformer returns an error and the context of what infra was build is lost, on the calling side in the `builder` service the workflow ends with an error which is stored in the DB alongside with the current state (which won't reflect the actual current state) and the desired state.

The issue is, now we want to delete the secret and thus delete the infra, which will fail. Why ? because we've lost information about what was build and what was not, and when we enter the deletion stage of the config we work with the current state which doesn't reflect the actual current stage.

the generated `.pb.go` files ignore any response if an error is returned by the RPC.

https://github.com/berops/claudie/blob/555143786993ca227771dad0247e7439909b1ae4/proto/pb/terraformer_grpc.pb.go#L39-L46

**If an error occurs while a service is processing the changes and some get applied and some won't, if we send a response along with the error, so that the calling side would have the changes in the current state reflecting the infra, it gets ignored on the calling side.**

The aim is to resolve this issue coming up with an solution, such that we have both the changes made and the error on the calling side of the gRPC call.

Possible ideas:
 - Ignore `error` in the gRPC calls (i.e. always return a `nil` error), instead include it in the response and the calling side would have to check the error in the response body.
- gRPC [rich error model](https://grpc.io/docs/guides/error/#richer-error-model)

### Exit criteria

- [ ]  The calling side should receive both the error and the current state of the cluster (including any partial changes made until the point of failure where the error occurred)

Despire:
Fixes #795 

In BuildInfrastructure two types of responses are returned, if no errors it works same as current,
If any error occurs during provisioning of the infrastructure a response is returned where the current state reflects the actual current state so that when builder saves the error it has the correct. current state and the infra is possible to be deleted


MiroslavRepka (approved):
LGTM, thanks @Despire

services/terraformer/server/domain/usecases/build_infrastructure.go
```diff
@@ -14,66 +11,108 @@ import (
...
 		logger.Info().Msgf("Infrastructure successfully created for cluster")
 		return nil
 	})
+
 	if err != nil {
-		log.Err(err).Str("project", request.ProjectName).Msgf("Error encountered while building cluster")
-		return nil, fmt.Errorf("error while building cluster %s for project %s : %w", request.Desired.ClusterInfo.Name, request.ProjectName, err)
-	}
+		// Given a failure in the building process construct the response such that those that successfully
+		// build will have their current state = desired state.
```
cloudziu, line 65: Will this also resolve #861?

Despire: No this just handles the case where the deletion of the cluster would fail if only part of the infra was build.

Makefile
```diff
@@ -95,9 +95,9 @@ TARGETARCH = $$(go env GOHOSTARCH)
 REV = $$(git rev-parse --short HEAD)
 SERVICES = $$(command ls services/)
 containerimgs:
-	sed -i "s/image: ghcr.io\/berops\/claudie\/autoscaler-adapter/&:$(REV)/" services/kuber/templates/cluster-autoscaler.goyaml
+	sed -i "" "s/image: ghcr.io\/berops\/claudie\/autoscaler-adapter/&:$(REV)/" services/kuber/templates/cluster-autoscaler.goyaml
```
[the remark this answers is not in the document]
Despire: locally on my machines sed creates a backup file and this just makes it so that it doesn't use a backup file. Unsure if this also is the case for linux versions.

PR #876, opened 2023-06-16
Feat/labels taints

### LINKED ISSUE #829: Feature: support adding taints to node pools
### Motivation

Taints and labels should be able to be added to node pools to prevent workloads scheduled onto nodes where they don't belong.

### Description

Taint and label sections in input-manifest to define user defined taints and labels:

example
```
nodePools:
  dynamic:
    - name: hetzner-control
      providerSpec:
        name: hetzner-1
        region: nbg1
        zone: nbg1-dc3
      count: 1
      serverType: cpx11
      image: ubuntu-22.04
      storageDiskSize: 50
      taints:
        key1=value1:NoSchedule
        key2=value2:NoExecute
      labels:
        key1=value1
        key2=value2
```

### Exit criteria

- [ ] node pools can be created with custom taints
- [ ] node pools can be created with custom labels

MiroslavRepka:
Closes #829 

This PR adds support for user-defined labels and taints for the nodes in the node pool. Both, labels and taints are added via `patch`, as it greatly simplifies logic when labels/taints are added/changed/removed.

It's important to note, that if components like kubelet will add any new labels, Claudie will have to update its logic, so it will not overwrite it.


MarioUhrik:
Looks nice, Miro

I think there should be an example for this in the main huge example manifest

internal/manifest/manifest.go
```diff
@@ -86,13 +84,15 @@ type Kubernetes struct {
...
-	StorageDiskSize  int64            `validate:"omitempty,gte=50" yaml:"storageDiskSize"`
-	AutoscalerConfig AutoscalerConfig `validate:"required_without=Count,excluded_with=Count" yaml:"autoscaler"`
+	Name             string            `validate:"required" yaml:"name"`
+	ProviderSpec     ProviderSpec      `validate:"required" yaml:"providerSpec"`
+	Count            int32             `validate:"required_without=AutoscalerConfig,excluded_with=AutoscalerConfig" yaml:"count"`
+	ServerType       string            `validate:"required" yaml:"serverType"`
+	Image            string            `validate:"required" yaml:"image"`
+	StorageDiskSize  int64             `validate:"omitempty,gte=50" yaml:"storageDiskSize"`
+	AutoscalerConfig AutoscalerConfig  `validate:"required_without=Count,excluded_with=Count" yaml:"autoscaler"`
+	Labels           map[string]string `validate:"omitempty" yaml:"labels"`
```
cloudziu, line 94: Validate labels with DNS compatible [naming convention?](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)

cloudziu (approved):
Great job, left a single comment 👍

docs/input-manifest/api-reference.md
```diff
@@ -248,6 +248,30 @@ Dynamic nodepools are defined for cloud provider machines that Claudie is expect
   
   Autoscaler configuration for this nodepool. Mutually exclusive with `count`.
 
+- `labels`
+
+  Map of user defined labels, which will be applied on every node in the node pool. This field is optional.
+
+  By default, Claudie applies following labels on every node in the cluster.
```
Despire, line 255: should mention that the user defined labels are applied in addition to the default labels ? I can imagine that it might be confusing

Same with the defaults `taints`

internal/nodes/metadata.go
```diff
@@ -0,0 +1,119 @@
...
+	m[string(KubernetesRegion)] = sanitiseString(np.GetDynamicNodePool().Region)
+	// Other labels.
+	m[string(KubernetesOs)] = "linux"   // Only Linux is supported.
+	m[string(KubernetesArch)] = "amd64" // TODO add arch https://github.com/berops/claudie/issues/665
+	m[string(KubeoneOs)] = "ubuntu"     // Only supported Os
+
+	// Add custom user defined labels.
+	for k, v := range np.Labels {
+		m[k] = v
+	}
```
Despire, line 48: What if someones overwrites some of our default labels?

Despire (approved):
LGTM 👍

internal/nodes/metadata.go
MiroslavRepka: Interesting, what would you suggest we do in this case?

cloudziu: First build custom labels, and later overwrite the map with "well-know" labels, to ensure that they won't get changed. WDYT?

MiroslavRepka: I would actually go the other way around. The claudie default labels are quite specific, so if the customer wishes to overwrite them, I think they will have a reason for that. Can't imagine a case where they would mistakenly overwrite them.

MiroslavRepka: In the internal communication it was agreed we will keep default labels even if the user will try to overwrite them.

internal/manifest/validate_node_pool.go
```diff
@@ -2,8 +2,14 @@ package manifest
...
 import (
 	"fmt"
+	"regexp"
 
 	"github.com/go-playground/validator/v10"
+	k8sV1 "k8s.io/api/core/v1"
+)
+
+const (
+	labelRegex = "^(?:[a-z0-9A-Z](?:[-_.]?[a-z0-9A-Z]){0,61})?(?:\\.[a-z0-9A-Z](?:[-_.]?[a-z0-9A-Z]){0,61})*(?:\\/[a-z0-9A-Z](?:[-_.]?[a-z0-9A-Z]){0,61})?$"
```
cloudziu, line 12: This is the regexp to validate is `label` proper? If so, why arent we using `"github.com/go-playground/validator/v10"` for that? https://github.com/go-playground/validator#baked-in-validations

services/kuber/server/nodes/patch.go
```diff
@@ -9,41 +10,95 @@ import (
...
 	if log.Logger.GetLevel() == zerolog.DebugLevel {
 		kc.Stdout = comm.GetStdOut(clusterID)
 		kc.Stderr = comm.GetStdErr(clusterID)
 	}
-	return &Patcher{kc: kc, nodepools: cluster.ClusterInfo.NodePools, clusterID: clusterID}
+
+	return &Patcher{kc: kc, desiredNodepools: cluster.ClusterInfo.NodePools, clusterID: clusterID}
 }
 
 func (p *Patcher) PatchProviderID(logger zerolog.Logger) error {
```
cloudziu, line 46: Won't it better to add the logger to the `*Pather` struct?

services/kuber/server/server.go
```diff
@@ -301,7 +301,7 @@ func (s *server) StoreKubeconfig(ctx context.Context, req *pb.StoreKubeconfigReq
 	logger := utils.CreateLoggerWithClusterName(clusterID)
 	if namespace := envs.Namespace; namespace == "" {
 		//NOTE: DEBUG print
-		// logger.Info().Msgf("The kubeconfig for %s\n%s:", clusterID, cluster.Kubeconfig)
+		logger.Info().Msgf("The kubeconfig for %s\n%s:", clusterID, cluster.Kubeconfig)
```
cloudziu, line 304: To debug, or not to debug, that is the question. :D

cloudziu (approved):
LGTM 👍 left some minor questions

internal/manifest/validate_node_pool.go
MiroslavRepka: I believe the regex is proper, tried it on a few cases where it failed/passed accordingly.

I used this regex as we cannot use the validator, since we need to iterate over keys/values of a map, not just to validate a single string. However, if you know how to use the go validator for a map in this way, let me know :).

Despire (approved):
LGTM 👍

cloudziu (approved):
LGMT 👍

cloudziu (approved):

Despire (approved):
👍

PR #884, opened 2023-06-20
adding loadbalancing to provider examples

katapultcloud:


MiroslavRepka (approved):
LGTM just a quick question. Should we also add some application load balancer? For instance for port `443` or `80`?

PR #893, opened 2023-06-22
add FAQ page

### LINKED ISSUE #718: Add FAQ document explaining what Claudie is and what Claudie isn't
To clear the doubts. To explain, where Claudie makes sense and indicate where it doesn't.

```
an example:
Q: Does Claudie makes sense as a pure K8s orchestration on a single cloud-provider IaaS?
A: It will work well, however Claudie, being opinionated and focused on multi-cloud scenarios will come with its own set of downsides. The main one being the necessity to get a public IPv4 address for each node. However, if you're are building the cluster and you'd like to keep the options open to use some of the multi-cloud features in future (i.e. workload migration to another cloud provider, or extending the cluster with nodepools from another cloud provider in a multi-cloud fashion), then Claudie is the cluster provisioner for you.
```

@bernardhalas  agreed to help with creating the questions

The purpose of this task is to gather the questions and their answers into a markdown doc, put it next to https://github.com/berops/claudie/blob/master/docs/use-cases, and have this referenced from within ideally here https://github.com/berops/claudie/blame/master/README.md#L28

Despire:
closes #718

### DIFF
--- a/docs/faq/FAQ.md
+++ b/docs/faq/FAQ.md
@@ -23,14 +23,14 @@ Claudie doesn't make sense when you rely on specific features of a cloud provide
 
 ### Is there any networking performance impact due to the introduction of the VPN layer?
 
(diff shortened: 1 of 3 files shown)


bernardhalas (approved):
LGTM, thank you!

PR #899, opened 2023-06-24
Revert "fix href to feedback form"

bernardhalas:
Reverts berops/claudie#887

It seems this commit broke the feedback form (it worked earlier today). Therefore the revert.


Despire (approved):
LGTM

PR #903, opened 2023-06-26
Fix/grpc keepalive

### LINKED ISSUE #885: Feature: Make gRPC connections more robust
### Motivation

Currently, when there is a gRPC connection between services, and some network disturbance occurs, the connection will fail and the service will not retry to establish the connection again. This results in errors which could be prevented.

### Description

The issue aims to improve the robustness of the gRPC connections by adding retries and possibly something more.

### Exit criteria
- [ ] The gRPC connections are reasonably robust to handle minor network disruptions

Despire:
closes #885 

by setting the keepalive params for a client connection and adding a retry policy on RPC that fail due to code `unavailable` 

`Unavailable indicates the service is currently unavailable`

The client will try for roughly ~10 minutes for the service to become available and retry the RPC, given that our calls are idempotent this shouldn't be an issues.

It may happen that two calls at the same time will be executed (the one previosly started where the connection was lost and the new one), however. In terraformer we use a lock so that shouldn't be an issues and with the rest of the calls it shouldn't raise any negative side effects.

### DIFF
--- a/internal/utils/grpc.go
+++ b/internal/utils/grpc.go
@@ -2,10 +2,13 @@ package utils
 
 import (
 	"fmt"
-
+	grpc_retry "github.com/grpc-ecosystem/go-grpc-middleware/retry"
 	"github.com/rs/zerolog/log"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/keepalive"
+	"time"
 )
 
 // CloseClientConnection is a wrapper around grpc.ClientConn Close function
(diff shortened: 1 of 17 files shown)


internal/utils/grpc.go
```diff
@@ -16,10 +19,48 @@ func CloseClientConnection(connection *grpc.ClientConn) {
 }
 
 func GrpcDialWithInsecure(serviceName string, serviceURL string) (*grpc.ClientConn, error) {
-	cc, err := grpc.Dial(serviceURL, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	kacp := keepalive.ClientParameters{
+		Time:                60 * time.Second, // ping the server every ~60 seconds.
+		Timeout:             20 * time.Minute, // wait up to 20 minutes for a packet to be acknowledged.
```
MiroslavRepka, line 24: Isn't this value a bit high?

internal/utils/grpc.go
```diff
@@ -16,10 +19,48 @@ func CloseClientConnection(connection *grpc.ClientConn) {
...
+		PermitWithoutStream: true,             // send pings if there are no active streams (RPCs).
+	}
+
+	// retry policy for RPC calls.
+	opts := []grpc_retry.CallOption{
+		grpc_retry.WithBackoff(grpc_retry.BackoffLinearWithJitter(45*time.Second, 0.1)),
+		grpc_retry.WithMax(15),
```
MiroslavRepka, line 49: Ditto, arent 15 retries a bit too much?

internal/utils/grpc.go
```diff
@@ -16,10 +19,48 @@ func CloseClientConnection(connection *grpc.ClientConn) {
 }
 
 func GrpcDialWithInsecure(serviceName string, serviceURL string) (*grpc.ClientConn, error) {
```
MiroslavRepka, line 21: Is this function even used anywhere now?

internal/utils/grpc.go
Despire: 15 every 45 seconds is ~11 mins, which I think is okay-ish, we could half it to 6 minutes I would be okay either way

internal/utils/grpc.go
Despire: In two `_test` files I can also replaced them there and remove this function

internal/utils/grpc.go
Despire: It may be, we can halve it to 10 minutes, what number would you suggest ?

internal/utils/grpc.go
MiroslavRepka: I would probably prefer to halve it if thats okay

internal/utils/grpc.go
MiroslavRepka: Yeah, might as well replace it

internal/utils/grpc.go
MiroslavRepka: [gRPC defaults](https://grpc.io/docs/guides/keepalive/#keepalive-configuration-specification) seems to be 20s for both client and server.
image

Despire: I think 20s is a bit low, I would set it to 5 mins.

MiroslavRepka: Sure, let's set it to 5min then

MiroslavRepka (commented):
LGTM, thanks @Despire!

MiroslavRepka (approved):

cloudziu (approved):
Awesome! Thanks :D

Despire:
Had to also adjust the server settings as otherwise the server would just sent 

`[transport] Client received GoAway with error code ENHANCE_YOUR_CALM and debug data equal to ASCII "too_many_pings".`

internal/utils/grpc.go
```diff
@@ -18,6 +19,22 @@ func CloseClientConnection(connection *grpc.ClientConn) {
...
+		grpc.KeepaliveEnforcementPolicy(keepalive.EnforcementPolicy{
+			MinTime:             45 * time.Second, // If a client pings more than once every 45 seconds, terminate the connection
+			PermitWithoutStream: true,             // Allow pings even when there are no active streams
+		}),
+		grpc.KeepaliveParams(keepalive.ServerParameters{
+			MaxConnectionIdle:     math.MaxInt64,   // If a client is idle for INFINITE seconds, send a GOAWAY.
+			MaxConnectionAge:      math.MaxInt64,   // If any connection is alive for more than INIFINITE seconds, send a GOAWAY.
+			MaxConnectionAgeGrace: math.MaxInt64,   // Allow INIFNITE seconds for pending RPCs to complete before forcibly closing connections.
+			Time:                  2 * time.Hour,   // Ping the client if it is idle for 2 Hours to ensure the connection is still active.
+			Timeout:               5 * time.Minute, // Wait 5 minutes for the ping ack before assuming the connection is dead.
```
cloudziu, line 33: I'm not sure about that. If a service isn't responding for 5min I wouldn't expect any answer at all. The default value is 20seconds here. Is there a reason why we want to elevate it to 5min?

cloudziu (approved):

MiroslavRepka (approved):

PR #920, opened 2023-06-29
Use utils.GetClusterID where possible

### LINKED ISSUE #827: Chore: Use helper GetClusterID method where applicable
### Description
In *`internal/utils/log.go`* file, we have a function named `GetClusterID` -
```
func GetClusterID(clusterInfo *pb.ClusterInfo) string {
	if clusterInfo != nil {
		return fmt.Sprintf("%s-%s", clusterInfo.Name, clusterInfo.Hash)
	}
	return ""
}
```
I noticed many places inside the *`services`* folder, where this function can be used but is not used.

### Exit criteria
The `GetClusterID` function is used in those potential places.

- [ ] Inside *`services`* folder, wherever we have`fmt.Sprintf("%s-%s", clusterInfo.Name, clusterInfo.Hash)`, replace it with `utils.GetClusterID`.

cloudziu:
This PR resolves #827

### DIFF
--- a/manifests/claudie/kustomization.yaml
+++ b/manifests/claudie/kustomization.yaml
@@ -56,7 +56,7 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: ghcr.io/berops/claudie/ansibler
-  newTag: 26995c5-2140
+  newTag: 7c41cbc-2156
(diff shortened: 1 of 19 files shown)


Despire (approved):
LGTM 👍

MiroslavRepka (approved):
LGTM

PR #934, opened 2023-07-03
include arch label

### LINKED ISSUE #665: Feature: Implement `kubernetes.io/arch` labels for node templates in Cluster Autoscaler
Can be worked on after #644 is merged.

### Motivation

Cluster Autoscaler requests a template of a potential new node, which includes all labels it will have. Currently, this template does not include the `kubernetes.io/arch` label, as not all cloud libraries included information about architecture type. This issue aims to add this label to the node template.

### Description

The node template in the `node_manager` package should include the `kubernetes.io/arch` label, with the correct value.

### Exit criteria

- [ ] `kubernetes.io/arch` label is now part of the node template

Despire:
closes #665 

The API for retrieving the architecture isn't uniform across all cloud SDKs, 

For GCP, Azure The architecture is retrieved via the Image client and then the Architecture is set to all the cached types. For OCI a regex match is done for the ARM machine type, but since the ARM type on OCI is a Flex instance, which claudie doesn't surpport, I couldn't test it


bernardhalas:
Just curious, grepping the architecture from `/proc/cpuinfo` was not an option or `uname -i`? (It would also work on on-prem nodepools).

Despire:
Afaik, this is for a template that the autoscaler requests, and for retrieving it from `/proc/cpuinfo` I would need to ssh into an existing node of the cluster which would be more work that wouldn't be worth it. cc @MiroslavRepka

bernardhalas:
One more question then - does this label appear on the nodes?

(If so, we'll need to figure out how to set it for the on-prem nodes.)

MiroslavRepka:
Just to clarify/reiterate from the daily meeting. The label `kubernetes.io/arch` is automatically added to the node by kubelet. [ref](https://kubernetes.io/docs/reference/labels-annotations-taints/#kubernetes-io-arch)

However, as Cluster Autoscaler needs the template of the node, to simulate scheduling, we need to populate this label without creating a new node. Imagine the case where you have some pending pods, with the node selector set to `kubernetes.io/arch: arm64`. Before this PR, Cluster Autoscaler would not know if the potential new node has an appropriate value for this pod.

MiroslavRepka (approved):
LGTM

cloudziu (approved):
LGTM 👍

cloudziu (approved):

PR #951, opened 2023-07-12
Add security contexts claudie manifests

### LINKED ISSUE #844: Pod/Container security context in the manifests
We should perform security hardening of the manifests

cloudziu:
Closes #844 

Droped CAPs are only on part of the microservices, where I was able to ssh into the pod and check what linux capabilities are used. Ones that doesn't have that specified - do not have any shell installed

@UPDATE
- Dropped all `CAP_`'s
- Added the security context to other services (not only ones that have a shell)
- added additional constraints:
```yaml
            allowPrivilegeEscalation: false
            privileged: false
            readOnlyRootFilesystem: true           
```


manifests/claudie/terraformer.yaml
```diff
@@ -92,7 +121,7 @@ spec:
               valueFrom:
                 configMapKeyRef:
                   name: env
-                  key: AWS_REGION
+                  key: AWS_REGION
```
MarioUhrik, line 124: we can probably remove this diff

manifests/claudie/ansibler.yaml
```diff
@@ -18,8 +18,41 @@ spec:
...
+          - mountPath: /bin/services/ansibler/server/clusters
+            name: data
+          - mountPath: /.ansible
+            name: temp                  
+          securityContext:
+            capabilities:
+              drop:
+                - all
+              add:
+                - chown
```
MarioUhrik, line 42: Were you able to identify the minimum amount of capabilities in order to run Claudie components?
How did you manage this?

manifests/claudie/ansibler.yaml
```diff
@@ -18,8 +18,41 @@ spec:
         app.kubernetes.io/name: ansibler
     spec:
       terminationGracePeriodSeconds: 1320
+      securityContext:
+        runAsUser: 1000
```
MarioUhrik, line 22: How about further securityContext settings, e.g. not running in privileged mode, running as non-root, using read only root file system, etc ?

MarioUhrik:
> Ones that doesn't have that specified - do not have any shell installed

I am not 100% sure, but I feel as though we should deal with those too

manifests/claudie/ansibler.yaml
cloudziu: I followed this guide: https://cloud.redhat.com/blog/linux-capabilities-in-openshift

In short:
1. Exec into a pod and run `cat /proc/$PID/status | grep Cap`
2. Decode the hex in `capsh --decode`. This will output the cap used by the process
3. Apply them in the manifest and test :)

manifests/claudie/ansibler.yaml
cloudziu: That is a good point. I can look into that as well.

manifests/claudie/ansibler.yaml
cloudziu: So apparently I was wrong. I did a simple test and dropped all `CAP_`'s in each of the microservice - and I didn't saw any distruptions (tested localy). It must mean that the method I shared with you is describing something else. Like "what are the cap's that the current process haves" and not "what cap the current process is using".

MarioUhrik: gotcha, thanks Adam
That's exactly what I was wondering. Now it's clear :D

MarioUhrik (approved):

manifests/claudie/terraformer.yaml
```diff
@@ -92,7 +113,7 @@ spec:
...
```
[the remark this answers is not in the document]
cloudziu: IDK why... those irytating spaces are constantly sneaking in 😡 Thanks, I will fix that!

MarioUhrik (approved):

PR #958, opened 2023-07-18
Feat/backup claduie

### LINKED ISSUE #897: Feature: Claudie backup/restore
### Motivation

Claudie should be able to create a backup of its own data stores (MongoDB, DynamoDB, MinIO). Furthermore, Claudie should be able to restore itself from those backups.

### Exit criteria

- [ ] Claudie is able to create a backup
- [ ] Claudie is able to restore itself from the backup
- [ ] The process of creating a backup and restoring the Claudie is documented

Despire:
closes #897


MarioUhrik (approved):
LGTM, nice work Matus

For my curiosity, why the `kustomization.yaml` changes?

MarioUhrik:
> > why the kustomization.yaml changes?
> 
> Otherwise the secrets or config maps wouldn't be backed up and when restoring it would fail
> 
> Everything thats part of claudie should have the label so the backup will be complete

Perfect, thanks.

My point is, if the labels are missing from somewhere, we should add them to the manifests.
But yes, for configmaps and secrets, the only way to have those labels is via kustomize options, so your solution seems optimal

docs/creating-claudie-backup/creating-claudie-backup.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Backing up Claudie
+
+All resources that are deployed or created by Claudie can be identified with the following label:
+
+```
+    app.kubernetes.io/part-of: claudie
+```
+
+!!! note "If you want to include your deployed Input Manifests to be part of the backup you'll have to add the same label to them."
```
bernardhalas, line 25: Should we add this label to all the input manifest examples, so that by default the input manifest is backed-up too?
It's clear, that in the IaC world, this should be implicit, but I wonder if the non-IaC users would expect the backup to include the input manifest.

MarioUhrik: I agree 100%

Perhaps we can open an issue for that

Despire: I'll create an issue for that

bernardhalas: https://github.com/berops/claudie/issues/959

MarioUhrik: thanks, guys :)

PR #964, opened 2023-07-24
add lb check

Despire:


cloudziu (approved):
LGMT 👍 This should fix that problem

PR #966, opened 2023-07-25
comment setting upstream in Set up git author step

JKBGIT1:
This PR should fix not working release-docs pipeline by not using `git push --set-upstream origin master` in `Set up git author` step.
I have tested this in my own [repository](https://github.com/JKBGIT1/claudie-docs).
You can [see](https://github.com/JKBGIT1/claudie-docs/actions/runs/5656890589/job/15324807882), that pipeline did run successfully, when the `git push --set-upstream origin master` was commented. And you can also [see](https://github.com/JKBGIT1/claudie-docs/actions/runs/5656947668/job/15324984767), that it fails, when it wasn't commented.

If this PR gets merged I would like to rerun [this](https://github.com/berops/claudie/actions/runs/5507554544), just to make sure it is also working in claudie repository.

### DIFF
--- a/.github/workflows/release-docs.yml
+++ b/.github/workflows/release-docs.yml
@@ -31,12 +31,8 @@ jobs:
       
       - name: Set up git author
         run: |
-          remote_repo="https://${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
           git config --global user.name "${GITHUB_ACTOR}"
           git config --global user.email "${GITHUB_ACTOR}@users.noreply.github.com"
-          git remote rm origin
-          git remote add origin "${remote_repo}"
-          # git push --set-upstream origin master
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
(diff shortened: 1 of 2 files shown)


MarioUhrik:
Hi @JKBGIT1, I am struggling to understand the technicalities behind this.
Can you explain why removing the command makes everything work again?

JKBGIT1:
Hi @MarioUhrik TBH I don't know. Based on the error from the pipeline I came to a conclusion, that it is failing because of push command... So I tried it without that.

MarioUhrik:
I see, and why did we want to push in the first place?

JKBGIT1:
Hi @MarioUhrik that's a great point, because thanks to that I realized, why it was implemented this way.
We actually need to push to master branch, when there is new Changelog file. Because `mkdocs.yml` doesn't contain reference to this new Changelog file and the reference to it is added in this pipeline (if the reference will not be added, the Changelog file will not be displayed on docs.claudie.io). 
And when there isn't a new Changelog file, there won't be a push to the master branch, because `mkdocs.yml` isn't updated.

JKBGIT1:
Hi guys @Despire @cloudziu @MarioUhrik, the pipeline is fixed. I have tested both scenarios:

- with push to master branch ([see](https://github.com/JKBGIT1/claudie-docs/actions/runs/5713834057))
- without push to master branch ([see](https://github.com/JKBGIT1/claudie-docs/actions/runs/5713808320/job/15479890405))

The reason behind falling `git push` was missing `git fetch` and `git checkout master`

cloudziu (approved):

PR #981, opened 2023-08-01
limit spawned child processes

### LINKED ISSUE #692: Feature: Limit number of goroutines spawning other processes.
### Motivation

Currently, Claudie is building a lot of the infrastructure concurrently. This is great if there is a just a few processes running, however, once a certain threshold is reached, the pods run out of memory quickly. An example can be seen in `Terraformer`, when an e2e testing is running, the pod consumes above 1Gb of memory and once the pod's resources are consumed, the Terraform processes start to fail. Therefore, we should limit the number of goroutines which spawn other processes like terraform/ansible/kubeone, to ensure pods will not be running out of memory thus failing the processes.

### Description

Currently, there is no limit to how many processes (like terraform, ansible, ...) are running inside pods in Claudie. We should limit it to assure those processes will not fail due to insufficient memory.

### Exit criteria

- [ ] Services like `Terraformer`, `Ansibler` and `Kube-eleven` now limit how many processes are running concurrently.
- [ ] Each service has some sensible value set to how many processes it can tolerate
- [ ] Resources are adjusted to accommodate those concurrent processes

Despire:
closes #692 

Limit the number of go-routines (i.e. clusters) that spawn ansible, terraform, kubeOne to 5.
I've adjusted the Resource limits based on the changes.

### DIFF
--- a/manifests/claudie/ansibler.yaml
+++ b/manifests/claudie/ansibler.yaml
@@ -46,10 +46,10 @@ spec:
           resources:
             requests:
               cpu: 700m
-              memory: 768Mi
+              memory: 500Mi
             limits:
               cpu: 1024m
-              memory: 1248Mi
+              memory: 900Mi
           env:
             - name: ANSIBLER_PORT
               valueFrom:
--- a/manifests/claudie/kube-eleven.yaml
+++ b/manifests/claudie/kube-eleven.yaml
@@ -42,10 +42,10 @@ spec:
           resources:
             requests:
               cpu: 500m
-              memory: 500Mi
+              memory: 150Mi
             limits:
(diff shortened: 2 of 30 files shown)


bernardhalas:
Thank you for this one @Despire . Any deadlocks seen during your testing?

Despire:
>Thank you for this one @Despire . Any deadlocks seen during your testing?

@bernardhalas , everything seems to work fine, nothing gets stuck or any deadlocks.

cloudziu (approved):

MarioUhrik:
Hi guys, just wondering, are there any performance drawbacks to this PR?

PR #984, opened 2023-08-03
migrate to cilium

### LINKED ISSUE #961: Feature: Cilium POC
### Motivation
[Cilium](https://github.com/cilium/cilium) is a modern CNI implementation. It is powered by the eBPF features available in modern kernels and offers great observability features, which are typically not possible out-of-the-box when using Calico (despite the fact that Calico is also making a switch and some parts of it are being re-written in eBPF). Cilium has a growing ecosystem we'd like to plug into.

### Description
This issue is a consideration and a PoC for swapping Calico for Cilium. The main reasons why:
- support for [Hubble](https://github.com/cilium/hubble)
- support for [Cilium-mesh](https://docs.cilium.io/en/stable/#getting-started)
- support for [pwru](https://github.com/cilium/pwru)

This needs to be PoC'd, because Claudie is built on Wireguard (which is implemented using eBPF too) and in the past (2+ yrs ago) it wasn't compatible with Cilium. In the meantime, the compatibility has been worked on, but it's not clear if in the Claudie way it'll "just work".

### Exit criteria

- [ ] Claudie-like cluster is built using Wireguard and Cilium as a CNI
- [ ] Hubble and Cilium Mesh can be deployed and used without issues

Despire:
closes #961 

Afther some changes to the nodepools some of the cilium pods wouldn't get `ready` to solve this `kubectl rollout restart daemonset cilium` is executed after changes to the cluster (i.e deleting nodes or adding nodes)

Part of this PR is also using a new version of Longhorn (hasn't been updated for while)

for cilium this config was used in kubeone:

```yaml
clusterNetwork:
  kubeProxy:
    skipInstallation: true
  cni:
    cilium:
      enableHubble: false
      kubeProxyReplacement: "strict"
```

This basically installs cilium with eBPF, and doesn't install hubble by default. [Ref](https://github.com/kubermatic/kubeone/blob/main/docs/api_reference/v1beta1.en.md#ciliumspec)

[Tests](https://github.com/berops/claudie/actions/runs/5783585420) passed without issues


cloudziu (approved):
LGMT 👍

PR #1019, opened 2023-08-22
Feat/add timestamp to all stages in mongo

JKBGIT1:
This PR adds timestamp to all stages, which are stored in mongo. As we discussed [here](https://berops.slack.com/archives/C04JSCX9YCB/p1692694867566209), I have only keep it in `context-box`. However I am not 100% sure, if I have addressed everything and I would also like to have one more question, if I may. 
Is it okay, if we don't include `Timestamp` in `ConvertToGRPCWorkflow` func?


Despire (approved):
>if I may.
Is it okay, if we don't include Timestamp in ConvertToGRPCWorkflow func?

Yes this is fine. The idea is to only add the timestamp when saving to the DB, outside it has no usefulness as of now.

LGTM 👍

Despire:
Please format the code correctly before merging https://github.com/berops/claudie/actions/runs/5939012354

PR #1055, opened 2023-10-02
Chore/set new request and limits on services

JKBGIT1:
This PR sets new requests and limits on some Claudie services


cloudziu (approved):

katapultcloud (approved):
LGTM! Now we'll need about 1core and 2.5G of memory to be scheduled.

PR #1060, opened 2023-10-06
Update changelog-0.6.x.md

Despire:
prep for new release.


JKBGIT1 (approved):

JKBGIT1 (commented):
BTW shouldn't we also include something like new request and limits were set?
https://github.com/berops/claudie/pull/1055

cloudziu (approved):

PR #1095, opened 2023-11-10
Feat/metrics

### LINKED ISSUE #994: Claudie application metrics
Feature for Claudie to expose application-level metrics, that can be monitored with, for example, Prometheus. This feature should allow us to have more in-depth visibility of what Claudie is doing. Example metric ideas:
- number of concurrent jobs executed by a service
- number of clusters/inpuitmanifest currently managed
- number of warnings/errors on which base alerts can be created

Despire:
Closes #994 

For the gRPC services (Terraformer, Ansibler, Kube-Eleven, Kuber, ContextBox)

Only these metrics are collected https://github.com/berops/claudie/blob/fb94505bf236c272fa33cd107f6290e6851d037b/internal/utils/metrics/middleware.go#L16-L50

For Builder https://github.com/berops/claudie/blob/fb94505bf236c272fa33cd107f6290e6851d037b/services/builder/domain/usecases/metrics/metrics.go#L14-L98

For scheduler
https://github.com/berops/claudie/blob/fb94505bf236c272fa33cd107f6290e6851d037b/services/scheduler/domain/usecases/metrics/metrics.go#L8-L19

Additionally for contexbox two more (number of enqued) merics are collected

https://github.com/berops/claudie/blob/fb94505bf236c272fa33cd107f6290e6851d037b/services/context-box/server/domain/usecases/metrics/metrics.go#L8-L17

I believe this should give enough inside as to what each of the services are doing

Let me know if I should add anything more or delete some of these metrics

### DIFF
--- a/internal/utils/metrics/middleware.go
+++ b/internal/utils/metrics/middleware.go
@@ -30,7 +30,7 @@ var (
 
 	Latency = prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
-			Name: "claudie_grpc_grpc_call_latency_seconds",
+			Name: "claudie__grpc_call_latency_seconds",
 			Help: "Latency of gRPC API calls in seconds.",
 			Buckets: []float64{
 				.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10, // up to 10sec
--- a/internal/utils/metrics/middleware.go
+++ b/internal/utils/metrics/middleware.go
@@ -30,7 +30,7 @@ var (
 
 	Latency = prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
-			Name: "claudie__grpc_call_latency_seconds",
+			Name: "claudie_grpc_call_latency_seconds",
 			Help: "Latency of gRPC API calls in seconds.",
 			Buckets: []float64{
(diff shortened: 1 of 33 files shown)


cloudziu (approved):
The only thing that is bothering me inconsistent default metric port on the services.  Beside that LGTM 👍

cloudziu:
> I can make them the same i.e. 9090 and that shouldn't matter once they're behind a container just local deployment will be a hassle

I will be easier to define a `PodMonitor` / `ServiceMonitor` when the metric endpoint is listening on the same port, so I would appreciate that.

Despire:
Okay, will do.

cloudziu (approved):

internal/utils/metrics/middleware.go
```diff
@@ -0,0 +1,76 @@
...
+		prometheus.CounterOpts{
+			Name: "claudie_grpc_request_count",
+			Help: "Total number of gRPC API calls.",
+		},
+		[]string{MethodLabel},
+	)
+
+	Latency = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name: "claudie_grpc_grpc_call_latency_seconds",
```
JKBGIT1, line 33: Are there two `grpc` on purpose? I am currently working on the Grafana Dashboard and this just came up to my mind.

cloudziu (approved):

PR #1113, opened 2023-11-25
Hetzner: Only create node volume if size is > 0

### LINKED ISSUE #1110: Hetzner autoscaled nodepool without volume
It seems like i can't launch a Hetzner autoscaled nodepool without a volume attached.
I've tried without `storageDiskSize` set and, as shown, with 0 as value, but both creates a volume for the node, with 50gb.
This is the code:
```
      - name: compute-hetzner-autoscaled
        providerSpec:
          name: hetzner
          region: hel1
          zone: hel1-dc2
        serverType: cax21
        storageDiskSize: 0
        image: ubuntu-22.04
        autoscaler:
          min: 1
          max: 10
```

Danielss89:
This PR fixes #1110 

Another option is to use a boolean flag instead, like `createVolume` true|false.

### DIFF
--- a/services/terraformer/templates/aws/node.tpl
+++ b/services/terraformer/templates/aws/node.tpl
@@ -82,7 +82,7 @@ EOF
 }
 
 {{- if eq $.ClusterType "K8s" }}
-    {{- if not $nodepool.IsControl }}
+    {{- if not $nodepool.IsControl and $nodepool.NodePool.StorageDiskSize > 0 }}
 resource "aws_ebs_volume" "{{ $node.Name }}_volume" {
   provider          = aws.nodepool_{{ $nodepool.NodePool.Region }}
   availability_zone = "{{ $nodepool.NodePool.Zone }}"
--- a/services/terraformer/templates/azure/node.tpl
+++ b/services/terraformer/templates/azure/node.tpl
@@ -116,7 +116,7 @@ PROT
 }
 
 {{- if eq $.ClusterType "K8s" }}
-    {{- if not $nodepool.IsControl }}
+    {{- if not $nodepool.IsControl and $nodepool.NodePool.StorageDiskSize > 0 }}
 resource "azurerm_managed_disk" "{{ $node.Name }}_disk" {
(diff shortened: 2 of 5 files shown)


Danielss89:
I added fix for the other providers. 
I don't see any code about the disks in ansibler.

Danielss89:
I do see some logic in `services/autoscaler-adapter/node_manager/node_manager.go#71-74` but i'm unsure as what to do there, as it seems like it sets some default value if the value is 0 or below. Should it just return `nil` if value is 0 or below instead?

MiroslavRepka:
Hey there, as per documentation 
```
- storageDiskSize
Size of the storage disk on the nodes in the nodepool in GB. The OS disk is created automatically with predefined size of 100GB for kubernetes nodes and 50GB for Loadbalancer nodes.

Default value is 50, minimum value is 50. Value is used only for compute nodes.

This field is optional, however, if compute nodepool does not define it, default value is used for creation of storage disk. Control nodepools and Loadbalancer nodepools ignore this field.
```
which is implemented [here](https://github.com/berops/claudie/blob/aae3dca94b636661f3b00396787fa7a324580f78/internal/manifest/utils.go#L161C3-L161C3), thus the condition you added will not have any effect.

I would suggest that if user explicitly defines `storageDiskSize` as 0, then it should not get created. Should be relatively easy to implement, might give it a shot myself if I find some time over the weekend. :) 

cc @bernardhalas @cloudziu please let me know what do you think about this.

bernardhalas:
Thanks for elaborating @MiroslavRepka .

So let's say we start supporting `storageDiskSize: 0` on compute nodes. Would we need to adjust anything in Ansibler/Kuber?

MiroslavRepka:
I believe only terraformer and scheduler would require change. Terraformer needs to not create the volumes as well as not try to attach them (this is missing from this PR). Scheduler would have to respect 0 and not set default value.

Despire:
@Danielss89 thank you for your input. Closing this as resolved in https://github.com/berops/claudie/pull/1138

PR #1114, opened 2023-11-26
Add secret UTF-8 validation

### LINKED ISSUE #1101: Bug: scheduler stage error cluster with a static node defined
### Current Behaviour
Scheduler reports an error when a local (on premise) static node is defined and used in a cluster.  
The error persists and is continuously reported by claudie-operator even the manifest is edited so that it no longer consists the offending parts.  
Manifest delete command locks.

### Expected Behaviour
All stages go through and build completes with status=DONE

### Steps To Reproduce
1. Load an input manifest for private cluster like
```
apiVersion: claudie.io/v1beta1
kind: InputManifest
metadata:
  labels:
    app.kubernetes.io/part-of: claudie
  name: localmanifest
spec:
  kubernetes:
    clusters:
    - name: local-cluster
      network: 192.168.2.0/24
      pools:
        compute:
        - laptop-2
        control:
        - laptop-1
      version: v1.24.0
  nodePools:
    static:
    - name: laptop-1
      nodes:
      - endpoint: 192.168.122.51
        secretRef:
          name: static-node-key
          namespace: claudie
    - name: laptop-2
      nodes:
      - endpoint: 192.168.122.26
        secretRef:
          name: static-node-key
          namespace: claudie
```
3. scheduler log yields:
```
2023-11-10T08:53:45Z ERR Error while processing config error="error while saving the config: failed to save config via github.com/berops/claudie/proto/pb.ContextBoxServiceClient.SaveConfigScheduler-fm : rpc error: code = Internal desc = grpc: error while marshaling: string field contains invalid UTF-8"
```

Manifest definition file was checked against UTF-8 compliance and linted.

cloudziu:
closes #1101
This PR implements UTF-8 secret validation in `claudie-operator`.

### DIFF
--- a/services/claudie-operator/pkg/api/v1beta1/inputmanifest_helpers.go
+++ b/services/claudie-operator/pkg/api/v1beta1/inputmanifest_helpers.go
@@ -44,19 +44,14 @@ func (im *InputManifest) GetNamespacedNameDashed() string {
 // it is also validating if the string is a proper UTF-8 string
 func (pwd *ProviderWithData) GetSecretField(name SecretField) (string, error) {
 	if value, ok := pwd.Secret.Data[string(name)]; ok {
-		if !validateSecretUTF8(string(value)) {
+		// Ref: https://github.com/berops/claudie/issues/1101#issuecomment-1820793262
+		if !utf8.ValidString(string(value)) {
 			return "", fmt.Errorf("field %s is not a valid UTF-8 string", name)
 		}
 		return string(value), nil
 	} else {
 		return "", fmt.Errorf("field %s not found", name)
 	}
-}
-
-// validateSecretUTF8 validates whether s consists entirely of valid UTF-8-encoded runes
-// Ref: https://github.com/berops/claudie/issues/1101#issuecomment-1820793262
-func validateSecretUTF8(data string) bool {
-	return utf8.ValidString(data)
 }
 
 // GetStatuses returns the inputmanifest.Status field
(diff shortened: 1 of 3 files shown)


services/claudie-operator/pkg/api/v1beta1/inputmanifest_helpers.go
```diff
@@ -38,14 +41,24 @@ func (im *InputManifest) GetNamespacedNameDashed() string {
...
 		return "", fmt.Errorf("field %s not found", name)
 	}
 }
 
+// validateSecretUTF8 validates whether s consists entirely of valid UTF-8-encoded runes
+// Ref: https://github.com/berops/claudie/issues/1101#issuecomment-1820793262
+func validateSecretUTF8(data string) bool {
+	return utf8.ValidString(data)
+}
+
```
Despire, line 61: This is not needed. Calling `utf8.ValidString()` can be done directly.

Despire (commented):
Looks good 👍 

Did you also try it with invalid utf8 chars ?

In go they're replaced when calling some stdlib functions with valid characters.

https://stackoverflow.com/questions/36426327/why-utf8-validstring-function-not-detecting-invalid-unicode-characters

services/claudie-operator/pkg/api/v1beta1/inputmanifest_helpers.go
cloudziu: > Did you also try it with invalid utf8 chars ?
Yes, I was testing it by running `cat /dev/urandom | head -n 3 | base64` and putting the value into the secret. It wasn't passing the validation.

JKBGIT1 (approved):
LGTM :+1:

Despire (approved):
👍

PR #1119, opened 2023-11-27
Add note about own ingress controller to LB docs

Danielss89:


Despire (approved):
Thanks for the input 👍

Issue #1126, opened 2023-11-30
Bug: Claudie not processing new changes after an error


Danielss89:
I just hit a bug where Claudie is not processing newly applied changes from the input manifest after it has hit an error during something else.
I wanted to remove a static node from my cluster, and applied the modified input-manifest. Nothing seemed to happen so i looked at the status for Claudie and i found this error:
```
{
  "clusters": {
    "wox-cluster": {
      "message": "error in Kuber for cluster wox-cluster project default-wox : error while calling PatchNodes: rpc error: code = Unknown desc = error while patching labels on nodes for wox-cluster-oabexs3 : error while patching one or more nodes with labels",
      "phase": "KUBER",
      "state": "ERROR"
    }
  },
  "state": "ERROR"
}
```
This is the logs for kuber:
```
2023-11-28T23:43:07Z INF ../go/internal/command/cmd.go:57 > Next retry in 10s... module=kuber
2023-11-28T23:43:17Z WRN ../go/internal/command/cmd.go:145 > Retrying command kubectl patch node syd-1 -p '[{"op":"replace","path":"/metadata/labels/claudie.io~1pr
ovider-instance","value":"static-provider"}]' --kubeconfig '*****'... (1/3) module=kuber
wox-cluster-oabexs3    bash: fork: retry: Resource temporarily unavailable
wox-cluster-oabexs3    bash: fork: retry: Resource temporarily unavailable
wox-cluster-oabexs3    bash: fork: retry: Resource temporarily unavailable
wox-cluster-oabexs3    bash: fork: retry: Resource temporarily unavailable
wox-cluster-oabexs3    bash: fork: Resource temporarily unavailable
2023-11-28T23:43:32Z WRN ../go/internal/command/cmd.go:66 > Error encountered while executing kubectl patch node syd-1 -p '[{"op":"replace","path":"/metadata/label
s/claudie.io~1provider-instance","value":"static-provider"}]' --kubeconfig '*****' : exit status 254 module=kuber
2023-11-28T23:43:32Z INF ../go/internal/command/cmd.go:57 > Next retry in 20s... module=kuber
2023-11-28T23:43:52Z WRN ../go/internal/command/cmd.go:145 > Retrying command kubectl patch node syd-1 -p '[{"op":"replace","path":"/metadata/labels/claudie.io~1pr
ovider-instance","value":"static-provider"}]' --kubeconfig '*****'... (2/3) module=kuber
[...]

PR #1137, opened 2023-12-01
remove process substitution

Despire:
closes https://github.com/berops/claudie/issues/1126

The issue is when the combination `bash -c` `<()` is used while the entrypoint of the docker image is not a shell. Thus replace the proces subsitution `<()` by directly creating a temporary file.

for a quick verification of the issue one can try 

docker run -it --rm [docker.io/library/alpine:3.18.4](http://docker.io/library/alpine:3.18.4) tail -f /dev/null 

and in another terminal exec into the container and  execute a simple command like `bash -c 'echo <(echo 1)'` and look at the `ps` output, the subprocess won't be terminated

### DIFF
--- a/internal/kubectl/kubectl.go
+++ b/internal/kubectl/kubectl.go
@@ -42,14 +42,7 @@ func (k *Kubectl) KubectlApply(manifest string, options ...string) error {
 	if err != nil {
 		return err
 	}
-
-	if cleanup != nil {
-		defer func() {
-			if err := cleanup(); err != nil {
-				log.Err(err).Msg("failed to cleanup kubeconfig")
-			}
-		}()
-	}
+	defer cleanup()
 
 	command := fmt.Sprintf("kubectl apply -f %s %s", manifest, arg)
 	return k.run(command, options...)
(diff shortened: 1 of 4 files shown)


bernardhalas (approved):
I'm no Gopher, but just wondering if the copy/pasted code is worth delegating to a small function.

internal/kubectl/kubectl.go
```diff
@@ -186,15 +384,36 @@ func (k Kubectl) runWithOutput(command string, options ...string) ([]byte, error
...
+		return "", nil, nil
+	}
+
+	id := uuid.New()
+	tmpName := fmt.Sprintf("/tmp/k%s", hex.EncodeToString(id[:]))
+	if err := os.WriteFile(tmpName, []byte(k.Kubeconfig), os.ModePerm); err != nil {
+		return "", nil, err
 	}
+
+	cleanup := func() error { return os.Remove(tmpName) }
```
cloudziu, line 416:
```suggestion
cleanup := func() {
    if err := os.Remove(tmpName); err != nil {
        log.Err(err).Msg("failed to cleanup kubeconfig")
    }
}
```

And then do a shoret `defer`: 
```
if cleanup != nil {
		defer cleanup()
	}
```
WDYT?

cloudziu (approved):

internal/kubectl/kubectl.go
bernardhalas: Or even `defer cleanup(doCleanup)` so even the condition could be evaluated in the function itself.

cloudziu (approved):
LGTM 👍

bernardhalas (approved):
Thanks for the refactor @Despire

PR #1163, opened 2023-12-13
update cmd to obtain SSH key and node IPs

JKBGIT1:
This PR fix not working cmds to obtain SSH private key and node IPs from Claudie secret.


cloudziu (approved):
Good point!

PR #1168, opened 2023-12-21
Chore/increase requests and limits

JKBGIT1:
Update memory resources on `context-box`, `autoscaler-adapter` and `cluster-autoscaler`.


bernardhalas (approved):
Thanks for the adjustment!

Issue #1124, opened 2023-11-28
Bug: Terraformer fails if cloudflare DNS record for LB exists


Danielss89:
I'm playing around with Claudie and creating and deleting multiple clusters.
Now i hit a bug where the terraformer pod fails because the dns entry for the loadbalancer already exists in cloudflare.
I don't think it should fail when this happens. It should just continue the setup, as the entry exists with the correct values.
Here is a log:
```
2023-11-28T13:17:34Z WRN Retrying command terraform apply --auto-approve --parallelism=8 -json... (1/5) module=terraformer
2023-11-28T13:18:06Z WRN Error encountered while executing terraform apply --auto-approve --parallelism=8 -json : exit status 1 module=terraformer
2023-11-28T13:18:06Z ERR failed to execute cmd: terraform apply --auto-approve --parallelism=8 -json:
cloudflare_record.record-65-108-223-27: Creation errored after 30s
diagnostic:
    severity: error
    summary: expected DNS record to not already be present but already exists
    address: cloudflare_record.record-65-108-223-27
 module=terraformer
2023-11-28T13:18:06Z INF Next retry in 20s... module=terraformer
2023-11-28T13:18:26Z WRN Retrying command terraform apply --auto-approve --parallelism=8 -json... (2/5) module=terraformer
2023-11-28T13:18:57Z WRN Error encountered while executing terraform apply --auto-approve --parallelism=8 -json : exit status 1 module=terraformer
2023-11-28T13:18:57Z ERR failed to execute cmd: terraform apply --auto-approve --parallelism=8 -json:
cloudflare_record.record-65-108-223-27: Creation errored after 30s
diagnostic:
    severity: error
    summary: expected DNS record to not already be present but already exists
    address: cloudflare_record.record-65-108-223-27
 module=terraformer
2023-11-28T13:18:57Z INF Next retry in 40s... module=terraformer
2023-11-28T13:19:37Z WRN Retrying command terraform apply --auto-approve --parallelism=8 -json... (3/5) module=terraformer
2023-11-28T13:20:12Z WRN Error encountered while executing terraform apply --auto-approve --parallelism=8 -json : exit status 1 module=terraformer
2023-11-28T13:20:12Z ERR failed to execute cmd: terraform apply --auto-approve --parallelism=8 -json:
cloudflare_record.record-65-108-223-27: Creation errored after 30s
diagnostic:
    severity: error
    summary: expected DNS record to not already be present but already exists
[...]

PR #1171, opened 2023-12-29
Fix/dns

Despire:
attemp at resolving #1124.

I wasn't able to reproduce the issue where the DNS record wouldn't be deleted by Claudie, however there might be a scenario and thus I've separated the deletion of the LB infra and K8s infra.


bernardhalas (approved):
LGTM. Thank you.

PR #1176, opened 2024-01-08
Cluster-autoscaler tag validation

### LINKED ISSUE #1122: Feature: Restrict k8s version on patch version
I just tried to deploy a cluster for v1.26.8 but i hit an error for the loadbalancer, as it can't pull the cluster-autoscaler image for v1.26.8, as it doesn't exist. Newest version of the cluster-autoscaler is 1.26.5. It seems as it doesn't follow k8s versioning as k8s newest patch for v1.26 is 11(v1.26.11).
So Claudie should probably do a check on the patch version of the cluster the user tries to deploy(i know it checks minor version already) and deny the user to deploy the cluster.

cloudziu:
closes #1122 
This PR implements the logic for choosing upstream cluster-autoscaler components. It will lookup cluster-autoscaler container registry and choose the latest PATCH version of defined Kubernetes version.

### DIFF
--- a/services/kuber/server/domain/utils/autoscaler/cluster_autoscaler.go
+++ b/services/kuber/server/domain/utils/autoscaler/cluster_autoscaler.go
@@ -1,6 +1,7 @@
 package autoscaler
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -210,9 +211,13 @@ func getClusterAutoscaleVersions() ([]string, error) {
 	var TagsResponse TagsResponse
 
 	// Query CA registry
-	response, err := http.Get(clusterAutoscalerRegistry)
+	req, err := http.NewRequestWithContext(context.TODO(), http.MethodGet, clusterAutoscalerRegistry, nil)
 	if err != nil {
-		return nil, err
+			return nil, err
+	}
+	response, err := http.DefaultClient.Do(req)
+	if err != nil {
+			return nil, err
 	}
 	defer response.Body.Close()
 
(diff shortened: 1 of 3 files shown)


services/kuber/server/domain/utils/autoscaler/cluster_autoscaler.go
```diff
@@ -147,7 +158,90 @@ func getK8sVersion(version string) (string, error) {
...
+
+	// Query CA registry
+	req, err := http.NewRequestWithContext(context.TODO(), http.MethodGet, clusterAutoscalerRegistry, nil)
+	if err != nil {
+		return nil, err
+	}
+	response, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
```
Despire, line 221: This could by replaced by 

```go
resp, err := http.Get(clusterAutoscalerRegistry)
if err != nil {
...
}
...
```

Despire (approved):
Looks Great 👍 .

services/kuber/server/domain/utils/autoscaler/cluster_autoscaler.go
cloudziu: Initiali I did it like that but `golinter` was complaining 😢 
https://github.com/berops/claudie/commit/51994d4293f75bdddb53167751ace90eb0b554f7

Despire: I see, linter seems too strict on this one.

We can keep this then.

PR #1182, opened 2024-01-12
Revert "add "proxy_protocol on;" to the LoadBalancer nginx config"

JKBGIT1:
Reverts berops/claudie#1178

`proxy_protocol on;` is responsible for an error in `kube-eleven` when Claudie wants to build an `InputManifest` with the LoadBalancer cluster.

### DIFF
--- a/services/ansibler/templates/conf.gotpl
+++ b/services/ansibler/templates/conf.gotpl
@@ -9,7 +9,6 @@ upstream {{ $role.Role.Name }}{
 server  {
     listen {{ $role.Role.Port }};
     proxy_pass {{ $role.Role.Name}};
-    proxy_protocol on;
     proxy_next_upstream on;
 }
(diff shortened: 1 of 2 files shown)


cloudziu (approved):

services/ansibler/templates/conf.gotpl
```diff
@@ -9,8 +9,7 @@ upstream {{ $role.Role.Name }}{
 server  {
     listen {{ $role.Role.Port }};
     proxy_pass {{ $role.Role.Name}};
-    proxy_protocol on;
```
Despire, line 12: Shouldn't this line be part of the `listen` statement ?

`listen {{ $role.Role.Port }} proxy_protocol`

I think 
`proxy_protocol on;` on its own does something else.

Despire (approved):

Issue #962, opened 2023-07-21
Feature: Roadwarrior on-prem mode PoC


bernardhalas:
### Motivation

Claudie supports a sub-set of hybrid-cloud use cases. The common denominator of these use cases is the need that Claudie can directly reach the interfaces of all the nodes that compose the cluster (either on public or private IPs).

However, currently, we don't support a hybrid-cloud scenario where Claudie runs in the cloud, and it allows a Claudie-made cluster to be composed of nodes that are behind NAT, having just private IPs and not being directly reachable by Claudie. This is what some VPN implementations call a "Roadwarrior" mode.

### Description

Figure out the networking and connectivity challenges for building clusters that would allow joining nodes that are not directly reachable, but hidden "behind NAT". Propose a technological idea on how to implement Claudie cluster creation workflow.

### Exit criteria

- [ ] Networking concept for joining Roadwarrior nodes is agreed-upon among the team
- [ ] Workflow for joining Roadwarrior nodes is proposed
- [ ] Proof-of-concept is executed manually, and the feedback is processed by the team

Issue #815, opened 2023-05-23
Feature: support for external s3, dynamodb and mongodb


katapultcloud:
### Motivation

Using external services will make Claudie migrations much easier. Also if users decide to use external service to store sensitive content, they can do that.

### Description

Claudie supports storing Claudie output in external services, by defining option to use external services with correct access keys and or connections strings.

### Exit criteria

- [ ] s3
- [ ] dynamodb
- [ ] mongodb

Danielss89:
I can see this be valuable, as a user might want to move Claudie to a cluster managed by Claudie.
So user sets up a throw-away cluster, installs Claudie, sets up new cluster with Claudie, moves Claudie to new cluster, removes throw-away cluster.

PR #1184, opened 2024-01-17
Fix/static nodes

### LINKED ISSUE #1141: Bug: Prevent same static node being added to multiple clusters
### Current Behaviour
Currently it seems like nothing prevents the same static node being a part of multiple cluster manifests. We should prevent such a scenario from happening.

### Expected Behaviour
If a node is added into a cluster before it's properly decommissioned from the previous cluster, the operation will not happen until a successful decommissioning is done in the first place. Respective log error message is emitted.

Despire:
Closes #1141 

Check if the same static node is re-used.

1. Within the same config (by tracking the nodepool name)
2. Across different configs (by tracking the IP of every node) No two static nodes can have the same IP, thus if the same IP is used in different config this means the same static node is referenced.

This will work for now, but once https://github.com/berops/claudie/issues/962 is being worked on this should be revisited.

### DIFF
--- a/.github/workflows/CI-pipeline.yml
+++ b/.github/workflows/CI-pipeline.yml
@@ -315,9 +315,14 @@ jobs:
           
           for file in test-set5/*; do
               if [ -f "$file" ]; then
-                ENDPOINT=${IP_ARR[0]} ./yq e '.spec.nodePools.static.[0].nodes.[0].endpoint = strenv(ENDPOINT)' $file -i 
-                ENDPOINT=${IP_ARR[1]} ./yq e '.spec.nodePools.static.[1].nodes.[0].endpoint = strenv(ENDPOINT)' $file -i
-                ENDPOINT=${IP_ARR[2]} ./yq e '.spec.nodePools.static.[1].nodes.[1].endpoint = strenv(ENDPOINT)' $file -i
+                  filename=$(basename "$file")
+                  if [[ $filename == "1.yaml" ]]; then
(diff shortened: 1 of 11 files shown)


bernardhalas:
Hi @Despire,
Will this cover also the scenario when a different Claudie instance will try to use a static node which is under control by another Claudie control plane?

Despire:
>Will this cover also the scenario when a different Claudie instance 

Do you mean 2 claudie deployed in two management clusters ? If so they share the same mongoDB then yes, otherwise no.

Despire:
@bernardhalas 

Do you want to implement this ? 'Cause this might be a bigger issue, tracking all of the static nodes across different claudie instances running on different management clusters.

bernardhalas:
> > Will this cover also the scenario when a different Claudie instance

> Do you mean 2 claudie deployed in two management clusters ? If they share the same mongoDB then yes, otherwise no.

Can we have 2 claudie instances that would share the same DB?

> Do you want to implement this ? 'Cause this might be a bigger issue, tracking all of the static nodes across different claudie instances running on different management clusters.

If we had each Claudie instance with its ID, it would be pretty easy to leave a tag on the node saying to which instance it belongs. But I assume it's OK like this for now.

Despire:
> Can we have 2 claudie instances that would share the same DB?

Once this issue https://github.com/berops/claudie/issues/815 is implemented than it would be possible to share a common database.

> If we had each Claudie instance with its ID, it would be pretty easy to leave a tag on the node saying to which instance it belongs. But I assume it's OK like this for now.

I can open an issue about this so we can discuss this further

cloudziu (approved):

PR #1196, opened 2024-01-31
fix wrong cmds to obtain loadbalancer private key and IPs

JKBGIT1:
This PR just fixes wrong cmds to obtain the LB private key and LB IPs in the docs.


cloudziu (approved):
LGTM 👍

PR #1211, opened 2024-02-06
Fix removing state lock from dynamodb

### LINKED ISSUE #1190: Bug: Claudie not removing state lock in dybamodb
### Current Behaviour
While I was working on #815 I noticed that Claudie isn't removing state locks from DynamoDB. Additionally it will log a success even if the lock isn't removed. I suspect that it migh be related to this:
https://github.com/berops/claudie/blob/d5cd0105309deba1fffc00cb1a7b4725acb5fb87/services/terraformer/server/adapters/outbound/dynamodb_adapter.go#L86
Where part of the state lock address is `minioBucketName` instead of `dynamoDBTableName`.

### Expected Behaviour
Claudie properly removed state lock from dynamodb. If the state isn't removed it will log an error

cloudziu:
closes #1190

### DIFF
--- a/manifests/claudie/kustomization.yaml
+++ b/manifests/claudie/kustomization.yaml
@@ -74,4 +74,4 @@ images:
 - name: ghcr.io/berops/claudie/scheduler
   newTag: 99460eb-2579
 - name: ghcr.io/berops/claudie/terraformer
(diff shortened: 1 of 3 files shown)


Despire (approved):
LGTM 👍

PR #1215, opened 2024-02-08
Fix operatur status message.

### LINKED ISSUE #1207: Bug: Misleading status for InputManifests
### Current Behaviour

When you successfully create a cluster (Hetzner for example) and adjust the config to add another cloud provider (Genesis Cloud for example) but with bad credentials such (so that the creation will fail). After the failure you delete the input manifest and after all the retries the state of it will be `IN_PROGRESS` although in the DB it is in ERROR

Screenshot 2024-02-02 at 12 45 58

Screenshot 2024-02-02 at 12 46 08

### Expected Behaviour

The status should be in `ERROR` as well.

### Steps To Reproduce
1. Build a successful cluster (Hetzner for example)
2. Add another cloud provider (Genesis Cloud for example) but with bad credentials
3. The workflow will fail, Delete the inputmanifest and after all the replies check the state  via `kubectl get inputmanifests`
4. It will show you `IN_PROGRESS` although when you look into the DB it will be in ERROR

cloudziu:
Closes #1207

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ builder:
 	GOLANG_LOG=debug PROMETHEUS_PORT=9092 go run ./services/builder
 # Start Terraformer service on a local environment, exposed on port 50052
 terraformer:
-	GOLANG_LOG=debug PROMETHEUS_PORT=9093 go run ./services/terraformer/server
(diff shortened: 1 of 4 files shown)


Despire (approved):
LGTM 👍

Issue #1217, opened 2024-02-09
Bug: Custom StorageClass created by Claudie fails to provision PV


JKBGIT1:
### Current Behaviour

Custom StorageClass created by Claudie fails to provision PV after creating a PVC that uses this StorageClass.

### Expected Behaviour
 
A PVC should be provisioned when one creates a PVC using a custom StorageClass created by Claudie. 

### Steps To Reproduce
1. Use the InputManifest below.
```yaml
apiVersion: claudie.io/v1beta1
kind: InputManifest
metadata:
  name: gcp-manifest
  labels:
    app.kubernetes.io/part-of: claudie
spec:
  providers:
    - name: gcp-1
      providerType: gcp
      secretRef:
        name: gcp-secret
        namespace: default
  nodePools:
    dynamic:
      - name: control-gcp-1
        providerSpec:
          name: gcp-1
          region: europe-west1
          zone: europe-west1-c
        count: 1
        serverType: e2-medium
        image: ubuntu-os-cloud/ubuntu-2204-jammy-v20221206
      - name: compute-gcp-1
        providerSpec:
          name: gcp-1
          region: europe-west3
          zone: europe-west3-a
        count: 3
        serverType: e2-medium
        image: ubuntu-os-cloud/ubuntu-2204-jammy-v20221206
        storageDiskSize: 50
  kubernetes:
    clusters:
      - name: jakub-testing
        version: v1.26.0
        network: 192.168.2.0/24
        pools:
          control:
            - control-gcp-1
          compute:
            - compute-gcp-1
```
2. Deploy the following PVC in the Claudie-managed cluster.
```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: test-longhorn-pvc
spec:
  storageClassName: longhorn-gcp-1-zone
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 1Gi
```
3. The PVC will be stuck in the `Pending` state and fail to provision a PV. Below you can see the logs from the Events of the created PVC.
```
[...]

PR #1219, opened 2024-02-12
fix not working custom storage classes

JKBGIT1:
closes https://github.com/berops/claudie/issues/1217

`fromBackup` needs to have an empty value ([see](https://longhorn.io/docs/1.6.0/references/storage-class-parameters/#from-backup-field-parametersfrombackup)). We used `false` before.

was tested locally, and everything worked.


Despire (approved):
LGTM 👍

cloudziu (approved):

PR #1232, opened 2024-02-19
Add custom annotations on NodePools

### LINKED ISSUE #1216: Feature: Support Annotations in the InputManifest
### Motivation
If an end-user wants to set Longhorn to only use storage on a specific set of nodes, he/she needs to mark them with a specific value in the `node.longhorn.io/default-node-tags` annotation. Currently it is only possible to do that through the Longhorn UI, kubectl, etc., but we do not support adding annotations to the nodes through the InputManifest. This can potentially results in removing the manually added annotations, when Claudie runs any workflow.

### Exit criteria

- [ ] Allow end-users to annotate their nodes through the InputManifest.

cloudziu:
close #1216


bernardhalas:
While I was looking for the syntax of this feature I saw that this PR is missing the update in the docs (for sure at least the API reference update should reflect this change).

cloudziu:
@bernardhalas Good point! Also I noticed that I should update E2E tests, and add example annotations there.

docs/input-manifest/api-reference.md
```diff
@@ -246,6 +246,12 @@ Dynamic nodepools are defined for cloud provider machines that Claudie is expect
   
   To see the default labels Claudie applies on each node, refer to [this section](#default-labels).
 
+- `annotations`
+
+  Map of user defined annotations, ch will be applied on every node in the node pool. This field is optional.
```
Despire, line 251: `ch will be applied ...` I believe `ch` is a typo

docs/input-manifest/api-reference.md
```diff
@@ -298,6 +304,13 @@ Static nodepools are defined for static machines which Claudie will not manage.
...
```
Despire, line 309: ditto

services/kuber/server/domain/utils/nodes/patch.go
```diff
@@ -89,6 +89,27 @@ func (p *Patcher) PatchLabels() error {
...
 
+func (p *Patcher) PatchAnnotations() error {
+	var err error
+	for _, np := range p.desiredNodepools {
+		nodeAnnotations := np.Annotations
+		for _, node := range np.Nodes {
+			nodeName := strings.TrimPrefix(node.Name, fmt.Sprintf("%s-", p.clusterID))
+			for key, value := range nodeAnnotations {
+				patchPath, err1 := buildJSONPatchString("replace", "/metadata/annotations/"+key, value)
+				if err1 != nil {
```
Despire, line 100: This will be repeated multiple times for each node. Doesn't this mean that only the last annotation will be applied ? since by calling 
`buildJSONPatchString("replace", "/metadata/annotations/"+key, value)` it will be replaced

cloudziu: The replace address will be `/metadata/annotations/annotation_key` so only the annotation itself can be replaced. Its not replacing the whole `/metadata/annotations` each time. The same way labales are working atm.

Despire: Yes, sorry I missed the `+key` part

PR #1245, opened 2024-02-27
Update healthcheck endpoints

### LINKED ISSUE #1236: Bug: `builder` and `context-operator` ReadinessProbe failing
### Current Behaviour
When Claudie is deployed in `kind` locally, without any modification to the upstream deployment, `builder` and `claudie-operator` often fall into `Not Ready` state.

### Expected Behaviour
Claudie services should come up quickly and should not lose readiness.

### Steps To Reproduce
1. `kind create cluster --name=claudie-bootstrap`
2. `kubectl apply -f kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.12.0/cert-manager.yaml`
3. Wait 2-3 mins for the CRD registration
4. `kubectl apply -f https://github.com/berops/claudie/releases/latest/download/claudie.yaml`
5. Observe readiness for `builder` and `claudie-operator`

### Anything else to note
Increasing the timeout in the  readiness probes from 1 to 3 seconds mitigates the issue, but:
`builder` doesn't have CPU limits, so in case of this service, this is not about resources

cloudziu:
closes #1236
This PR is changing Readines probes to Liveness probes in some of the services.
It also fixes healthchecks for for scheduler, which wasn't working at all.


Despire (approved):
LGTM 👍

PR #1280, opened 2024-03-17
Update detailed-guide.md

Despire:


cloudziu (approved):
LGTM 👍 @JKBGIT1 so now we can trigger update docs with a command in comment, correct?

bernardhalas:
We should be more careful with PRs and their reviews. There are still references to old API in the docs.

bernardhalas (changes requested):
There are still references to old API in:
`docs/contributing/local-testing.md` and
`docs/storage/storage-solution.md`

cloudziu:
hmmm, good point. Initially, I reviewed only the changes present in the PR itself

bernardhalas (approved):

PR #1293, opened 2024-03-25
update healthchecks for builder

### LINKED ISSUE #1292: Bug: `builder` is restarted upon connectivity loss
### Current Behaviour
`builder` connects to services `context-box`, `terraformer`, `ansibler`, `kube-eleven` and `kuber`. The connection handling among these services varies.

If a connection to `context-box` is lost, `builder` tries to re-establish it. However, if a connection to the other four services is lost, `builder` doesn't attempt to reconnect, but relies on NOK status sent to `/healthz` endpoint, which is read by the liveness probe, which in turn restarts the `builder` service. This is suboptimal, because if there are frequent network interrupts, then there are frequent `builder` service restarts and the overall UX is impacted.

### Expected Behaviour
Connection recovery to all five services is handled in the same way.  `builder` tries to reconnect for some time (ideally configurable, but with a reasonable default value if not provided). After this times out, `builder` reports NOK to the liveness probe endpoint and reliest on a K8s-triggered restart.

### Steps To Reproduce
Ref: https://github.com/berops/claudie/issues/1287

Despire:
Closes #1292 

After 4 minutes of trying to reconnect, builders will start returning health issues

### DIFF
--- a/services/builder/main.go
+++ b/services/builder/main.go
@@ -21,8 +21,6 @@ import (
 	"github.com/rs/zerolog/log"
 
 	"golang.org/x/sync/errgroup"
-
-	"google.golang.org/grpc/connectivity"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


services/builder/main.go
```diff
@@ -32,23 +32,53 @@ const (
...
+	healthCheck := func(check *HealthCheck) {
+		now := time.Now()
+		updateTimeSinceFailure(&now, &check.timeSinceTerraformFailure, usecases.Terraformer.PerformHealthCheck())
+		updateTimeSinceFailure(&now, &check.timeSinceAnsiblerFailure, usecases.Ansibler.PerformHealthCheck())
+		updateTimeSinceFailure(&now, &check.timeSinceKubeElevenFailure, usecases.KubeEleven.PerformHealthCheck())
+		updateTimeSinceFailure(&now, &check.timeSinceKuberFailure, usecases.Kuber.PerformHealthCheck())
+		updateTimeSinceFailure(&now, &check.timeSinceContextBoxFailure, usecases.ContextBox.PerformHealthCheck())
+	}
+	checkFailure := func(t *time.Time, service string, perr error) error {
+		if t != nil && time.Since(*t) >= 4*time.Minute {
```
cloudziu, line 60: Isn't 4 minutes bit to long? If the healtheck is performed every 30s, it will require 8 failures.

cloudziu (approved):
LGTM 👍 , left one minor comment

bernardhalas:
Based on the grooming discussion with @cloudziu , we understood that connection handling to `context-box` was different compared to other services (`terraformer`, `ansibler`, `kuber`, `kube-eleven`). The change here modifies the `context-box` connector too. Could you clarify if our initial expectation from #1292 was wrong?

services/builder/main.go
bernardhalas: I think 4 mins is not high. The connectivity problem is not directly linked to the health of the service. In most cases, the connectivity status should not be affected by a service restart at all.

Despire:
> The change here modifies the context-box connector too. Could you clarify if our initial expectation from https://github.com/berops/claudie/issues/1292 was wrong?

The healthcheck was performed same for context-box, there was just an additional check that would monitor the connection before polling the next config from contextbox

Despire:
I've update the healthcheck

It will now monitor the connections every 5 seconds

And it will not pull any new configs until all of the services (terraformer, ansibler etc...) are healthy. (so not only checking context box before polling a new config but all of the servives)

services/builder/main.go
Despire: I think 4mins is enough, I've tested this and it takes just >90secs to restart terraformer

cloudziu (approved):

PR #1313, opened 2024-04-02
Update dependencies

Despire:
This PR updates pending dependnecies.

Also slightly changes the timeout of grpc Connections.

Now a ping is send every ~20 seconds and waiting ~20seconds for a ack before assuming the connection is dead.

Builder now also refreshes the connections to each service every ~20secs, which previosly was at ~5secs https://github.com/berops/claudie/pull/1293


cloudziu (approved):

Despire:
Screenshot 2024-04-02 at 13 25 21

The pods have been running for almost 2 hours and no restarts occured

PR #1324, opened 2024-04-04
Add namespace filtering in operator. Update CI-pipeline. Update docs

### LINKED ISSUE #1266: Feature: Add support for namespace filtering for Claudie Operator
### Motivation
Currently `claudie-operator` needs cluster-wide access because it watches input manifests and secrets across the whole cluster. We need to be able to tell it to read only some namespaces. The reasons why is:
- security (limiting its read access only to necessary namespaces)
- ability to run 2 instances of Claudie in the same cluster (where each would watch a different sent of namespaces)

### Description
There are multiple ways how to implement the namespace filter. Let's research them within the task and discuss here the best option to choose.

### Exit criteria

- [ ] Namespace filtering implemented
- [ ] Docs updated

cloudziu:
close #1266

### DIFF
--- a/docs/getting-started/detailed-guide.md
+++ b/docs/getting-started/detailed-guide.md
@@ -97,6 +97,14 @@ This detailed guide for Claudie serves as a resource for providing an overview o
     terraformer-fd664b7ff-dd2h7    1/1     Running     0               8m9s
     ```
 
+    !!! warning "Changing the namespace" 
+        By default, Claudie will deploy in the `claudie` namespace, and it will watch for `Input Manifest` and provider `Secrets` in this namespace. If you would like to add additional namespace to watch - overwrite `CLAUDIE_NAMESPACES` environment variable in `claudie-operator` deployment. Example:
+        ```
+        env:
+          - name: CLAUDIE_NAMESPACES
+            value: "claudie,different-namespace"
+        ``` 
+
     !!! warning "Troubleshoot!" 
         If you experience problems refer to our [troubleshooting guide](https://docs.claudie.io/latest/troubleshooting/troubleshooting/). 
 
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -14,3 +14,10 @@ func SanitiseString(s string) string {
 	sanitised = strings.ReplaceAll(sanitised, "_", "-")
 	return sanitised
 }
+
+// watchNamespaceList takes a string input of the form "namespace1,namespace-2,namespace3"
+// and splits it into a slice of strings based on the comma separator.
+// It returns the slice containing individual namespace strings.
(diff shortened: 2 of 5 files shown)


internal/utils/utils.go
```diff
@@ -14,3 +14,10 @@ func SanitiseString(s string) string {
 	sanitised = strings.ReplaceAll(sanitised, "_", "-")
 	return sanitised
 }
+
+// watchNamespaceList takes a string input of the form "namespace1,namespace-2,namespace3"
```
JKBGIT1, line 18: Did you add `-` in namespace2 for any particular reason?

services/claudie-operator/main.go
```diff
@@ -35,20 +37,26 @@ const (
...
 var (
-	// port is the port number that the server will serve. It will be defaulted to 9443 if unspecified.
+	// portStr is the port number that the server will serve. It will be defaulted to 9443 if unspecified.
 	portStr string
 	// certDir is the directory that contains the server key and certificate. The server key and certificate.
 	certDir string
 	// path under which the validation webhook will serve
 	webhookPath string
+	// namespaceSelector filters namespaces to watch for inputManifest resources
+	// takes a string input of the form "namespace1,namespace-2,namespace3"
```
JKBGIT1, line 47: Same question with the hyphen if I may.

JKBGIT1 (approved):
LGTM :+1:

internal/utils/utils.go
JKBGIT1: Thanks. I asked because it confused me a little.

docs/getting-started/detailed-guide.md
```diff
@@ -97,6 +97,14 @@ This detailed guide for Claudie serves as a resource for providing an overview o
     terraformer-fd664b7ff-dd2h7    1/1     Running     0               8m9s
     ```
 
+    !!! warning "Changing the namespace" 
+        By default, Claudie will deploy in the `claudie` namespace, and it will watch for `Input Manifest` and provider `Secrets` in this namespace. If you would like to add additional namespace to watch - overwrite `CLAUDIE_NAMESPACES` environment variable in `claudie-operator` deployment. Example:
```
bernardhalas, line 101:
```suggestion
        By default, Claudie will monitor all namespaces, and it will watch for `Input Manifest` and provider `Secrets` in the cluster. If you would like to limit the namespaces to watch - overwrite `CLAUDIE_NAMESPACES` environment variable in `claudie-operator` deployment. Example:
```

bernardhalas (changes requested):
Let's change the default mode when no value is set in the env var to a cluster-wide behavior.

bernardhalas (approved):
LGTM, thanks for the adjustment

PR #1328, opened 2024-04-08
increase worker_connections per worker process

### LINKED ISSUE #1327: Chore: Increase number of LB worker threads until a more robust configuration feature is added
### Description
We're running the LB services on the VMs, which have a capacity to run many more worker threads than the default value. Some our users are hitting the limits and get errors like:
```
2024/04/01 21:29:00 [alert] 677222#677222: 768 worker_connections are not enough
```

Until we find a better way how to configure the LB options, let's increase significantly the default value of the worker connections.

### Exit criteria

- [ ] Increase default value of `worker_connections`

Despire:
Closes #1327


bernardhalas (approved):
LGTM, thanks for the quick action here!
FYI @Danielss89

Despire:
Each worker process will have now a limit of 65535 possible connections (instead of the current 768 default)
```
oot        7956       1  0 11:15 ?        00:00:00 nginx: master process /usr/sbin/nginx -g daemon on; master_process on;
www-data    7957    7956  0 11:15 ?        00:00:00 nginx: worker process
www-data    7958    7956  0 11:15 ?        00:00:00 nginx: worker process
root        7981    6205  0 11:23 pts/1    00:00:00 grep --color=auto nginx
root@ip-10-0-1-208:/home/ubuntu# cat /proc/7957/limits /proc/7958/limits 
Limit                     Soft Limit           Hard Limit           Units     
...
Max open files            65535                65535                files     
...  
Limit                     Soft Limit           Hard Limit           Units     
...
Max open files            65535                65535                files     
...

```

Issue #1297, opened 2024-03-27
Feature: For static nodes, allow other usernames with root access.


Rohithzr:
### Motivation
Allowing another user (optionally password) on the machine with sudo access would make it accessible to more virtual machines.

### Description
I have a scenario where I want to add various static nodes, programatically. I do not have `root` user enabled on those machines however the `ubuntu` user has passwordless sudo access.

The username specification could come with a warning saying if we are unable to do `sudo su -` then it fails.

I propose the below script that does the same.

```
#!/bin/bash

if [ "$EUID" -eq 0 ]; then
  echo "User has root access"
  exit 0
fi

if [ -n "$ROOT_PASSWORD" ]; then
  echo "$ROOT_PASSWORD" | sudo -S true 2>/dev/null
  if [ $? -eq 0 ]; then
    echo "User has root access with the provided password"
    exit 0
  fi
fi

echo "User does not have root access"
exit 1
```

### Exit criteria

- [ ] Implement the feature
- [ ] Add feature to the Claudie

PR #1335, opened 2024-04-11
Feat/allow other usernames with root access

JKBGIT1:
closes https://github.com/berops/claudie/issues/1297

This PR allows other usernames with root access in the static nodes. `claudie-operator` checks if a user with the provided username can SSH into the node and whether the user is part of the `sudo` group. Because if the user isn't in the `sudo` group Claudie assumes the user doesn't have required root privileges. 

If the provided username can't be used for SSH or the user with this username isn't in the `sudo` group, `claudie-operator` doesn't start the workflow. Users can see the errors regarding this failure in the logs of the `claudie-operator` pod only.

Any comments and suggestions are most welcome.


services/claudie-operator/pkg/controller/controller.go
```diff
@@ -55,7 +58,61 @@ func (r *InputManifestReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+				snwd.Username = n.Username
+
+				// Parse private key.
+				pk, err := ssh.ParsePrivateKey(snwd.Secret.Data["privatekey"])
+				if err != nil {
+					log.Error(err, fmt.Sprintf("Failed to parse private key for %s.", snwd.Endpoint))
+					return ctrl.Result{RequeueAfter: REQUEUE_AFTER_ERROR}, nil
+				}
+
+				// Define the SSH client configuration.
```
cloudziu, line 74: I think that this is a bad idea to implement the validation here. If you do so, the connection will be established each time the reconciliation goes on, so it can go on every few seconds. 

I think it will be better to implement this in ansibler, as that service is responsible for the OS layer tasks

cloudziu (changes requested):
Nice work, thanks!
Left some comments

services/claudie-operator/pkg/controller/controller.go
JKBGIT1: I agree. I hadn't realized that it would establish the connection on every reconciliation. I just wanted to perform this verification as soon as possible to give users fast feedback. Is there also any other way to implement this? Besides the current one and the `ansibler` one you proposed.

cloudziu: I think that ansibler is the earliest you can get for static nodes. Terraformer doesn't interact with static nodes, so ansibler is the next one. Maybe @Despire or @bernardhalas could share an opinion on that.

JKBGIT1: Maybe one more thing to note. How will the user reconcile, if this check fails in the `ansibler`? Does changing the username attribute trigger the reconciliation?

cloudziu: Similary like any other attribute. The workflow would have to fail/finnish, and then the changes will be applied.

JKBGIT1: Okay, and to provide faster feedback to the user what do you think about performing this check in `builder`? Is it possible?

JKBGIT1: I would like to add it [here](https://github.com/berops/claudie/blob/20f48814a4e3378adaebedc63e32f667f67d1e92/services/builder/domain/usecases/workflow_helpers.go#L78), before sending request to the `terraformer`.

Despire: I think ansibler would be the goto place to insert this kind of stuff. Otherwise we will end up with some validation distributed everywhere.

When you think about it. The default is root. If someone wants to use another username then he must be sure it has root privileges. otherwise setup up the node will fail.

JKBGIT1: > The default is root. If someone wants to use another username then he/she must be sure it has root privileges.

In that case, should we even perform the checks?

Despire: I believe not since the playbooks will fail.

JKBGIT1: That's true, but it will take about ~30 minutes to fail, because of this [task](https://github.com/berops/claudie/blob/20f48814a4e3378adaebedc63e32f667f67d1e92/services/ansibler/server/ansible-playbooks/wireguard.yml#L8).

JKBGIT1: Eventually, I decided to remove the check from the `claudie-operator` and don't implement it at all.

cloudziu (approved):

Issue #1312, opened 2024-04-02
Bug: Claudie doesn't allow nodes from two different AWS provider instances in the same region


bernardhalas:
### Current Behaviour
Currently, Claudie doesn't allow the following configuration:

```
spec:
  providers:
    - name: aws-team1
      providerType: aws
...
    - name: aws-team2
      providerType: aws
...
  nodePools:
    dynamic:
      - name: control-aws
        providerSpec:
          name: aws-team1
          region: eu-west-1
      - name: compute-aws
        providerSpec:
          name: aws-team2
          region: eu-west-1
```

### Expected Behaviour
Sometimes, the users might want to deploy nodepools in the same region, by two different providers (e.g. for billing, reporting or security issues).

### Anything else to note
Currently, I see the options to set `provider.alias` to e.g.:
- `nodepool_{{ $provider.name }}_{{ $provider.region }}`
- `nodepool_{{ substr(sha256($provider.name + $provider.region), 0, 4) }}`
- `nodepool_{{ $random }}`

### References
- https://github.com/berops/claudie/issues/1296
- https://github.com/berops/claudie/blob/master/services/terraformer/templates/aws/provider.tpl

JKBGIT1:
Depends on https://github.com/berops/claudie/issues/1296

Issue #1296, opened 2024-03-26
Bug: Claudie doesn't support multiple Provider configurations for the same cloud provider


JKBGIT1:
### Current Behaviour

Using multiple [Provider](https://docs.claudie.io/latest/input-manifest/api-reference/#providers) (at least two) configurations for the same cloud provider results in the following error in `terraformer`

```
Error: Duplicate provider configuration
jakub-blog-test-z7ep12d │ 
jakub-blog-test-z7ep12d │   on jakub-blog-test-z7ep12d-azure-2-provider.tf line 1:
jakub-blog-test-z7ep12d │    1: provider "azurerm" {
jakub-blog-test-z7ep12d │ 
jakub-blog-test-z7ep12d │ A provider configuration for "azurerm" with alias "nodepool" was already
jakub-blog-test-z7ep12d │ given at jakub-blog-test-z7ep12d-azure-1-provider.tf:1,1-19. Each
jakub-blog-test-z7ep12d │ configuration for the same provider must have a distinct alias.
```

### Expected Behaviour

Claudie builds an infrastructure for the same cloud provider using multiple Provider configurations.

### Steps To Reproduce
1. Create at least two secrets for [Providers](https://docs.claudie.io/latest/input-manifest/api-reference/#providers) configuration. I created one secret with credentials for our sponsorship subscription and the other one with our pay-as-you-go subscription credentials.
2. Apply an InputManifest that provisions at least two node pools, each with a different Provider configuration. I've used the following one
```yaml
apiVersion: claudie.io/v1beta1
kind: InputManifest
metadata:
  name: jakub-blog-test
  labels:
    app.kubernetes.io/part-of: claudie
spec:
  providers:
    - name: cloudflare-1
      providerType: cloudflare
      secretRef:
        name: cloudflare-secret
        namespace: default
    - name: azure-1
      providerType: azure
      secretRef:
        name: berops-azure-secret
        namespace: default
    - name: azure-2
      providerType: azure
      secretRef:
        name: sponsor-azure-secret
        namespace: default

  nodePools:
    dynamic:
      - name: control-azure-1
        providerSpec:
          name: azure-1
          region: westeurope
          zone: "1"
        count: 1
        serverType: Standard_B2s
        image: Canonical:0001-com-ubuntu-minimal-jammy:minimal-22_04-lts:22.04.202212120

      - name: compute-azure-1
        providerSpec:
          name: azure-1
          region: westeurope
          zone: "2"
        count: 1
[...]

PR #1348, opened 2024-04-26
Fix/multiple providers

Despire:
Closes https://github.com/berops/claudie/issues/1296 
Closes https://github.com/berops/claudie/issues/1312

This PR also introduces a limit on names for:

clusters: 28 chars.
nodepools: 14 chars.

Note: Mostly the changes are in the terraform files where everything has a "provider.specName" suffix so that no collisions will be generated when using more than one of the same provider


cloudziu (approved):
LGTM 👍

internal/manifest/manifest.go
```diff
@@ -24,25 +24,25 @@ type Provider struct {
 }
 
 type HetznerDNS struct {
-	Name     string `validate:"required" yaml:"name"`
+	Name     string `validate:"required,max=15" yaml:"name"`
```
JKBGIT1, line 27: Shouldn't we write in the [API reference](https://docs.claudie.io/latest/input-manifest/api-reference/#providers) that the name of the cloud provider is limited by 15 chars?

JKBGIT1 (approved):
LGMT :+1:
Just one more question. Why do we use region in the `provider.alias` for most of the cloud providers and some (Hetzner) we don't?

Despire:
> Why do we use region in the provider.alias for most of the cloud providers and some (Hetzner) we don't?

This is the case for the hetzner as the networking stuff doesn't have to be region specific. Moreover, hetzner doesn't allow to create the same ssh keys (they check based on the key fingerprint) so it is not possible to have more than one hetzner provider in the same cluster (unless two completely different hetzner accounts will be used in which case it will work)

JKBGIT1:
```
alias = "nodepool_{{ $region }}_{{ $.Provider.SpecName }}"
```
And why isn't it possible to omit `{{ $region }}` in the `alias` and use only `{{ $.Provider.SpecName }}` in this line? It's for AWS, but that shouldn't matter if I got it right.

Despire:
> And why isn't it possible to omit `{{ $region }}` in the alias and use only `{{ $.Provider.SpecName }}` in this line? It's for AWS, but that shouldn't matter if I got it right.

For some providers you need the region (for example for aws and gcp, oci..., https://github.com/berops/claudie/blob/d3c7e8c11d7af9531768a2a870667577fe1c9f4d/services/terraformer/templates/aws/provider.tpl#L5 ) as the resources are region specific. So you actually need the provider for that specific region.

JKBGIT1:
Okay, got it. Thanks.

Issue #1344, opened 2024-04-23
Bug: Changing provider of a nodepool leads to corrupt cluster


Despire:
### Current Behaviour

```
    - name: aws-control-1
      providerSpec:
        name: aws-1
        region: eu-north-1
        zone: eu-north-1a
      count: 2
      serverType: t3.small
      #ubuntu
      image: ami-03df6dea56f8aa618
```

Changing the above provider `aws-1` to `aws-2` results in a corrupt cluster especially if the nodpool is used as a control plane.

### Expected Behaviour

This edge-case should be handled correctly in claudie.

### Steps To Reproduce
1. Use a nodepool as a control-plane in a cluster
2. change the provider of the nodepool
3. watch the workflow fail in kube-eleven

Despire:
add validation for guarding againts chaning the provider of a nodepool.

PR #1371, opened 2024-05-06
Guard against changing the cloud provider

JKBGIT1:
closes https://github.com/berops/claudie/issues/1344


Despire (approved):
Looks great 👍

PR #1379, opened 2024-05-09
prepare new version

Despire:


docs/CHANGELOG/changelog-0.8.x.md
```diff
@@ -0,0 +1,44 @@
+# Claudie `v0.8`
+
+!!! warning "Due to updating terraform files the `v0.8.x` will not be backwards compatible with `v0.7.x`"
```
JKBGIT1, line 3: IMO we should explicitly tell the users that upgrading the Claudie from v0.7.x to v0.8.0 will re-create their infrastructure. WDYT?

JKBGIT1 (approved):

PR #1399, opened 2024-05-21
context-box healthcheck

### LINKED ISSUE #1325: Chore: `context-box` restarts when it can't ping MongoDB
### Current Behaviour
If `context-box` can't ping MongoDB as part of its health check it restarts.

### Expected Behaviour

Do we want to handle this problem this way? Why should restarting the `context-box` fix the connection to another party?

### Steps To Reproduce
I experienced this when I deployed Claudie on a cluster without the required images. The `context-box` image got pulled before the Mongo one, therefore the pod with the `context-box` container was up before the pod with Mongo.
However, you can replicate it even when you prevent Mongo from scheduling.

Despire:
closes #1325 

Context-box will continue to restart if MongoDB is not deployed. As Context-box depends solely on mongoDB without which it cannot start.  The current state where contextbox restarts if mongodb is not deployed is okay I think, we can revisit this in the future, but currently without major changes to the context-box service I don't think it's a good idea.

However, I've added more time to check for MongoDB readiness before restarting the service and also added a healthcheck in case the MongoDB pod got terminated (which will not issue a restart of the context-box, the restart only happens when it is first deployed as a new pod)

### DIFF
--- a/manifests/claudie/context-box.yaml
+++ b/manifests/claudie/context-box.yaml
@@ -66,12 +66,6 @@ spec:
               service: context-box-readiness
             initialDelaySeconds: 5
             periodSeconds: 30
-          livenessProbe:
-            grpc:
-              port: 50055
-              service: context-box-liveness
-            initialDelaySeconds: 5
-            periodSeconds: 30
 ---
 kind: Service
 apiVersion: v1
--- a/manifests/claudie/kustomization.yaml
+++ b/manifests/claudie/kustomization.yaml
@@ -66,7 +66,7 @@ images:
 - name: ghcr.io/berops/claudie/claudie-operator
   newTag: 43f201d-2805
 - name: ghcr.io/berops/claudie/context-box
-  newTag: 43f201d-2805
+  newTag: ab6f6da-2807
 - name: ghcr.io/berops/claudie/kube-eleven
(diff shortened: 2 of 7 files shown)


JKBGIT1:
> Context-box will continue to restart if MongoDB is not deployed. As Context-box depends solely on mongoDB without which it cannot start.

What about making the `context-box` not ready and forbidding it to process any requests? If I am not mistaken something similar works in `builder`. `builder` doesn't process any request, unless all services (terraformer, ...) that it depends on are ready.

Despire:
> context-box not ready and forbidding it to process any requests?

afaik, this is the default in kubernetes when a service is not ready traffic will not get forwarded to it.

JKBGIT1:
In that case, what do you think about going with the not-ready approach?

Despire:
That is the idea of this PR

i.e. when the pod is deployed it will crashloop until the DB is created (I think this is ok)

Then a healthcheck has been added, and if the DB goes down or something while context-box is running the healthcheck will fail and traffic will not get dirrected to context-box until the DB is running again.

JKBGIT1:
You are right, I didn't read this in the PR description.
> which will not issue a restart of the context-box, the restart only happens when it is first deployed as a new pod

PR #1401, opened 2024-05-24
Chore/support ubuntu 24.04 in Azure and hetzner

JKBGIT1:
This PR adds support for Ubuntu 24.04 in Azure and Hetzner.

Support for Ubuntu 24.04 in GCP, OCI, AWS, and Genesis Cloud hasn't been implemented yet, because new changes in their TF files will trigger instances re-creation after the user upgrades the Claudie version. On top of that re-creation of AWS instances breaks an entire cluster, because Claudie won't be able to connect to these AWS instances using SSH to Install VPN.

For now, we decided to avoid another breaking change and started searching for a new way to add changes in TF files to avoid re-creations of the VM instances.

### DIFF
--- a/services/ansibler/server/ansible-playbooks/wireguard/tasks/configure.yml
+++ b/services/ansibler/server/ansible-playbooks/wireguard/tasks/configure.yml
@@ -48,6 +48,12 @@
 - name: Set MTU size for wg0
   ansible.builtin.shell: ifconfig {{ wg_interface_name }} mtu {{ mtu_size }}
 
+- name: Ensure file /etc/dhcp/dhclient.conf exists
+  ansible.builtin.copy:
+    content: ""
+    dest: /etc/dhcp/dhclient.conf
+    force: false
+
 - name: Make MTU change persistent across reboots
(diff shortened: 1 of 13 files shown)


services/ansibler/server/ansible-playbooks/wireguard/tasks/configure.yml
```diff
@@ -48,6 +48,12 @@
 - name: Set MTU size for wg0
   ansible.builtin.shell: ifconfig {{ wg_interface_name }} mtu {{ mtu_size }}
 
+- name: Ensure file /etc/dhcp/dhclient.conf exists
+  ansible.builtin.copy:
+    content: ""
+    dest: /etc/dhcp/dhclient.conf
+    force: false
+
```
Despire, line 56: Why do we need this ?

JKBGIT1: Yes, it is missing in Ubuntu 24.04 and the playbook failed due to that.

Despire (approved):

Issue #783, opened 2023-05-11
Feature: Workaround for pulling data from *.google.com on Hetzner nodes


bernardhalas:
### Motivation
Hetzner nodes sometimes get assigned IPs which are blacklisted on google (and maybe on some other providers too). This prevents Claudie from reliable provisioning of Hetzner nodes even if GCP nodepools are not used, because there are keys and packages downloaded during K8s provisioning stage by KubeEleven, from Google servers.

### Description
Try to figure out a workaround that would allow us to provision non-GCP clusters reliably on Hetzner even if all Hetzner boxes get assigned IPs from Google's blacklist.

### Exit criteria

- [ ] Provision K8s on Hetzner boxes with bad IPs

This is related to https://github.com/berops/claudie/issues/710. There's the same root-cause behind this, however:
- we can't easily fix the Hetzner-GCP clusters
- we should be able to find a workaround for Hetzner-non-GCP clusters with respect to the K8s provisioning stage; this the scope of this issue.

cloudziu:
I think that this PR Kubeone enhancement could be a good starting point to exploring further options.
https://github.com/kubermatic/kubeone/issues/88

cloudziu:
🎉🎉🎉🎉🎉🎉🎉

Issue #1272, opened 2024-03-13
Bug: 403 Forbidden when pulling pause containers


cloudziu:
This issue is related to: https://github.com/kubernetes/registry.k8s.io/issues/261 and https://github.com/berops/claudie/issues/783

In a cluster deployed on Hetzner, after a cluster is deployed one of the compute nodes stays in `NotReady` state. From the node logs we can see:
```
Failed to create pod sandbox: rpc error: code = Unknown desc = failed to get sandbox image "registry.k8s.io/pause:3.9": failed to pull image "registry.k8s.io/pause:3.9": failed to pull and unpack image "registry.k8s.io/pause:3.9": failed to resolve reference "registry.k8s.io/pause:3.9": unexpected status from HEAD request to https://europe-west3-docker.pkg.dev/v2/k8s-artifacts-prod/images/pause/manifests/3.9: 403 Forbidden
```
I've tried to manually pull the images from the VM with and the result is the same 403 Forbidden:
```
# ctr --debug image pull --http-dump --http-trace -k registry.k8s.io/pause:3.9
DEBU[0000] fetching                                      image="registry.k8s.io/pause:3.9"
DEBU[0000] resolving                                     host=registry.k8s.io
DEBU[0000] do request                                    host=registry.k8s.io request.header.accept="application/vnd.docker.distribution.manifest.v2+json, application/vnd.docker.distribution.manifest.list.v2+json, application/vnd.oci.image.manifest.v1+json, application/vnd.oci.image.index.v1+json, */*" request.header.user-agent=containerd/1.6.28 request.method=HEAD url="https://registry.k8s.io/v2/pause/manifests/3.9"
INFO[0000] HEAD /v2/pause/manifests/3.9 HTTP/1.1
INFO[0000] Host: registry.k8s.io
INFO[0000] Accept: application/vnd.docker.distribution.manifest.v2+json, application/vnd.docker.distribution.manifest.list.v2+json, application/vnd.oci.image.manifest.v1+json, application/vnd.oci.image.index.v1+json, */*
INFO[0000] User-Agent: containerd/1.6.28
INFO[0000]
DEBU[0000] DNS lookup                                    host=registry.k8s.io
DEBU[0000] DNS lookup complete                           coalesced=false result=34.96.108.209
DEBU[0000] Connection successful                         remote_addr="34.96.108.209:443" reused=false
INFO[0000] HTTP/1.1 307 Temporary Redirect
INFO[0000] Transfer-Encoding: chunked
[...]

PR #1440, opened 2024-07-10
Feat/use claudie http proxy when building cluster in hetzner

JKBGIT1:
closes https://github.com/berops/claudie/issues/783 https://github.com/berops/claudie/issues/1272

This PR adds support for HTTP proxy in `kube-eleven`. The default workflow is as follows. Claudie utilizes an HTTP proxy (http://proxy.claudie.io:48213) only when a user wants to build a K8s cluster that contains at least 1 node from the Hetzner Cloud provider. If the k8s cluster doesn't have any nodes from the Hetzner Cloud provider Claudie won't use an HTTP proxy in `kube-eleven` to build the desired k8s cluster.

The usage of an HTTP proxy is customizable. A user can turn off the HTTP proxy usage by setting `HTTP_PROXY_MODE` to `off`. This means Claudie won't use the HTTP proxy at all (even when there are nodes in the K8s cluster from the Hetzner cloud provider). On the other hand, the user can turn on the HTTP proxy usage whenever building a K8s cluster (even when there aren't any nodes in the K8s cluster from the Hetzner cloud provider) by setting `HTTP_PROXY_MODE` to `on`.

Besides that users can use their own HTTP proxy by setting `HTTP_PROXY_URL` to point to their proxy URL.


bernardhalas:
Probably we should mention the use of value `HTTP_PROXY_MODE=default` in [docs/http-proxy/http-proxy.md](docs/http-proxy/http-proxy.md).

Despire (approved):
LGTM 👍

services/kube-eleven/server/domain/utils/kube-eleven/kube_eleven.go
```diff
@@ -153,16 +155,67 @@ func (k *KubeEleven) generateTemplateData() templateData {
 
 	var potentialEndpointNode *spec.Node
 	data.Nodepools, potentialEndpointNode = k.getClusterNodes()
-
 	data.APIEndpoint = k.findAPIEndpoint(potentialEndpointNode)
 
+	hasHetznerNodes := k.hasHetznerNodes(data.Nodepools)
+	httpProxyMode := utils.GetEnvDefault("HTTP_PROXY_MODE", defaulHttpProxyMode)
+
+	data.NoProxy = ""
```
Despire, line 163: You don't need to initialize with a default value, all values in Go are zero initialized, this line can be omitted.

services/kube-eleven/server/domain/utils/kube-eleven/kube_eleven.go
```diff
@@ -153,16 +155,67 @@ func (k *KubeEleven) generateTemplateData() templateData {
...
+				}
+			}
+		}
+		// data.NoProxy has to terminate with the comma
+		// if "svc" isn't in NoProxy the admission webhooks will fail, because they will be routed to proxy
+		data.NoProxy = fmt.Sprintf("%s,%s,", data.NoProxy, "svc")
+
+		data.HttpProxyUrl = utils.GetEnvDefault("HTTP_PROXY_URL", defaulHttpProxyUrl)
+	} else {
+		data.UtilizeHttpProxy = false
```
Despire, line 197: Ditto, by default data.UtilizeHttpProxy is false (its the zero value)

services/kube-eleven/server/domain/utils/kube-eleven/kube_eleven.go
```diff
@@ -153,16 +155,67 @@ func (k *KubeEleven) generateTemplateData() templateData {
...
+		// or it isn't turned off and there is at least 1 Hetzner node in the k8s cluster
+		data.UtilizeHttpProxy = true
+		// add nodes privat and public IPs to the NoProxy. Otherwise the kubeone proxy won't work properly
+		for _, nodePool := range data.Nodepools {
+			for _, node := range nodePool.Nodes {
+				if data.NoProxy == "" {
+					data.NoProxy = fmt.Sprintf("%s,%s", node.Node.Private, node.Node.Public)
+				} else {
+					data.NoProxy = fmt.Sprintf("%s,%s,%s", data.NoProxy, node.Node.Private, node.Node.Public)
+				}
```
Despire, line 175: What you can do here is to create a slice of strings and append to that (also in the case of Lbs) and later join them with a comman, i.e.

```
var noProxy []string

for _, nodePool := range data.Nodepools {
    for _, node := range nodePool.Nodes {
       noProxy = append(noProxy, node.Node.Private, node.Node.Public)
    }
}

....

data.NoProxy = fmt.Sprintf("%s,svc", strings.Join(noProxy, ","))
```

services/ansibler/server/domain/usecases/update_no_proxy_envs.go
```diff
@@ -0,0 +1,179 @@
...
+	// HTTP Proxy mode isn't "off" and cluster has a Hetzner node or cluster is being build using HTTP proxy
+	if httpProxyMode == "off" || (httpProxyMode == "default" && !hasHetznerNodeFlag) {
+		return &pb.UpdateNoProxyEnvsResponse{Current: request.Current, Desired: request.Desired}, nil
+	}
+
+	nodesChangedFlag := nodesChanged(request.Current.ClusterInfo, request.Desired.ClusterInfo)
+	// At least one node in the cluster has to be changed to perform the update of NO_PROXY and no_proxy env variables.
+	if !nodesChangedFlag {
+		return &pb.UpdateNoProxyEnvsResponse{Current: request.Current, Desired: request.Desired}, nil
+	}
```
Despire, line 53: I would remove this check, and just always update the noproxy envs, The workflow for building a cluster is run when you either add or delete nodes to the k8s cluster. (or also Lbs but in this case the redundancy wouldn't matter)

services/ansibler/server/domain/usecases/update_no_proxy_envs.go
```diff
@@ -0,0 +1,179 @@
...
+
+	if err := commonUtils.CreateKeysForStaticNodepools(commonUtils.GetCommonStaticNodePools(currentK8sClusterInfo.NodePools), clusterDirectory); err != nil {
+		return fmt.Errorf("failed to create key file(s) for static nodes : %w", err)
+	}
+
+	noProxyList := createNoProxyList(desiredK8sClusterInfo.GetNodePools(), desiredLbs)
+	if err := utils.GenerateInventoryFile(templates.NoProxyEnvsInventoryTemplate, clusterDirectory, noProxyInventoryFileParameters{
+		K8sNodepools: NodePools{
+			Dynamic: commonUtils.GetCommonDynamicControlPlaneNodes(currentK8sClusterInfo.NodePools, desiredK8sClusterInfo.NodePools),
+			Static:  commonUtils.GetCommonStaticControlPlaneNodes(currentK8sClusterInfo.NodePools, desiredK8sClusterInfo.NodePools),
```
Despire, line 135: Why do we need to extract common nodepools to both current and desired state ?
Why can't we use the nodepools from desired state ?

Despire (commented):
A nice PR 👍 , left some comments

services/ansibler/server/domain/usecases/update_no_proxy_envs.go
JKBGIT1: We have to check if the control-plane nodes are in both desired and current state because if they aren't in the current state, they aren't in the k8s cluster yet, thus you can't run `kubeadm init` cmds to update no proxy envs on them. On the other hand, if they aren't in the desired state, they were already destroyed by the `terraformer`.

Issue #1441, opened 2024-07-12
Feature: Terraform Templates Overrride


Despire:
### Problem:
How TF works - if you do a breaking change in TF files, TF replaces the resources. In our case, if the nodes are replaced (e.g. due to an Ubuntu image update), the nodes are destroyed by Terraform and then they're re-created. However, after re-creation, they're not part of the K8s cluster. The old nodes have not been gracefully removed from the cluster either, therefore the control plane sees them as `NotReady` and the users can't recover from this manually.
Now, we need to be able to change Terraform code for nodepools. This is needed in case of upgrading the OS version in the TF files or in case of adding new features that come with breaking changes in TF HCL code.
### Proposal:
Ideally, we should find a way to deliver breaking changes in TF via rolling updates. That is, we create a new nodepool using the new TF code and then remove the old nodepool built using the old TF code. This way the cluster can sustain such updates without downtime. One way how to achieve this would be the following.
We'll separate the TF code into a new repository. This would be a GitHub/berops repository by default, but we allow users to override it. We also enable users to specify the repository tag and commit hash.
Unless the tag has been overridden by the user, the repository tag will match the running Claudie release version (e.g. 0.8.1).
If the tag is specified, Claudie will use the TF code from the particular commit with the tag.
If the tag has not been specified and Claudie defaults to the tag matching the current Claudie release version, upon an upgrade, Claudie will automatically do a rolling update by deploying the nodepool with the same configuration and TF code from the new release (e.g. 0.8.2).
However, if the tag was specified by the user to 0.8.1, Claudie upgrade to 0.8.2 would do no change to the nodepool infra, as the user pin of the TF code has a higher priority
### Effects:
- Unless the user pins a specific configuration repository tag, Claudie will perform rolling updates for TF code updates.
- If the TF code didn't change between Claudie versions, it's perfectly OK to have the same commit having multiple tags (e.g. 0.8.1, 0.8.2, 0.9.0); in that case, Claudie will be smart and won't trigger a rolling update.
- This allows users to override TF configuration by forking our upstream configuration repository and introducing their changes.
### Open questions:
- Design API for introducing git repo with TF templates (CRD/ConfigMap/InputManifest).
- Is the TF code repository going to be used also for the future repo of the `kuber` manifest overrides?
- Probably a suitable way to introduce this would be by rewriting TF templates to TF modules, because we can nicely control the TF module URL and TF module version from `main.tf`/`main.tpl`. To be discussed.
[...]

PR #1460, opened 2024-08-05
Feat: Support for External Terraform Templates

Despire:
**NOTE: The large number of diff (lines added/deleted) are from restructuring the Protobuf files.**

First part of https://github.com/berops/claudie/issues/1441

This PR introduces work with external terraform files which can be downloaded from any publicly
available git repository.

The syntax for specifing the template repository was added at the provider level.

```
  gcp:
    - name: gcp-1
      templates:
        repository: "https://github.com/berops/claudie-config"
        tag: v0.1.2
        path: "templates/terraformer/gcp"
       ...
```

- if no templates are specified it will always default to the latest commit on the Master/Main branch of the respective cloudprovider on the berops repository.

- if templates are specified, but no tag is present it will default to the latest commit of the Master/Main branch of the respective repository.

The template repository need to follow a certain convention to work properly.
For example:
If we consider an external template repository accessible via a public git repository at:

	https://github.com/berops/claudie-config

The repository may either only contain the necessary templates files or they can be stored at a
subtree. To handle this a path is required to be passed along the public git repository such as:

	templates/terraformer/gcp

This denotes that the necessary templates for Google Cloud Platform can be found in the subtree at:

	claudie-config/templates/terraformer/gcp

To only deal with the necessary template files a sparse-checkout is used when downloading the external
repository to have a local mirror present which will then be used to generate the terraform files.
When using the template files for generation the subtree present at "claudie-config/templates/terraformer/gcp"
the directory is traversed and the following rules apply:

  - if a subdirectory with name "provider" is present, all files within this directory will be considered as related to
    Providers for interacting with the API of respective Cloud Providers, SaaS providers etc. When using the templates
    for generation, the struct [templates.Provider](https://github.com/berops/claudie/blob/5dc0e7c8f5503a6f2c202a982f5c4aa11bed0346/services/terraformer/server/domain/utils/templates/structures.go#L54) will be passed for each file individually.

  - if a subdirectory with name "networking" is present all files within this directory will be considered as related
    spawning a common networking infrastructure for all nodepools from a single provider. The files in this subdirectory
    will use the providers generated in the previous step. When using the templates the struct [templates.Networking](https://github.com/berops/claudie/blob/5dc0e7c8f5503a6f2c202a982f5c4aa11bed0346/services/terraformer/server/domain/utils/templates/structures.go#L92)
    will be passed for each file individually.

  - if a subdirectory with name "nodepool" is present all files within this directory will be considered as related
    to spawning the VM instances along with attached disk and related resources for a single node coming from a specific
    nodepool. When using the templates the struct [templates.Nodepools](https://github.com/berops/claudie/blob/5dc0e7c8f5503a6f2c202a982f5c4aa11bed0346/services/terraformer/server/domain/utils/templates/structures.go#L138) will be passed for each file individually.

  - if a subdirectory with name "dns" is present, all files within this directory will be considered as related to DNS.
    Thus, the [templates.DNS](https://github.com/berops/claudie/blob/5dc0e7c8f5503a6f2c202a982f5c4aa11bed0346/services/terraformer/server/domain/utils/templates/structures.go#L151) struct will be passed for each file when generating the templates.
    Note: This subdirectory should contain its own file that will generate the Provider needed for interacting with
    the necessary API of the respective cloud providers (the ones that will be generated from the "provider" subdirectory
    will not be used in this case).

The complete structure of a subtree for a single provider for external templates located at claudie-config/templates/terraformer/gcp
can look as follows:

	└── terraformer
	    |── gcp
	    │	├── dns
	    │   	└── dns.tpl
	    │	├── networking
	    │		└── networking.tpl
	    │	├── nodepool
	    │		├── node.tpl
	    │		└── node_networking.tpl
	    │	└── provider
	    │		└── provider.tpl
		...

Examples of external templates can be found on:  https://github.com/berops/claudie-config


docs/input-manifest/external-templates.md
```diff
@@ -0,0 +1,82 @@
...
+        path: "templates/terraformer/genesiscloud"
+      secretRef:
+        name: genesiscloud-secret
+        namespace: secrets
+...
+```
+
+- if no templates are specified it will always default to the latest commit on the Master/Main branch of the respective cloudprovider on the berops repository (i.e. ` https://github.com/Despire/claudie-config`).
+
+- if templates are specified, but no tag is present it will default to the latest commit of the Master/Main branch of the respective repository.
```
bernardhalas, line 26: We should also mention, what's the default `path` in case no definition is provided in the `templates` stanza.

Despire: Do you mean this? Or literally write the default "template" structure that will be inserted

https://github.com/berops/claudie/blob/cb419b8e39ab50c769167c8f6a9aa7fbc92cbc4b/docs/input-manifest/external-templates.md?plain=1#L24

bernardhalas: Apologies for not being clear enough, allow me to rephrase:
We should also mention, what's the default `path` in case no definition is provided in the `templates.path` stanza. Or is `templates.path` mandatory for `templates`?

Despire: it is mandatory.

`templates.repository` and `templates.path` are mandatory, i.e this identify where templates for a provider can be found.

`templates.tag` is optional.

bernardhalas: Understood. All good then, thanks for the clarification.

bernardhalas (approved):
LGTM, thank you!

PR #1510, opened 2024-09-18
use az sponsorship subscription in e2e

JKBGIT1:
This PR uses AZ sponsorship subscription in e2e tests for nodepools and Pay-As-You-Go for hostnames.


Despire (approved):
LGTM 👍

Issue #1480, opened 2024-08-21
Chore: Upgrade Longhorn to 1.7.0


JKBGIT1:
### Description

Upgrade Longhorn to 1.7.0 to support the possibility of controlling the share-manager scheduling.

### Exit criteria

- [x] Upgrade Longhorn according to https://longhorn.io/docs/1.7.0/deploy/upgrade/
- [x] Check if upgrading Claudie to the version with the latest Longhorn doesn't break an existing cluster

PR #1511, opened 2024-09-19
upgrade Longhron from 1.6.0 to 1.7.0

JKBGIT1:
closes https://github.com/berops/claudie/issues/1480

The upgrade was tested following these steps:
1. Create a cluster with Longhorn v1.6.0
2. Deploy workload with PVC in the created cluster (I deployed a kube-prometheus stack)
3. Upgrade Longhorn to v1.7.0 and deploy new containers on Claudie cluster
4. Add a new node to the InputManifest and apply the new configuration
5. Make sure a new node was added to the k8s cluster, the Longhorn was upgraded and there wasn't data lost.

In any case, I recommend doing backups before upgrading to a Claudie version with Longhorn v1.7.0.

### DIFF
--- a/manifests/testing-framework/test-sets/test-set1/1.yaml
+++ b/manifests/testing-framework/test-sets/test-set1/1.yaml
@@ -44,21 +44,21 @@ spec:
       secretRef:
         name: aws-secret
         namespace: e2e-secrets
-    - name: azure-1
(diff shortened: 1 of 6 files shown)


Despire (approved):
LGTM 👍

Issue #1514, opened 2024-09-23
Bug: GCP VMs can't reach their metadata servers when using proxy


JKBGIT1:
### Current Behaviour

Claudie can't utilize a proxy when using the GCP VMs because the proxy server can't reach the GCP metadata servers.

The following logs are from the proxy server.

```
1727094082.406      0 104.199.110.247 TCP_MISS/502 4066 GET http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/ - HIER_NONE/- text/html
1727094108.056      1 104.199.110.247 TCP_MISS/502 4066 GET http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/ - HIER_NONE/- text/html
1727094159.310      4 104.199.110.247 TCP_MISS/502 4066 GET http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/ - HIER_NONE/- text/html
```

### Expected Behaviour

Add GCP metadata servers to no proxy envs. This way the GCP VMs will succeed in reaching their metadata servers because the requests won't go through the proxy server.

### Steps To Reproduce
1. Turn on the `HTTP_PROXY_MODE` by setting its value to `on`
2. Create a manifest with GCP VM
3. Apply the manifest and wait for the failure in `kube-eleven`

PR #1517, opened 2024-09-24
add GCP metadata server IP and hostnames to no proxy envs

JKBGIT1:
closes https://github.com/berops/claudie/issues/1514

This supports building K8s clusters with GCP VMs using an HTTP proxy.

When using an HTTP Proxy this `metadata,metadata.google.internal,169.254.169.254,metadata.google.internal.` has to be in the no proxy envs. Otherwise Claudie fails to build the k8s cluster.

This PR also contains an update of `.env` file, cause it should run the e2e tests with proxy mode turned on and using an HTTP proxy that supports access to a limited set of domains. Once the e2e tests finish successfully the changes in the `.env` file will be reverted.

### DIFF
--- a/services/ansibler/server/domain/usecases/update_no_proxy_envs.go
+++ b/services/ansibler/server/domain/usecases/update_no_proxy_envs.go
@@ -134,7 +134,8 @@ func createNoProxyList(desiredNodePools []*spec.NodePool, desiredLbs []*spec.LBc
 	}
 
 	// if "svc" isn't in noProxyList the admission webhooks will fail, because they will be routed to proxy
-	noProxyList = fmt.Sprintf("%s,%s,", noProxyList, "svc")
+	// "metadata,metadata.google.internal,169.254.169.254,metadata.google.internal." are required for GCP VMs
+	noProxyList = fmt.Sprintf("%s,svc,metadata,metadata.google.internal,169.254.169.254,metadata.google.internal.,", noProxyList)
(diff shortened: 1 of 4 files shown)


services/ansibler/server/domain/usecases/update_no_proxy_envs.go
```diff
@@ -134,7 +134,8 @@ func createNoProxyList(desiredNodePools []*spec.NodePool, desiredLbs []*spec.LBc
 	}
 
 	// if "svc" isn't in noProxyList the admission webhooks will fail, because they will be routed to proxy
-	noProxyList = fmt.Sprintf("%s,%s,", noProxyList, "svc")
+	// "metadata,metadata.google.internal,169.254.169.254,metadata.google.internal." are required for GCP VMs
```
Despire, line 137: Is the IP `169.254.169.254` needed there or does it not work without the IP?

JKBGIT1: It is needed. At least according to https://cloud.google.com/compute/docs/troubleshooting/troubleshoot-metadata-server#failed-proxy

"...If the VM is behind a proxy, you must set the NO_PROXY configuration for both the IP address and Hostname."

But TBH I haven't tried to build the k8s cluster without the IP.

Despire (approved):
LGTM 👍

PR #1531, opened 2024-10-08
fix: controller status logging

Despire:
K8s Controller was not correctly reading the status of the manifest when multiple stages are being processed, i.e. rolling update.


JKBGIT1 (approved):
LGTM :+1:

PR #1535, opened 2024-10-10
chore: remove dangling directories left after nodepool errors

Despire:
On error, the directories where templates are generate are left as be. This poses an issue with rolling update when an invalid repository is used where the invalid templates will be left in that directory and any subsequent retries will fail until the invalid templates are removed manually.

This PR simply deleted the directory both on successfully generating the template and also when errored.


JKBGIT1 (approved):
LGTM :+1:

Issue #1526, opened 2024-10-04
Chore: Allow turning proxy off/on


Despire:
### Description

In https://github.com/berops/claudie/pull/1440 the proxy was introduced to claudie. At the moment we currently don't have a way to turn the proxy off once it is turned on. the `HTTP_PROXY` once set are not unset when the setting is turned of

### Exit criteria

- [ ] I'm able to turn the Proxy on/off

PR #1540, opened 2024-10-14
Allow turning proxy on and off

JKBGIT1:
closes https://github.com/berops/claudie/issues/1526


services/builder/domain/usecases/workflow_helpers.go
```diff
@@ -76,6 +76,76 @@ func (u *Usecases) buildCluster(ctx *builder.Context) (*builder.Context, error)
...
+		} else if desiredProxySettings.Mode == "on" {
+			updateProxyEnvsFlag = true
+		}
+	}
+
+	// HttProxyUrl and NoProxyList will be set before first task in ansibler and then updated after ansibler Install VPN phase.
+	ctx.ProxyEnvs = &spec.ProxyEnvs{
+		UpdateProxyEnvsFlag: updateProxyEnvsFlag,
+	}
+
```
Despire, line 148: move this into `internal/builder_context` file into its own function

```go
func (c *BuilderContext) DetermineProxyUpdate() error {
....
}
```


Further, rewrite the if else if chain into a switch that returns an error in the default case. By returning an error the build will not progress if we cannot determine if an proxy update is needed, avoiding failure.

```go
hetznerInDesired := builder.HasHetznerNode(ctx.DesiredCluster.ClusterInfo)
hetznerInCurrent := builder.HasHetznerNode(ctx.CurrentCluster.ClusterInfo)
switch {
case currProxySettings.Mode == offMode && desiredProxySettings.Mode == offMode:
...
case currProxySettings.Mode == defaultMode && desiredProxySettings.Mode == defaultMode && !hetznerInCurrent && !hetznerInDesired:
...
case currProxySettings.Mode == defaultMode && !hetznerInCurrent && desiredProxySettings.Mode == offMode:
...
case currProxySettings.Mode == offMode && desiredProxySettings.Mode == defaultMode && !hetznerInDesired:
...
default:
   return errors.New("failed to determine if proxy update is needed")
}
```

services/manager/internal/service/create_desired_state.go
```diff
@@ -83,13 +83,21 @@ func createK8sClustersFromManifest(from *manifest.Manifest, into *store.Config)
 	//    catching newly created or existing (updated).
 	desiredClusters := make(map[string]struct{})
 	for _, cluster := range from.Kubernetes.Clusters {
+		useInstallationProxy := &spec.InstallationProxy{}
+
+		if cluster.InstallationProxy != nil {
+			useInstallationProxy.Mode = cluster.InstallationProxy.Mode
+			useInstallationProxy.Endpoint = cluster.InstallationProxy.Endpoint
+		}
+
```
Despire, line 92: If the proxy is not set this will result in an empty `*spec.InstallationProxy` however in builder you are by default setting the `default` setting and after the workflow in builder finishes the current state and desired state will diverge as desired state as created in this function will have
```go
&spec.InstallationProxy{
Mode: ""
Endpoing : ""
}
```

and the current state after the workflow finishes

```go
&spec.InstallationProxy{
Mode: "default"
Endpoing : ""
}
```

Despire (changes requested):
Generally the PR looks great, there are some changes I've described.

manifests/testing-framework/testing-framework.yaml
```diff
@@ -84,6 +84,12 @@ spec:
         - name: rolling-update-2
           secret:
             secretName: rolling-update-2
+        - name: proxy-with-hetzner
+          secret:
+            secretName: proxy-with-hetzner
+        - name: proxy-without-hetzner
+          secret:
+            secretName: proxy-without-hetzner
```
Despire, line 92: You also need to reference this in the `volumeMounts` in https://github.com/berops/claudie/blob/741945a1eb77c8355785e5715339e9b666c835bb/manifests/testing-framework/testing-framework.yaml

Despire (changes requested):

Despire (approved):
LGTM

Issue #1522, opened 2024-09-30
Bug: Autoscaler missing support for GenesisCloud


Despire:
### Current Behaviour

Autoscaler-adapter service is missing implementation for GenesisCloud provider

https://github.com/berops/claudie/blob/5ea5c4ca6d56e9d1c15121c0a2f2ddbad5c62e41/services/autoscaler-adapter/node_manager/node_manager.go#L17-L27

### Expected Behaviour

Add support for GenesisCloud for autoscaling.

PR #1543, opened 2024-10-15
Fix/genesis autoscaling

Despire:
Closes  https://github.com/berops/claudie/issues/1522

### DIFF
--- a/services/autoscaler-adapter/node_manager/utils.go
+++ b/services/autoscaler-adapter/node_manager/utils.go
@@ -1,6 +1,8 @@
 package node_manager
 
 import (
+	"regexp"
+	"strconv"
 	"strings"
 
 	"cloud.google.com/go/compute/apiv1/computepb"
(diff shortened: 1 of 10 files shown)


services/autoscaler-adapter/node_manager/utils.go
```diff
@@ -10,6 +12,163 @@ import (
...
+		"vcpu-12_memory-36g_disk-80g_nvidia1080ti-3",
+		"vcpu-16_memory-48g_disk-80g_nvidia1080ti-4",
+		"vcpu-20_memory-60g_disk-80g_nvidia1080ti-5",
+		"vcpu-24_memory-72g_disk-80g_nvidia1080ti-6",
+		"vcpu-28_memory-84g_disk-80g_nvidia1080ti-7",
+		"vcpu-32_memory-96g_disk-80g_nvidia1080ti-8",
+	}
+
+	m := make(map[string]*typeInfo, len(instanceTypes))
+	for _, typ := range instanceTypes {
```
JKBGIT1, line 151:
```suggestion
	for _, type := range instanceTypes {
```

for some reason `typ` looks strange to me

services/autoscaler-adapter/node_manager/utils.go
```diff
@@ -10,6 +12,163 @@ import (
...
+		serverName := typ
```
JKBGIT1, line 152:
```suggestion
		serverName := type
```

has to be applied when the previous does

JKBGIT1 (approved):
Left some small suggestions. Otherwise LGTM :+1:

services/autoscaler-adapter/node_manager/utils.go
Despire: type is a Go keyword.

Issue #1538, opened 2024-10-11
Feature: Make Certificate valid for all control nodes.


Despire:
### Description

When a k8s cluster is build without a loadbalancer the generate kubeconfig is valid for only on selected node. If we use multiple nodes for the control plane the certificate will not be valid for all of them.

### Exit criteria

- [ ] Explore ways to make the generate certificate valid for all control nodes when a loadbalancer is not used

PR #1546, opened 2024-10-17
feat: add support for alternative names

Despire:
Closes https://github.com/berops/claudie/issues/1538

This PR introduces alternative names that are used for the generated kubeconfig.

What this means, is that in addition to the selected api endpoint (Whether it is a loadbalancer or a control plane node), **All of the control plane nodes will be a valid endpoint for the kubeconfig to be used on.**


bernardhalas (approved):
Thank you @Despire , LGTM.

Issue #1545, opened 2024-10-16
Feature: labeling static nodes


eminden:
### Motivation
Current implementation patches the node labels with predefined constants embedded into go code. It would be nice to make it configurable through the crd.

### Description
Following  labels are overwritten for static nodes, it would be nice to dynamically change these from InputManifest crd as it is done for dynamic node pools. It makes sence in case I am having static nodes in different regions and zones.

```
Labels:             claudie.io/provider=static-provider
                    claudie.io/provider-instance=static-provider
                    topology.kubernetes.io/region=static-region
                    topology.kubernetes.io/zone=static-zone
```

PR #1550, opened 2024-10-21
feat: allow to overwrite default labels for static nodepools

Despire:
This PR allows to overwrite default labels for static nodes. 
Further, if only the labels, annotations, taints were changed the config was not rescheduled, this PR also introduces a changes that fixes this.

Closes https://github.com/berops/claudie/issues/1545


JKBGIT1 (approved):
LGTM :+1:

Issue #1567, opened 2024-11-06
Feature: Rollback on autoscaler failing to add new node


Despire:
### Description

Autoscaler adds new nodes into a nodepool based on the resource usage within the spawned cluster. Currently there is no retry/error handling on the case where the addition of a node fails (could be due to varying reasons such as quota limits, issues with various registries etc.).

On error the autoscaling event should rollback to the old state, i.e. remove the infrastructure (if spawned).

### Exit criteria

- [ ] Rollback on autoscaling error

PR #1582, opened 2024-11-19
Feat/autoscaler reconciliation

Despire:
Closes https://github.com/berops/claudie/issues/1567


services/manager/internal/service/watchers.go
```diff
@@ -134,7 +145,12 @@ func (g *GRPC) WatchForPendingDocuments(ctx context.Context) error {
 			continue
 		}
 
-		logger.Info().Msgf("Config has been successfully processed and moved to the %q state", manifest.Scheduled.String())
+		switch result {
+		case NotReady:
```
JKBGIT1, line 149: Do we need to catch the `NotReady` case when it does nothing?

Despire: No we don't, but the linter will catch it.

Issue #1579, opened 2024-11-15
Chore: Restructure E2E pipeline


Despire:
### Description

The structure of the E2E pipeline has become a bottleneck in the experience of a fast feedback loop when changes are made to the codebase. For each change, all existing test-sets are executed, which is not good, as for example a change in the proxy would trigger all remaining test-sets, or a change in the rolling-up update would also trigger all test-sets to be executed, and so on.

The suggestion is to restructure the E2E pipeline so that only the test-sets relevant to the changes are run, resulting in faster feedback for fixing potential bugs.

We should also run the entire E2E pipeline every two weeks or before a release, if possible, to check if any of the changes have affected other test sets.

### Exit criteria

- [ ] Restructure E2E test-set so that only the test-set relevant for the code changes are executed, This could be selected manually or automatically based on the files changed

PR #1583, opened 2024-11-21
Restructure E2E tests

Despire:
Closes https://github.com/berops/claudie/issues/1579

This PR introduced 5 labels which when set on a pull request will trigger specific test-sets so that we do not execute the others that are unrelevant for the changes made to the codebase.


JKBGIT1 (approved):
:+1:

Issue #1365, opened 2024-05-02
Chore: Ubuntu 22.04 is not supported


smulikHakipod:
Hello!
Tried to install using Ubuntu 24.04 VM and got:

JKBGIT1:
The next release will present support for Ubuntu 24.04 in Hetzner and Azure (https://github.com/berops/claudie/pull/1401).

Adding support for Ubuntu 24.04 in the rest of the Cloud providers will cause VM re-creation after the user upgrades Claudie (it requires updating their TF files). Therefore we decided to avoid presenting these changes for now and we will work on a better solution to present new changes in the TF files.

PR #1599, opened 2024-12-06
update kubeone version & adjust ansible-playbook

Despire:
adds `backrefs` to the update proxy envs ansible playbook. Without this the environment variables would be appended to the end of the file each time and after 5 runes the entire /etc/environment file would be cluttered with these variables. With `backrefs` they will be overwritten every time.

Closes https://github.com/berops/claudie/issues/1598
Closes https://github.com/berops/claudie/issues/1365

Based on the Kubeone 1.9 release docs, there does seem not to be any backwards incompatibility issues


JKBGIT1 (approved):
LGTM

PR #1603, opened 2024-12-13
Chore/node local dns

### LINKED ISSUE #1425: Bug: Node-local-dns broken
### Current Behaviour
Deploying dnsutils and running `dig www.google.com` takes too long (20 ms), which indicates that node-local-dns is broken. After deploying node-local-dns manually, I get resolution time down to 1-2 ms.

### Expected Behaviour
node-local-dns is not used, whereas it should be.

### Steps To Reproduce
1. deploy dnsutils
2. execute `dig www.google.com`
3. observe resolution speed

By following the [setup guide](https://kubernetes.io/docs/tasks/administer-cluster/nodelocaldns/#configuration), I was able to fix the deployment.

Despire:
Closes #1425


JKBGIT1 (approved):
LGTM :+1: 
So, we eventually decided to deploy node-local-dns using kubeone, right?

Despire:
> So, we eventually decided to deploy node-local-dns using kubeone, right?

Yes

Issue #1484, opened 2024-08-30
Bug: Don't Allow control plane nodepool to have a count of 0.


Despire:
### Current Behaviour
if a control plane nodepool has a 0 node count, certain invariants break within the code, but does not make even sense to have a control nodepool with 0 node count.

### Expected Behaviour

Validation should be added to check if atleast 1 node will be among all defined control nodepools

### Steps To Reproduce

1. Add validation for control nodepool count to be > 0

PR #1605, opened 2024-12-13
chore: improve validation error msg

Despire:
Closes https://github.com/berops/claudie/issues/1484

Validation does not allow a nodepool count to be 0, thus the above issue description does no longer apply, however this PR introduces a more readable error to be printed.


JKBGIT1 (approved):
LGTM :+1:

Issue #1604, opened 2024-12-13
Bug: Changing kubernetes version does not trigger build process


Despire:
### Current Behaviour

Changing the kubernetes version to a higher version does not trigger the build process, but moves the manifest to `Done` state nevertheless.

```yaml
  kubernetes:
    clusters:
      - name: test
        version: 1.29.0 -> 1.30.0
```

### Expected Behaviour

Changing just the kubernetes version should trigger the build process

PR #1607, opened 2024-12-16
fix: add upgrade for k8s version

Despire:
adds task for upgrading k8s version

Closes https://github.com/berops/claudie/issues/1604


JKBGIT1 (approved):
LGTM :+1:

Issue #1597, opened 2024-12-05
Bug: Certificate with alternative names is invalid


Despire:
### Current Behaviour

Spawn a cluster with 3 control plane nodes from the same nodepool withtout a loadbalancer, then remove 2 of them. All of the control plane nodes are used as alternative names if a loadbalancer is not present so that the generated kubeconfig can be used with any of the nodes.

The issue is that after we delete a node, some of the information of these nodes seems to be still present which causes subsequent rejoin of the control plane nodes to fail the build process in the kube-eleven stage.

```
validating certificate period for ca certificate
I1205 10:05:23.295117   10584 certlist.go:156] [certs] Using the existing CA certificate \"/etc/kubernetes/pki/ca.crt\" and key \"/etc/kubernetes/pki/ca.key\"
I1205 10:05:23.769037   10584 certs.go:519] validating certificate period for apiserver certificate\nx509: certificate is valid for 10.96.0.1, 192.168.2.3, 130.61.73.68, 130.61.215.94, 130.61.170.70, not 130.61.28.8
certificate apiserver is invalid
``` 

### Expected Behaviour

If I delete control plane nodes from the same nodepool and then add them back to the cluster the workflow should continue as expected.

### Steps To Reproduce
1. Create a Cluster with 3 control plane nodes from the same nodepool without a loadbalancer
2. Delete 2 of the control plane nodes (i.e by decreasing the nodepool node count)
3. Add 2 control plane nodes from the same nodepool back (i.e. by increasing the nodepool count)
4. See failure in kube-eleven with the above mentioned error message.

### Anything else to note
I'm not 100% if this may fix it but kubedm  uses the [`ClusterConfiguration`](https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-config/) cm stored in the cluster for subsequent `kubeadm join/reset/upgrade command`

The certSANs of the deleted nodes are still present in that config map, it would be worth to look into , if deleting them (as part of the delete node process) would help us with this issues or more works needs to be done.

Issue #1576, opened 2024-11-14
Feature: Move all decision logic into the Manager service


Despire:
### Motivation

The manager service was introduced as the "brain" of claudie where the difference between the current state and desired state is create which results in a number of tasks to be worked on by the builder to achieve the desired state.

### Description

Currently, not all the decision logic is present in Manager. Ideally, we should move them to the manager service when handling the Diff between the current and desired state. The other services Ansibler, Terraformer, Kuber, Kube-eleven should have only very minimal or ideally no decision logic in them.

### Exit criteria

- [x] Explore if the generation of [CIDR](https://github.com/berops/claudie/blob/905988069a36f39dd1406ae08d16791184c15d52/services/terraformer/server/domain/utils/cluster-builder/cluster_builder.go#L68-L75) can be moved into the manager service and not be delayed until terraformer stage.
- [x] Ansibler Setup/Teardown of Loadbalancers should be split depending on whether the API endpoint change is needed. For example If the endpoint change is needed it will first [Teardown](https://github.com/berops/claudie/blob/905988069a36f39dd1406ae08d16791184c15d52/services/builder/domain/usecases/ansibler_caller.go#L36-L52) the Lbs and then use the state from the teardown in the [Setup](https://github.com/berops/claudie/blob/905988069a36f39dd1406ae08d16791184c15d52/services/builder/domain/usecases/ansibler_caller.go#L114) The case of changing the API endpoint should be handled in its own task and then the rest of the Loadbalancers should be build, the logic of determining this should be moved to the manager instead of having this in Ansibler and Builder
- [x] Determine if there is anything else in all of the mentioned services that should be moved to the manager service.

PR #1634, opened 2025-01-15
Chore/move decisions into manager

Despire:
Closes https://github.com/berops/claudie/issues/1576
Closes https://github.com/berops/claudie/issues/1597

The manager service now handles what will be build and how by the other services.


JKBGIT1 (approved):
:+1:

Issue #1633, opened 2025-01-15
Feature: Increase limits on Nodes.


Despire:
When there are many workloads deployed on a node in a k8s cluster at some point the error `Too many open files` can be encountered.

Nodes deployed in clusters by claudie have the default limits specified and claudie does not change them.

This feature request is to increase the default limits to a higher number to allow more workloads to be deployed on individual nodes.

PR #1679, opened 2025-03-06
update limits

Despire:
Closes https://github.com/berops/claudie/issues/1633


services/kuber/server/domain/utils/autoscaler/cluster_autoscaler.go
bernardhalas, line 1: I'd separate this into a new PR as it's not about limits.

JKBGIT1 (approved):

Issue #1707, opened 2025-03-31
Feature: Upgrade Longhorn to 1.7.2 or higher


JKBGIT1:
### Motivation
A critical bug fix was fixed in Longhorn 1.7.2 and later. The bug resolves issues with RecurringJobs, which handles scheduled backups. More details can be found in the related issue:
https://github.com/longhorn/longhorn/issues/7814
https://github.com/longhorn/longhorn/issues/7814#issuecomment-2412540397

### Description
Longhorn's scheduled backup process gets stuck when backing up a specific volume, potentially lasting indefinitely. As a result:
- No backup is created for the affected volume.
- Subsequent backups for other volumes remain blocked.
- Longhorn never initiates another round of scheduled backups because the stuck process never completes.

### Exit criteria

- [ ] Verify whether Longhorn can be upgraded without data loss or other issues.
- [ ] Upgrade Longhorn to 1.7.2 or higher

PR #1709, opened 2025-04-01
chore: update dependencies & longhorn to 1.8.1

Despire:
General dependencies update across the repository.

Updates longhorn to 1.8.1. Closes https://github.com/berops/claudie/issues/1707
 This was tested by creating a deployment with pvc with loghorn 1.7.0 (the current version)
  Upgrade to longhorn 1.8.1, all pods were deployed successfully, perform additional operation on the cluster without issues.
  The hash of the replicas were the same after the upgrade.
  Further longhorn also specifies in the release that 1.8.x is backwards compatible with 1.7.x https://github.com/longhorn/longhorn/releases/tag/v1.8.1
  
  For volumes created with the old instance one can follow: https://longhorn.io/kb/troubleshooting-some-old-instance-manager-pods-are-still-running-after-upgrade/
  It's possible to set via the Longhorn UI `Settings > Concurrent Automatic Engine Upgrade Per Node Limit` to a value higher than 0 to upgrade old volumes to the newly deployed manager
  
  Old engine-image pods will be terminated after ~60 minutes of not being used (after all volumes are upgraded to the newest longhorn version)
  
  Updates go to 1.24.
  
  Updates go dependencies


JKBGIT1 (approved):
LGTM :+1:

Issue #1722, opened 2025-04-02
Bug: Deletion panics if DNS failed to be build


Despire:
When a cluster with a kube api-server loadbalancer is build with invalid DNS input (invalid API key, or zone) the DNS field of the Loadbalancer will be `nil` 
https://github.com/berops/claudie/blob/e7b2856847331d47f37b8ecf2b842d2423400965/proto/spec/manifest.proto#L119

When we then try to delete the cluster, it panics on the following line, as the `.DNS` is `nil`.

https://github.com/berops/claudie/blob/e7b2856847331d47f37b8ecf2b842d2423400965/services/builder/domain/usecases/kube_eleven_caller.go#L47

The deletion process should always work even if the infra was partially build.

We only consider a LoadBalancer to be a valid ApiEndpoint if the DNS was build successfully (i.e. is not `nil`), thus we need to add the condition here:

https://github.com/berops/claudie/blob/e7b2856847331d47f37b8ecf2b842d2423400965/services/builder/domain/usecases/kube_eleven_caller.go#L46-L48

mohitsethia:
Hey, I want to contribute to the issue, but I see that we are only logging the error received from destroyK8sCluster().
https://github.com/berops/claudie/blob/e7b2856847331d47f37b8ecf2b842d2423400965/services/builder/domain/usecases/workflow_helpers.go#L136
Can you please help me understand what needs to be done?

mohitsethia:
Thanks for the description, it is clear. Let me add a condition to avoid panics, just one doubt if we should still call the [DestroyK8sCluser](https://github.com/berops/claudie/blob/e7b2856847331d47f37b8ecf2b842d2423400965/services/builder/domain/usecases/kube_eleven_caller.go#L51) if lbApiEndpoint is a empty string or just return from line 47 in case of DNS is not valid?

Despire:
Yes, it should still be called, cause some of the infrastructure may have been build, and we need to destroy that to cleanup.

PR #1724, opened 2025-04-03
fix panic caused by nil DNS struct

mohitsethia:
*Description*
Fixes a potential panic during cluster deletion when the DNS field of a LoadBalancer is nil.

Currently, if a cluster is built with invalid DNS input (such as an invalid API key or zone), the DNS field of the LoadBalancer can be nil. This causes a panic when attempting to delete the cluster, as the deletion process tries to access ep.Dns.Endpoint without checking if ep.Dns is valid.

This PR adds a condition to ensure ep.Dns is not nil before attempting to access ep.Dns.Endpoint. This ensures that the deletion process works correctly, even when the infrastructure is only partially built.

fixes: #1722


Despire (approved):
LGTM 👍

PR #1729, opened 2025-04-08
Chore/decrease ansible cleanup

Despire:


services/ansibler/server/domain/usecases/remove_utilities.go
```diff
@@ -77,7 +79,14 @@ func removeUtilities(clusterID string, vpnInfo *VPNInfo, processLimit *semaphore
 	}
 
 	ansible := utils.Ansible{
-		RetryCount:        3,
+		// since removing claudie utilities should only be used on nodes that are either deleted, or the cluster is being
```
JKBGIT1, line 82: How long does it take until the IM fails when we know the nodes are unreachable (destroyed manually by the user before removing from the IM)?

JKBGIT1 (approved):

services/ansibler/server/domain/usecases/remove_utilities.go
Despire: The IM does not start a delete of the nodes automatically, it needs to be triggered by the user.

JKBGIT1: Okay, but how long will it take until the user gets the feedback about unreachable nodes after the trigger?

Despire: It will immediately check if the nodes in the curren state are reachable and if not will prompt the user to remove them from the input manifest.

As for how long the deletion will take place, it depends on the size of the cluster and how many workloads are there....

Issue #1575, opened 2024-11-14
Bug: Ansible-Playbook stuck


Despire:
Ansible playbooks used by claudie do not have any timeouts, ocasionally this leads to an infinite block in our e2e pipeline where ssh-ing into the pod is needed to execute `killall ansible-playbook` to force a retry. The infinite block usually happens when downloading package or installing wireguard. 

This issue was. created so that when this happens we document what lead to the block.

PR #1757, opened 2025-06-06
add timeouts to package installs

Despire:
Closes https://github.com/berops/claudie/issues/1575

Adds a timeout of 8mins for tasks that install packages.

Additionally, removed duplicate install of wireguard via pipx which was not utilized anywhere


bernardhalas (approved):
LGTM, thank you

PR #1777, opened 2025-07-14
DNS Loadbalancing feat

### LINKED ISSUE #1759: Bug: Loadbalancer does not exclude unresponsive public IP
### Current Behaviour
When a LoadBalancer VM becomes unresponsive, its external IP address is still being advertised and included in the DNS record. This causes periodic timeouts when clients attempt to reach the kubeAPI server.

### Expected Behaviour
Unresponsive external IPs associated with dead LoadBalancer VM should be temporarily excluded from the active IP pool until the IP become again responsive

### Steps To Reproduce
1. Create a manifest that provisions two LoadBalancers.
2. Ensure your domain resolves to both public IPs associated with these LoadBalancers.
```
host brando.claudie.dev
        brando.claudie.dev has address 34.79.250.193
        brando.claudie.dev has address 91.99.148.241
```
3. Use tcpdump to confirm traffic is being correctly routed to both LoadBalancers when calling the Kubernetes API from a client `tcpdump -lni any host 34.79.250.193 or host 91.99.148.241`
4. Simulate an outage of one LoadBalancer, e.g., by shutting down the associated VM (in my case VM with IP 91.99.148.241)
5. Use `dig` command to inspect DNS TTL and confirm the downed IP is still being returned
```
dig brando.claudie.dev
       ;; ANSWER SECTION:
       brando.claudie.dev.     300     IN      A       91.99.148.241
       brando.claudie.dev.     300     IN      A       34.79.250.193
```
6. Attempt to access the kubeAPI repeatedly. You’ll observe that roughly every other request incurs a ~10 second delay due to round-robin DNS still resolving to the unresponsive IP. This behavior can be monitored by running tcpdump from step 3.
7. Even after expiration of DNS entry for dead member, the DNS record is still returned with a new 300s TTL
### Consideration
1. Periodically verify whether each LB is responsive and in case of failed LB remove unresponsive IP from dns entry. E.g using cloudflare [Manage monitors](https://developers.cloudflare.com/load-balancing/monitors/create-monitor/)
2. Not all DNS providers support managed health checks. In such cases, consider using a local failover solution like [Keepalived](https://www.keepalived.org/) to handle unresponsive IPs. Note that this approach requires L4 network connectivity between LoadBalancer VMs to allow the exchange of VRRP messages for IP failover.

m-brando:
Closes #1759

- Added functionality for cloudflare to recognize accountId parameter from manifest. This param is required when building cloudflare's tofu resources cloudflare_load_balancer_pool and cloudflare_load_balancer_monitor
- Updated google tf provider to be able to use DNS loadbalancing
- To test functionality you need to use templates from https://github.com/berops/claudie-config/tree/bux/fix-domain-balancing


services/terraformer/server/domain/utils/templates/structures.go
```diff
@@ -164,6 +164,8 @@ type (
 		// Provider hold the information (credentials, external templates etc.)
 		// that were passed by the user in the InputManifest.
 		Provider *spec.Provider
+		// Subscription is the subscription paid to use DNS load balancing features in cloudflare
+		CloudflareSubscription bool
```
Despire: We can't directly add new fields to the structure mostly due to compatibility reasons.

Someone running older Claudie versions might migrate to newer templates which use this field and will fail and their infra will be stuck until resolved manually.

Refactor this into the following (similar to the AlternativeNames extensions):

```diff
	// DNS wraps all data related to generating terraform files to spawn create the specified DNS
	// infrastructure as specified in the InputManifest.
	DNS struct {
		AlternativeNamesExtension *AlternativeNamesExtension
+		ProviderExtrasExtension   *ProviderExtrasExtension

		// ClusterName is the name of the Loadbalancer cluster the DNS is to be created for.
		ClusterName string
		// ClusterHash is the hash of the Loadbalancer cluster the DNS is to be created for.
		ClusterHash string
		// Hostname is the hostname specified in the InputManifest.
		Hostname string
		// DNSZone is the zone in which the DNS is to be created as specified in the InputManifest.
		DNSZone string
		// RecordData holds IP addresses of all the nodes of the Loadbalancer cluster the DNS is to be created for.
		RecordData RecordData
		// Provider hold the information (credentials, external templates etc.)
		// that were passed by the user in the InputManifest.
		Provider *spec.Provider
	}

	// AlternativeNamesExtension adds additional data to be passed to the template that can then be used.
	// To check for the presence of the extension within the templates the `hasExtension` function can
	// be used for the check.
	//
	// {{- if hasExtension .Data "AlternativeNamesExtension"}}
	// ... access .AlternativeNamesExtension
	// {{- end }}
	AlternativeNamesExtension struct {
		// Names are additional A records that will be created in addition to the hostname.
		// The use case is for a single load balancer to handle traffic from M domains instead of just 1.
		Names []string
	}

+	// ProviderExtrasExtension adds additional data to be passed to the template that can then be used.
+	// To check for the presence of the extension within the templates the `hasExtension` function can
+	// be used for the check.
+	//
+	//
+	// {{- if hasExtension .Data "ProviderExtrasExtension"}}
+	// ... access .ProviderExtrasExtension
+	// {{- end }}
+	ProviderExtrasExtension struct {
+		// SubscriptionAllowsHA reports whether the subscription associated with the passed in credentials
+		// allow High Available Load Balancing to be used for the created A-records for both the [DNS.Hostname]
+		// and [DNS.AlternativeNamesExtension]. Currently this is only needed for Cloudflare, more providers
+		// may be added in the future.
+		SubscriptionAllowsHA bool
	}
```

Then in the `dns.go` file when generating the templates you can simply do the following:

```diff
	data := templates.DNS{
		DNSZone:     dns.DnsZone,
		Hostname:    dns.Hostname,
		ClusterName: d.ClusterName,
		ClusterHash: d.ClusterHash,
		RecordData:  templates.RecordData{IP: templateIPData(nodeIPs)},
		Provider:    dns.Provider,

		AlternativeNamesExtension: new(templates.AlternativeNamesExtension),
+		ProviderExtrasExtension:   new(templates.ProviderExtrasExtension),
	}

+	if cloudflare := dns.Provider.GetCloudflare(); cloudflare != nil {
+		var err error
+		data.ProviderExtrasExtension.SubscriptionAllowsHA, err = cloudflare.GetCloudflareSubscription(logger, cloudflare.AccountID, cloudflare.Token)
+		if err != nil {
+			return fmt.Errorf("error while checking cloudflare load balancing subscription: %w", err)
+		}
+	}
...
```

The line `ProviderExtrasExtension:   new(templates.ProviderExtrasExtension)` means that the current version of claudie supports the `ProviderExtras` extension. Older version will not have it and thus will not support it.

Then in the templates you need to make a check for this 

```
{{- if hasExtension .Data "ProviderExtrasExtension"}}
 ... access .ProviderExtrasExtension
{{- end }}
```

As of now this will be used within the templates for cloudflare (as this is the only provider it is used, as of now)

proto/pb/spec/utils.go
```diff
@@ -214,3 +219,65 @@ func (r *Role) MergeTargetPools(o *Role) {
 		}
 	}
 }
+
+// GetCloudflareSubscription checks if the Cloudflare account has a Load Balancing subscription.
+func (x *CloudflareProvider) GetCloudflareSubscription(logger zerolog.Logger, accountID string, apiToken string) (bool, error) {
```
Despire: This function already has a cloudflare receiver, there is no need to include `CloudFlare` in the function name.

`GetCloudflareSubscription` -> `GetSubscription`

proto/pb/spec/utils.go
```diff
@@ -214,3 +219,65 @@ func (r *Role) MergeTargetPools(o *Role) {
...
+	for _, subscription := range subscriptions.Result {
+		if subscription.Product.Name == "prod_load_balancing" && subscriptions.Success == true {
+			sublogger.Info().Msgf("Found subscription for %s", subscription.Product.Name)
+			return true, nil
+		}
+	}
+	return false, fmt.Errorf("Subscription for Load Balancing not found")
+}
+
+func getCloudflareAPIResponse(url string, apiToken string) ([]byte, error) {
```
Despire, line 259: - You should check for the response status code.
- Instead of creating a http Client every time simply default to using `http.DefaultClient`
- Add a request timeout so that it does not hang forever.
- use `http.MethodGet` instead of string constant `GET`

The function could look something more like this

```diff
func getCloudflareAPIResponse(url string, apiToken string) ([]byte, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()

+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("could not create request: %w", err)
+	}

-       req, err := http.NewRequest("GET", url, nil)
-	if err != nil {
-		return nil, fmt.Errorf("could not create request: %w", err)
-	}

	req.Header.Set("Authorization", "Bearer "+apiToken)
	req.Header.Set("Content-Type", "application/json")
	
-	client := &http.Client{}

-	resp, err := client.Do(req)

+	resp, err := http.DefaultClient.Do(req)
	if err != nil {
		return nil, fmt.Errorf("Error making http request: %w", err)
	}
	defer resp.Body.Close()

+      // nolint
+	if !(resp.StatusCode >= 200 && resp.StatusCode < 300) {
+		return nil, fmt.Errorf("response with status code %v: %v", resp.StatusCode, resp.Status)
+	}

	body, err := io.ReadAll(resp.Body)
	if err != nil {
		return nil, fmt.Errorf("Error reading response body: %w", err)
	}

	return body, nil
}
```

proto/pb/spec/utils.go
```diff
@@ -214,3 +219,65 @@ func (r *Role) MergeTargetPools(o *Role) {
...
+			} `json:"product"`
+		} `json:"result"`
+		Success bool `json:"success"`
+	}
+
+	escapedAccountID := url.PathEscape(accountID)
+	urlSubscriptions := fmt.Sprintf("https://api.cloudflare.com/client/v4/accounts/%s/subscriptions", escapedAccountID)
+
+	responseSubscriptions, err := getCloudflareAPIResponse(urlSubscriptions, apiToken)
+
```
Despire: (nit) Omit this newline white space, so that `if err != nil` follows immediately

Despire (changes requested):
PR looks ok 👍 , but we will need to do a few changes which I've listed.

Also fix the linter issues, you can run the linter locally via `make lint`

services/terraformer/server/domain/utils/loadbalancer/dns.go
```diff
@@ -257,13 +257,24 @@ func (d *DNS) generateFiles(dnsID, dnsDir string, dns *spec.DNS, nodeIPs []strin
...
 
+	var cloudflareSubscription bool
+	if dns.Provider.GetCloudProviderName() == "cloudflare" {
+		var err error
+		cloudflareSubscription, err = dns.Provider.GetCloudflare().GetCloudflareSubscription(
+			logger, dns.Provider.GetCloudflare().GetAccountID(), dns.Provider.GetCloudflare().GetToken())
+		if err != nil {
+			return fmt.Errorf("Error while checking cloudflare Load Balancing subscription %w:", err)
+		}
+	}
```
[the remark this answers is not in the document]
Despire: This should be fixed once the linter warning are addressed.

Despire (approved):

Issue #1775, opened 2025-07-04
Chore: Upgrade provider versions across the templates


bernardhalas:
### Description
Upgrade TF provider versions. Be mindful of potential breaking changes in the provider releases.

### Exit criteria

- [ ] Providers are upgraded across all the templates

PR #1782, opened 2025-07-21
Upgraded tf providers to the current latest.(#1775)

m-brando:
works with this pull request https://github.com/berops/claudie-config/pull/8 which proposes changes in tf templates


Despire (approved):
LGTM 👍

Issue #1779, opened 2025-07-16
Bug: autoscaler pod doesn't release the memory


JKBGIT1:
### Current Behaviour

Every time an autoscaler triggers scaling, its memory usage increases. This memory is never released, even when there aren't any ongoing workflows (see the image below).

<img width="1609" height="279" alt="Image" src="https://github.com/user-attachments/assets/6df18033-8ead-4097-92dc-fd539027bd1f" />

### Expected Behaviour

The autoscaler stops piling up the memory and releases it after scaling.

### Steps To Reproduce
1. Create an InputManifest with the `autoscaler` nodepool
2. Keep triggering the scaling
3. Stop the scaling
4. Check the autoscaler pod's memory usage. You should see it kept growing over time and never goes down.

PR #1787, opened 2025-07-23
fix memory leak

Despire:
Closes https://github.com/berops/claudie/issues/1779

The memory leak due to not properly clossing the created gRPC connections.
Snapshots were taken 30 min apart, to idenfity where the leak occured.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resource management to prevent potential connection leaks during autoscaler operations. Connections are now properly closed after use, enhancing reliability.
* **Chores**
  * Updated container image tags for autoscaler-adapter and kuber components to latest versions.

### DIFF
--- a/manifests/claudie/kustomization.yaml
+++ b/manifests/claudie/kustomization.yaml
@@ -59,7 +59,7 @@ images:
 - name: ghcr.io/berops/claudie/ansibler
   newTag: 6223444-3450
 - name: ghcr.io/berops/claudie/autoscaler-adapter
-  newTag: 6223444-3450
+  newTag: ae0563d-3461
 - name: ghcr.io/berops/claudie/builder
   newTag: 6223444-3450
 - name: ghcr.io/berops/claudie/claudie-operator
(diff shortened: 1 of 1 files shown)


services/autoscaler-adapter/claudie_provider/claudie_node_groups.go
```diff
@@ -177,6 +177,11 @@ func (c *ClaudieCloudProvider) updateNodepool(ctx context.Context, nodepool *spe
 	if err != nil {
```
m-brando, line 177: This could be a one-liner to match our usual error handling style.

jakubhlavacka: I don't think this will work cause you need to assign the `manager` variable and want to use it outside of the `if` clause.

JKBGIT1 (approved):

services/autoscaler-adapter/claudie_provider/claudie_node_groups.go
Despire: you would have to do
```go
var manager *Manger
var err error
if manager, err = ....; err != nil {
}
```

the `if err := ...; err != nil ` is mostly used for error shadowing

Issue #1784, opened 2025-07-22
Bug: Deleted Node still tracked inside Current State


Despire:
### Current Behaviour

If an error occurs during the deletion of the node after the terraformer state, the current state will not be updated and will be left as is for it to be retried in the future. However, it can happen that the public IP associated with the VM that was deleted was assigned to someone else and claudie will consider that VM
as reachable and not schedule the deletion again.

### Expected Behaviour

After the VM instance and its public IP is deleted from terraformer this should be reflected in the current state
in both cases, partial failure and or success

Reproducing might be tricky, due to the nature of how public IPs are distributed on each provider, but generally

1. Delete a node from the cluster. The step where the node is deleted from the k8s cluster success and the terraformer step, and the next one in ansibler fails.
2. Let the IP be assigned so that it will be reachable.
3. Claudie will consider the node reachable and not schedule a deletion for it.

PR #1793, opened 2025-07-28
fix: return partially updated state instead of current state after deletion error

Despire:
Closes https://github.com/berops/claudie/issues/1784

The issue was in returning the current state on error during deletion, when we should have return the desired state (even if only partial changes were done)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy of cluster state reporting after node deletion failures, ensuring the updated state is communicated correctly to the manager.
* **Chores**
  * Updated container image to a newer version for improved stability and performance.


m-brando (approved):
LGTM

Issue #1853, opened 2025-09-19
Feature: Expand Autoscaler instance type cache to also include GPUs


Despire:
### Description
Currently the instance type cache in the autoscaler implementation https://github.com/berops/claudie/blob/e7265870852ccd9acfbce9fe9064a5ef40ff2cb4/services/autoscaler-adapter/node_manager/node_manager.go#L19 does not account for GPUs of the instance type.

This caching mechanism should be extended to also support GPU's and include them in the node template https://github.com/berops/claudie/blob/e7265870852ccd9acfbce9fe9064a5ef40ff2cb4/services/autoscaler-adapter/claudie_provider/node_template.go#L23 

https://github.com/berops/claudie/blob/e7265870852ccd9acfbce9fe9064a5ef40ff2cb4/services/autoscaler-adapter/node_manager/node_manager.go#L65

that is passed backed to the cluster-autoscaler.

### Exit criteria

- [ ] To make autoscaling also work with instance-types that have GPU's expand the missing features in the autoscaler implementation

Issue #1772, opened 2025-06-22
Feature: Kubernetes Addon Selection and Defaults Override


bernardhalas:
### Motivation
Claudie intends to be a fully usable Kubernetes distribution with little setup effort. For that reason, it needs to come pre-packaged with storage system (Longhorn) or observability stack (Prometheus, Alertmanager, Grafana,...). However, the defaults, that these addons come with do not suit all users and they wish to have a possibility to override them.

### Description
- Design a way how to opt-out from addons.
- Design a way how to override defaults for the addons.

### Exit criteria

- [ ] Users are able to opt-in/opt-out from the addons
- [ ] Users are able to override default addons deployment manifests.

bernardhalas:
The proposal is to utilize [Sveltos|https://sveltos.projectsveltos.io] for templating Kubernetes manifests. Sveltos can load values from both management and managed clusters, from ConfigMaps and from Secrets. This would mean that Claudie will be deployed together with Sveltos manifests on the management clusters, and Claudie will create the necessary ConfigMaps and Secrets, for Sveltos to consume.

Now, for customizing Kubernetes manifests the users are recommended to use [kustomize|https://kustomize.io]. This would mean, that Claudie will provide manifest bases and users will just supply the overlay for those bases.

This means, there would be 2 repositories to come into play:
1. `berops/claudie-kustomize-bases` for providing default Kustomize bases for Claudie addons; with a very rare needs for forking
2. `berops/claudie-templates` for providing overlays, patches and opt-ins/opt-outs for for Claudie addons; expected to be overriden by users for modifying the default Claudie behavior.

e.g.:

`https://github.com/berops/claudie-templates/addons/kustomize.yaml`:
```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources:
- https://github.com/berops/claudie-kustomize-bases//longhorn?ref=v0.9.10
- https://github.com/berops/claudie-kustomize-bases//prometheus?ref=v0.9.10

patches:
- patches/nvme-storage-class.yml
- patches/observability-taints.yml
```

Related to #1758 .

PR #1854, opened 2025-09-19
chore: expand machine spec to contain number of gpus

Despire:
Expands the machine spec to describe the number of GPUs the given instance type has. This will then be used in the Node Template for the cluster-autoscaler.

Partially also implements https://github.com/berops/claudie/issues/1853 where applicable. But it seems that for not all of the Claudie supported providers we can programmatically determine the number of NVIDIA GPUs attached.

Thats why the change in the InputManifest for a dynamic nodepool:

```
machineSpec:
  nvidiaGpu: 1
```

## Summary by CodeRabbit

- New Features
  - Dynamic nodepools can declare NVIDIA GPU counts; autoscaler exposes and honors GPU resources (per-node spec overrides provider info).

- Documentation
  - API reference and autoscaling guide updated with NVIDIA GPU field, clarified CPU/Memory wording, and example GPU-enabled nodepool.

- Bug Fixes
  - Autoscaler now guards against missing dynamic nodepool data and returns safely when capacity info is absent.

- API Changes
  - CPU and Memory validations relaxed to be optional/mutually dependent; CRD and proto schemas include nvidiaGpu.

- Tests
  - Diff-based tests made deterministic by sorting outputs.

- Chores
  - YAML/manifest formatting, image tags, and test cluster versions updated.

### DIFF
--- a/manifests/claudie/kustomization.yaml
+++ b/manifests/claudie/kustomization.yaml
@@ -1,74 +1,74 @@
 namespace: claudie
 resources:
-  - crd
-  - ns.yaml
-  - operator.yaml
-  - builder.yaml
-  - terraformer.yaml
-  - ansibler.yaml
-  - kube-eleven.yaml
-  - kuber.yaml
-  - manager.yaml
-  - cluster-rbac
-  - mongo
-  - minio
-  - dynamo
+- crd
+- ns.yaml
+- operator.yaml
+- builder.yaml
+- terraformer.yaml
+- ansibler.yaml
+- kube-eleven.yaml
+- kuber.yaml
+- manager.yaml
+- cluster-rbac
+- mongo
+- minio
+- dynamo
 
 # Alter ValidatingWebhookConfiguration and Certificate fields, so they will match the generated namespace
 replacements:
-  - source:
-      fieldPath: metadata.name
-      kind: Namespace
-    targets:
-      - fieldPaths:
-          - webhooks.*.namespaceSelector.matchLabels.kubernetes\.io/metadata\.name
-          - webhooks.*.clientConfig.service.namespace
-        select:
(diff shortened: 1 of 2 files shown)


manifests/claudie/crd/claudie.io_settings.yaml
```diff
@@ -1,84 +0,0 @@
----
```
jakubhlavacka: Why do you remove this file?

Despire: This is a leftover from the CRD that was introduced in the https://github.com/berops/claudie/pull/1735, where the CRD was suppose to be used to dynamically adjust the default templates, but the CRD was removed in favor of https://github.com/berops/claudie/issues/1772, but the generated CRD was accidentally left there. I'll remove it in a different PR and don't include it here.

Despire:
> I don't understand why we need to update the InputManifest

For the Autoscaling to work properly when we pass the Template of a Node From a NodePool  https://github.com/berops/claudie/blob/e7265870852ccd9acfbce9fe9064a5ef40ff2cb4/services/autoscaler-adapter/node_manager/node_manager.go#L65

we Need to tell it that it has GPU capacity available, which we currently do not do. But Also it seems that for most Cloud providers supported by Claudie there doesn't seem to be a way programatically to resolve if the instance has GPU's and if they're from Nvidia.

For Aws this is not an issue, but for the others it is, for example Google treats GPUs as someting as an "extension" and you can configure 1,2,4,8 GPUs, and you don't know programatically determining this, at least not for now,

So that's why a new field was added `machineSpec.NvidiaGpus` to tell claudie how many GPUs the instance-type of the nodepool has so that autoscaling with GPUs also works okay.

PR #1856, opened 2025-09-23
fix: noop task

Despire:
When a nodepool has 0 nodes, an additional task will be spawn for adding nodes even if we just change labels or annotations, which is unwanted.

This is because we do not include nodepool with 0 nodes in the diff, which this PR fixes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Zero-node autoscaled and static node pools are now represented explicitly as empty lists, ensuring consistent display in views, dashboards, API responses, and tooling.

* **Chores**
  * Updated image tags in deployment manifests for testing and runtime components to the latest build references.

### DIFF
--- a/manifests/claudie/kustomization.yaml
+++ b/manifests/claudie/kustomization.yaml
@@ -57,18 +57,18 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: ghcr.io/berops/claudie/ansibler
-  newTag: 54d8db7-3578
+  newTag: ae81a20-3587
 - name: ghcr.io/berops/claudie/autoscaler-adapter
-  newTag: 54d8db7-3578
+  newTag: ae81a20-3587
 - name: ghcr.io/berops/claudie/builder
-  newTag: 54d8db7-3578
+  newTag: ae81a20-3587
 - name: ghcr.io/berops/claudie/claudie-operator
-  newTag: 54d8db7-3578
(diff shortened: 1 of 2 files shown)


services/manager/internal/service/schedule_tasks.go
```diff
@@ -720,11 +720,13 @@ func NodePoolNodes(cluster *spec.K8Scluster) (map[string][]string, map[string][]
 
 	for _, nodePool := range cluster.GetClusterInfo().GetNodePools() {
 		if nodePool.GetDynamicNodePool() != nil {
+			dynamic[nodePool.Name] = []string{} // the nodepool could have 0 nodes, i.e. autoscaling to 0.
 			for _, node := range nodePool.Nodes {
 				dynamic[nodePool.Name] = append(dynamic[nodePool.Name], node.Name)
 			}
 		}
 		if nodePool.GetStaticNodePool() != nil {
+			static[nodePool.Name] = []string{} // the nodepool could have 0 nodes.
```
jakubhlavacka, line 729: Is it even possible to have a static node pool with 0 nodes? AFAIK, static nodes can't be autoscaled, so how would one achieve having 0 nodes in the static node pool?

Despire: Yes, you technically can have a static nodepool defined in the InputManifest, but with no Nodes, basically having a node count with 0 (This can also be done with dynamic nodepools)

Issue #1885, opened 2025-11-12
Bug: Docker ApiVersion


Despire:
While running any worklow with loadbalancers, the workflow will fail in the ansibler service with

```
LB - apiserver-lb-4imhyjr	TASK [run docker compose] ******************************************************
....
KeyError: 'ApiVersion'\r\n
```

PR #1886, opened 2025-11-12
Fix/docker issue

Despire:
Closes https://github.com/berops/claudie/issues/1885

Updates Dockerfiles and fixes DockerAPI issue.

The docker.compose ansible package has been removed in favor of shell commands

## Summary by CodeRabbit

* **Chores**
  * Base images bumped to Alpine 3.22 across services; Python and Ansible runtime components updated. Final images now include bash and bundled playbooks. Docker label formatting normalized.
  * Multiple deployment image tags and test-set image references updated.

* **Maintenance**
  * Deployment compose step replaced with an explicit shell-based startup to enforce deterministic launch behavior.

### DIFF
--- a/manifests/claudie/kustomization.yaml
+++ b/manifests/claudie/kustomization.yaml
@@ -57,18 +57,18 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: ghcr.io/berops/claudie/ansibler
-  newTag: fc45b61-3669
+  newTag: e147639-3670
 - name: ghcr.io/berops/claudie/autoscaler-adapter
-  newTag: fc45b61-3669
+  newTag: e147639-3670
 - name: ghcr.io/berops/claudie/builder
-  newTag: fc45b61-3669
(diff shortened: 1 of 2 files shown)


services/ansibler/Dockerfile
```diff
@@ -31,16 +31,17 @@ RUN apk add --no-cache \
     openssh-client \
     libgcc \
     yaml \
-    && ln -s /usr/bin/python3 /usr/bin/python \
-    && pip3 install --no-cache-dir ansible==10.1.0
+    && ln -sf /usr/bin/python3 /usr/bin/python \
+    && pip3 install --no-cache-dir --break-system-packages ansible==12.2.0
```
jakubhlavacka, line 35: I don't like `--break-system-packages`. Is it necessary?

Despire: Yes if we wan't to use the latest ansible version. Otherwise we would have to default to the base one in the alpine image which is older

JKBGIT1 (approved):
LGTM

PR #1903, opened 2025-11-25
Fix: drift with cloudflare account id when upgrading claudie versions

Despire:
Fixes a drift in cloudflare account id when updating to claudie versions.

The drift would result in the error

```
2025-11-25T17:30:35Z ERR Error encountered while building cluster: failed to create DNS record for apiserver-lb-2: error 
while creating current state dns.tf files for apiserver-lb-2-rzq3tiy-dns : error while checking cloudflare load balancing
 subscription: error while getting cloudflare api response for 'accounts/subscriptions', after 3 retries: response with status 
code 404: 404 Not Found cluster=apiserver-lb-2-rzq3tiy module=terraformer project=default-aws-hybrid-cloud

```

## Summary by CodeRabbit

* **Bug Fixes**
  * Missing Cloudflare AccountIDs in the current state are now copied from the desired state during reconciliation to improve backward compatibility.
  * Informational logging added when AccountIDs are transferred to increase visibility during state sync.

### DIFF
--- a/services/manager/internal/service/existing_state.go
+++ b/services/manager/internal/service/existing_state.go
@@ -30,7 +30,7 @@ func backwardsCompatibilityTransferMissingState(c *spec.Config) {
 						log.
 							Info().
 							Str("cluster", current.GetClusterInfo().Id()).
-							Msg("detected drift in current state for Cloudflare AccountID, transfering state from desired state")
+							Msg("detected drift in current state for Cloudflare AccountID, transferring state from desired state")
 						cc.AccountID = dc.AccountID
 					}
 				}


services/manager/internal/service/existing_state.go
```diff
@@ -5,12 +5,40 @@ import (
...
+func backwardsCompatiblityTransferMissingState(c *spec.Config) {
+	for _, state := range c.GetClusters() {
+		for _, current := range state.GetCurrent().GetLoadBalancers().GetClusters() {
+			// TODO: remove in future versions, cloudflare account id may not be correctly
+			// propagated to the current state when upgrading claudie versions, since the
+			// [manifest.Cloudflare.AccountID] has a validation which requires the presence
+			// of a valid, non-empty `account_id`, which might be missing in the current state,
+			// that will result in errors on subsequent workflows, simply transfer the `account_id`
+			// from the desired state to the current state, only if it's empty. That will take care
+			// of the drift introduced during claudie updates.
```
bernardhalas, line 24: Shall we log an issue for this, so that we get back to it in the future for the cleanup when it's no longer needed?

Despire: For the `TODO:` ?

I don't think we need an issue for that.

Issue #1898, opened 2025-11-20
Feature: Utilize native S3 OpenTofu locking


JKBGIT1:
### Motivation

With the [native S3 state locking in OpenTofu](https://opentofu.org/blog/opentofu-1-10-0/), we should be able to remove the dependency on DynamoDB and rely solely on MinIO for state locking. The goal is to reduce the number of different databases that Claudie depends on.

### Description

Explore and integrate OpenTofu’s native S3-based locking to replace the current DynamoDB-based mechanism.

### Exit criteria

- [ ] Explore the solution
- [ ] Implement native S3 state locking using MinIO
- [ ] Put an end to DynamoDB

PR #1906, opened 2025-11-26
Use native S3 state locking in `terraformer`

jakubhlavacka:
This PR implements native S3 state locking (using MinIO) + puts an end to DynamoDB.

closes https://github.com/berops/claudie/issues/1898

To remove dynamo from source code and docs, I have searched for `dynamo` and `create-table` occurances. Are there any other keywords to check?

## Summary by CodeRabbit

* **Refactoring**
  * Removed local DynamoDB usage and lock handling across the project; state locking consolidated to MinIO/S3-style.

* **Documentation**
  * Removed DynamoDB references and env vars from deployment, backup, troubleshooting, getting-started, and local-testing guides; updated workflow diagrams and explanations to reflect MinIO-only locking/state.

* **Chores**
  * Deleted DynamoDB manifests/configs and removed related runtime dependencies; updated tooling/image tags for Terraformer components.

### DIFF
--- a/manifests/claudie/kustomization.yaml
+++ b/manifests/claudie/kustomization.yaml
@@ -56,18 +56,18 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: ghcr.io/berops/claudie/ansibler
-  newTag: 1aa93b2-3703
+  newTag: d2f7fcd-3704
 - name: ghcr.io/berops/claudie/autoscaler-adapter
-  newTag: 1aa93b2-3703
+  newTag: d2f7fcd-3704
 - name: ghcr.io/berops/claudie/builder
-  newTag: 1aa93b2-3703
+  newTag: d2f7fcd-3704
 - name: ghcr.io/berops/claudie/claudie-operator
-  newTag: 1aa93b2-3703
+  newTag: d2f7fcd-3704
 - name: ghcr.io/berops/claudie/kube-eleven
-  newTag: 1aa93b2-3703
+  newTag: d2f7fcd-3704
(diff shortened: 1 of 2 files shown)


services/terraformer/server/domain/usecases/destroy_infrastructure.go
```diff
@@ -83,11 +77,8 @@ func (u *Usecases) DestroyInfrastructure(ctx context.Context, request *pb.Destro
 		logger.Info().Msgf("Successfully deleted Templates files for cluster %q", cluster.Id())
 
 		// In case of LoadBalancer type cluster,
-		// there are additional DNS related tofu state and state-lock files.
+		// there are additional DNS related tofu state.
 		if _, ok := cluster.(*loadbalancer.LBcluster); ok {
-			if err := u.DynamoDB.DeleteLockFile(ctx, request.ProjectName, cluster.Id(), dnsKeyFormatLockFile); err != nil {
-				logger.Warn().Msgf("Failed to delete lock file for %q-dns, assumming it was deleted/not created", cluster.Id())
-			}
```
Despire, line 90: Have you looked into MinIO, if there is anything related to the lock file, after the cluster  is deleted ?
'cause even when we used dynamodb we had to manually remove the lockfile, I wonder if this is now done automatically or not.

jakubhlavacka: Yes, I have. Currently, we have a state file and a `.tflock` file in MinIO. However, we don't have to delete the `.tflock` file because OpenTofu handles that. The `.tflock` file is created before an OpenTofu operation starts and deleted once the operation ends.

PR #1907, opened 2025-11-26
Feat/tofu cache (#1844)

### LINKED ISSUE #1844: Feature: Tofu providers cache
### Motivation

To speed up the building of the cluster, we can shave off a few seconds, or even up-to some minutes in maintaining a tofu providers plugin cache.

### Description

Explore if the env `TF_PLUGIN_CACHE_DIR=` is suitable to implement a provider plugin cache so that they do not need to be re-downloaded every-time a change happens inside the cluster.

### Exit criteria

- [ ] explore if `TF_PLUGIN_CACHE_DIR=` can be used for implementing a provider plugin cache

m-brando:
This PR implements tofu cache by setting the TF_PLUGIN_CACHE_DIR
environment variable.

It also updates how provider hashes are calculated in the Tofu lock file.
When a provider binary is available in the local cache, its contents are
used as the input to the hash function. If the provider is not found
in the cache, the hash is computed using data fetched from the OpenTofu registry.

closes #1844

## Summary by CodeRabbit

* **New Features**
  * Plugin cache directory support for provisioning and Terraform flows.
  * Provider locking added with non-fatal fallback (warning logs).
  * Commands now propagate custom environment variables.
  * Per-cluster temporary files for kube/config, kube-proxy and kubeadm patching with automatic cleanup.

* **Infrastructure**
  * Mounted cache volume for the provisioning component.
  * Updated container image tags across multiple services.

* **Tests**
  * Bumped Kubernetes versions in test manifests to 1.34.0 (string).

### DIFF
--- a/manifests/claudie/kustomization.yaml
+++ b/manifests/claudie/kustomization.yaml
@@ -56,18 +56,18 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: ghcr.io/berops/claudie/ansibler
-  newTag: 5498ab1-3731
+  newTag: 4f9e418-3735
 - name: ghcr.io/berops/claudie/autoscaler-adapter
-  newTag: 5498ab1-3731
+  newTag: 4f9e418-3735
 - name: ghcr.io/berops/claudie/builder
-  newTag: 5498ab1-3731
+  newTag: 4f9e418-3735
 - name: ghcr.io/berops/claudie/claudie-operator
-  newTag: 5498ab1-3731
+  newTag: 4f9e418-3735
 - name: ghcr.io/berops/claudie/kube-eleven
-  newTag: 5498ab1-3731
+  newTag: 4f9e418-3735
 - name: ghcr.io/berops/claudie/kuber
-  newTag: 5498ab1-3731
+  newTag: 4f9e418-3735
 - name: ghcr.io/berops/claudie/manager
-  newTag: 5498ab1-3731
+  newTag: 4f9e418-3735
 - name: ghcr.io/berops/claudie/terraformer
-  newTag: 5498ab1-3731
+  newTag: 4f9e418-3735
--- a/manifests/testing-framework/kustomization.yaml
+++ b/manifests/testing-framework/kustomization.yaml
@@ -90,4 +90,4 @@ secretGenerator:
 
 images:
 - name: ghcr.io/berops/claudie/testing-framework
-  newTag: 5498ab1-3731
+  newTag: 4f9e418-3735


services/terraformer/server/domain/utils/tofu/terraform.go
```diff
@@ -37,14 +42,54 @@ type Terraform struct {
...
+	stderrBuf := &bytes.Buffer{}
+	cmd.Stderr = stderrBuf
+
+	if err := cmd.Run(); err != nil {
+		// In case that cache mirror does not contain the required providers
+		// continue and fetch them from opentofu registry.
+		if strings.Contains(stderrBuf.String(), "Could not retrieve providers for locking") {
+			log.Info().Msg("OpenTofu failed to fetch the requested providers from mirror. Proceed to get providers from opentofu registry.")
+		} else {
+			return fmt.Errorf("failed to execute cmd: %s: %s", cmd, stderrBuf.String())
```
jakubhlavacka: What are the other errors that can appear?
Shouldn't we continue and fetch the providers from opentofu registry in any case?

Despire: I agree, 

```go
if err := cmd.Run(); err != nil {
 return fmt.Errorf(....)
}
```

And when calling the `ProvidersLock()`

The error can by just a warning and the flow can continue which will default to download the providers.

```go
	if err := tofu.ProvidersLock(); err != nil {
		logger.Warn().Msgf("error while locking providers in %v", err))
	}
```

Despire: I agree, 

```go
if err := cmd.Run(); err != nil {
 return fmt.Errorf(....)
}
```

And when calling the `ProvidersLock()`

The error can by just a warning and the flow can continue which will default to download the providers.

```go
	if err := tofu.ProvidersLock(); err != nil {
		logger.Warn().Msgf("error while locking providers in %v", err))
	}
```

m-brando: > What are the other errors that can appear?

The one I'm catching was the only error I encountered during testing. However, based on the [code](https://github.com/opentofu/opentofu/blob/a89667e29cc1653bd7a02a0ef7695cbe77747d15/internal/command/providers_lock.go#L43), also this error could occured at some point 
`Could not create temporary directory`

I guess, this would indicate some problem with FS that need manual intervention.

m-brando: I didn't refresh the page so I haven't seen @Despire  comments on this :)
But I agree, I will log the rest of errors as Warnning and continue with normal exexution.

services/terraformer/server/domain/utils/tofu/terraform.go
```diff
@@ -37,14 +42,57 @@ type Terraform struct {
...
+	absCache, err := filepath.Abs(t.CacheDir)
+	if err != nil {
+		return fmt.Errorf("failed to resolve absolute cache dir: %w", err)
+	}
+
+	if _, err := os.Stat(absCache); os.IsNotExist(err) {
+		if err := os.MkdirAll(absCache, os.ModePerm); err != nil {
+			return fmt.Errorf("failed to create cache directory %s : %w", absCache, err)
+		}
+	}
```
Despire: you can use `fileutils.CreateDirectory` utility function for creating the cache directory.

services/terraformer/server/domain/utils/tofu/terraform.go
```diff
@@ -37,14 +43,57 @@ type Terraform struct {
...
+	if err != nil {
+		return fmt.Errorf("failed to resolve absolute cache dir: %w", err)
+	}
+
+	if _, err := os.Stat(absCache); os.IsNotExist(err) {
+		if err := fileutils.CreateDirectory(absCache); err != nil {
+			return fmt.Errorf("failed to create cache directory %s : %w", absCache, err)
+		}
+	}
+
```
Despire: Replace all of this with `fileutils.CreateDirectory`.

services/terraformer/server/domain/utils/tofu/terraform.go
```diff
@@ -37,14 +43,55 @@ type Terraform struct {
...
+	stderrBuf := &bytes.Buffer{}
+	cmd.Stderr = stderrBuf
+
+	if err := cmd.Run(); err != nil {
+		// In case that cache mirror does not contain the required providers
+		// continue and fetch them from opentofu registry.
+		if strings.Contains(stderrBuf.String(), "Could not retrieve providers for locking") {
+			log.Info().Msg("OpenTofu failed to fetch the requested providers from mirror. Proceed to get providers from opentofu registry.")
+		} else {
```
Despire: Return the `err` here instead of matching against a `string`, the calling code should handle that error, which in our case will be logged and ignored.

thus simply do 

```go
if err := cmd.Run(); err != nil {
 return err
 }
 ```

File: services/terraformer/server/domain/utils/tofu/terraform.go [hunk 5]
```diff
@@ -37,14 +43,55 @@ type Terraform struct {
...
 	defer t.SpawnProcessLimit.Release(1)
 
+	absCache, err := filepath.Abs(t.CacheDir)
+	if err != nil {
+		return fmt.Errorf("failed to resolve absolute cache dir: %w", err)
+	}
+
 	//nolint
 	cmd := exec.Command("tofu", "init")
+	cmd.Env = append(os.Environ(), "TF_PLUGIN_CACHE_DIR="+absCache)
```
Despire: Don't copy all of the envs into the command, only the set the `TF_PLUGIN_CACHE_DIR`

services/terraformer/server/domain/utils/tofu/terraform.go
```diff
@@ -37,17 +42,49 @@ type Terraform struct {
...
+
+	//nolint
+	cmd := exec.Command("tofu", args...)
+	cmd.Dir = t.Directory
+
+	if err := cmd.Run(); err != nil {
+		return err
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
m-brando: There won't be conlict, bacasue fs-mirror path is not used as directory for binaries in our case. It si just used for lock file has calculation. 
only when run `tofu init` command we populate TF_PLUGIN_CACHE_DIR dir.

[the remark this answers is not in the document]
m-brando: @coderabbitai the commnad `tofu providers lock -fs-mirror=` direct OpenTofu to look for provider packages in the given local filesystem mirror directory, instead of in upstream registries.  It does not write or do any modification in directory, hence safe operation.

internal/command/cmd.go
```diff
@@ -185,6 +186,7 @@ func (c *Cmd) buildCmd() (*exec.Cmd, context.CancelFunc) {
 	cmd.Dir = c.Dir
 	cmd.Stdout = c.Stdout
 	cmd.Stderr = c.Stderr
+	cmd.Env = c.Env
```
Despire: In here mirror the same behavior as in the top level command.

i.e.

```go
cmd.Env = append(cmd.Environ(), c.Env...)
```

The [Cmd.Env] should contain specific environment variables to be set not the `cmd.Environ` of the top level command

services/terraformer/server/domain/utils/tofu/terraform.go
```diff
@@ -82,20 +81,20 @@ func (t *Terraform) Init() error {
...
 	if err := cmd.Run(); err != nil {
 		log.Warn().Msgf("Error encountered while executing %s from %s: %v", cmd, t.Directory, err)
 
-		os.Setenv("TF_PLUGIN_CACHE_DIR", absCache)
 		retryCmd := comm.Cmd{
 			Command: "tofu init",
 			Dir:     t.Directory,
 			Stdout:  cmd.Stdout,
 			Stderr:  cmd.Stderr,
+			Env:     append(cmd.Environ(), fmt.Sprintf("TF_PLUGIN_CACHE_DIR=%s", absCache)),
```
Despire: Here only set 
`[]string{fmt.Sprintf("TF_PLUGIN_CACHE_DIR=%s", absCache)}`

services/kuber/server/domain/usecases/patch_kubeadm.go
```diff
@@ -32,59 +59,95 @@ func (u *Usecases) PatchKubeadmConfigMap(ctx context.Context, request *pb.PatchK
...
+		k := kubectl.Kubectl{
+			Directory:         clusterDir,
+			Kubeconfig:        request.DesiredCluster.Kubeconfig,
+			MaxKubectlRetries: -1,
+		}
+
+		if err = k.KubectlApply(filepath.Base(file.Name()), "-n kube-system"); err != nil {
+			logger.Warn().Msgf("failed to patch kubeadm-config config map: %v", err)
 			continue
 		}
```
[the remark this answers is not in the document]
Despire: While I agree this could lead to resource exhaustion the for loop has an upper limit which will only lead to a maximum of 5 file handles, But nevertheless I'll change it to close it within that iteration and not defer it until the function exists.

services/kuber/server/domain/usecases/patch_kubeadm.go
```diff
@@ -95,11 +161,13 @@ func (u *Usecases) PatchKubeadmConfigMap(ctx context.Context, request *pb.PatchK
...
 	apiServer, ok := rawKubeadmConfig["apiServer"].(map[string]any)
 	if !ok {
-		logger.Warn().Msgf("Expected 'apiServer' field to be present in the 'ClusterConfiguration' for kubeadm, but was not")
 		return nil, fmt.Errorf("expected 'apiServer' field to present in the 'ClusterConfiguration' for kubeadm, but was missing: %w", err)
 	}
 
 	apiServer["certSANs"] = certSANs
+	if _, ok := rawKubeadmConfig["controlPlaneEndpoint"]; ok {
+		rawKubeadmConfig["controlPlaneEndpoint"] = net.JoinHostPort(certSANs[0], fmt.Sprint(manifest.APIServerPort))
+	}
```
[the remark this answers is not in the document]
Despire: this can not occur, there must be either a load balancer or a control node.

Issue #1840, opened 2025-08-25
Feature: NVIDIA operator containerd `config.toml`


Despire:
### Motivation

Make Nvidia operator work well with Claudie.

### Description

As of https://github.com/berops/claudie/issues/1786 Claudie includes a restart for the Nvidia toolkit daemonset that overwrites the `config.toml` on each node to use Nvidia as the default runtime, however. On each worklow the `config.toml` is overwritten by kubeone. During the time kubeone is running, newly spawned or restarted pods, the nodes will not have the correct `config.toml` and thus fail to run possibly crashlooping. Another scenario is that kubeone ends with an error and the fix in kuber will not be executed thus further delaying the restart of the Nvidia toolkit to revert the changes made by kubeone.

We need to make sure that the Nvidia operator is ingrated into claudie really well and that scenarios, such as the above described, do not happen.

### Exit criteria

- [ ] Come up with possible implementation
- [ ] Implement

bernardhalas:
4 options considered, design doc might be useful here as well:
- propose a different behavior in KubeOne
- migrate from KubeOne to kubeadm
- explore if `ebpf` lock on the `config.toml` would be a good approach
- use custom Claudie images (not a full solution, because this is not applicable to all scenarios - e.g. CloudRift or all on-prem)

PR #1916, opened 2025-12-03
Prevent `config.toml` override by kubeone

jakubhlavacka:
This PR creates a `tee` file in `/opt/claudie/bin` to prevent `kubeone` from overriding `config.toml` after the [nvidia-container-toolkit](https://github.com/NVIDIA/nvidia-container-toolkit) (part of the [gpu-operator](https://github.com/NVIDIA/gpu-operator)) updated this file to allow containers to use the GPU device.

The `/opt/claudie/tee` overrides the `/usr/bin/tee`. Suppose someone/something uses the `tee` command to write to `/etc/containerd/config.toml`. The overridden `tee` will discard the write and return a success to a parent process if there's a `/etc/containerd/conf.d` directory with a `*nvidia*` file. Otherwise, the `tee` command works as expected.

In order for the script to work, the `/opt/claudie/bin` has to be before `/usr/bin` in the PATH. Claudie handles this.

I have performed the following tests. The first was to ensure the `tee` override works.
1. Provisioned a downstream cluster with a GPU instance
2. Deployed GPU operator
3. Upscaled the downstream cluster
4. Deployed a Python application with `bash` on a GPU instance
5. Executed to the Python application and checked the `CUDA` utilization by running `torch.cuda.is_available()`
6. The command from the previous step returned `True`

The second one verified that the `/opt/claudie/tee` is cleaned from the static nodes.
1. Provisioned a downstream cluster with a static node
2. The static node had the `/opt/claudie/tee`
3. Removed the static node from the downstream cluster
4. Ensure the `/opt/claudie/tee` was successfully removed

You can read more in the [design doc](https://docs.google.com/document/d/13CCn-ff89YMNeTb2tXraBPj3_HwmEs2rBomaDSWD-8c/edit?usp=sharing)

closes https://github.com/berops/claudie/issues/1840

## Summary by CodeRabbit

* **New Features**
  * New API action to install a tee binary override across cluster nodes; integrated into infrastructure workflow.
  * New Ansible playbook that conditionally intercepts writes to containerd config when NVIDIA-related configs are present.

* **Removals**
  * Optional NVIDIA GPU operator rollout step removed from update and reconciliation workflows.

* **Chores**
  * Uninstall now removes tee override and PATH augmentation; added completion log, updated deployment image tags, and .gitignore.

### DIFF
--- a/manifests/claudie/kustomization.yaml
+++ b/manifests/claudie/kustomization.yaml
@@ -56,18 +56,18 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: ghcr.io/berops/claudie/ansibler
-  newTag: 87f38c9-3767
+  newTag: 322cee7-3768
 - name: ghcr.io/berops/claudie/autoscaler-adapter
-  newTag: 87f38c9-3767
+  newTag: 322cee7-3768
 - name: ghcr.io/berops/claudie/builder
-  newTag: 87f38c9-3767
+  newTag: 322cee7-3768
 - name: ghcr.io/berops/claudie/claudie-operator
-  newTag: 87f38c9-3767
+  newTag: 322cee7-3768
 - name: ghcr.io/berops/claudie/kube-eleven
-  newTag: 87f38c9-3767
+  newTag: 322cee7-3768
 - name: ghcr.io/berops/claudie/kuber
-  newTag: 87f38c9-3767
+  newTag: 322cee7-3768
 - name: ghcr.io/berops/claudie/manager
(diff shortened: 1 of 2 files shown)


proto/ansibler.proto
```diff
@@ -81,6 +81,14 @@ message RemoveClaudieUtilitiesRequest {
 
 message RemoveClaudieUtilitiesResponse {}
 
+message InstallTeeOverrideRequest {
+  spec.K8scluster current = 1;
```
Despire: Don't send current state, its not used in the function.

Issue #1831, opened 2025-08-15
Chore: Speed up Ansible stage


Despire:
### Description

Currently the Ansible stage is the main bottleneck in terms of how much time is spend when building larger clusters ~150+ nodes.

### Exit criteria

- [ ] Find ways to optimize Ansible stage to speed up building larger clusters.

PR #1917, opened 2025-12-05
Feat/ansible perf

m-brando:
This PR addresses issue #1831.

Improvements:
- Removed unnecessary fact gathering.
- Replaced loops with blockinfile where applicable.
- Introduced `async` with `poll: 0` to run non-blocking tasks, combined with later state checks using `async_status`.
- Reorganized tasks and play structure for more efficient execution.

overall speedup of execution is ~20%

The WireGuard playbook remains the slowest part of the workflow. It could be parallelized further, but doing so would require synchronizing host facts, particularly the public key values used when rendering the WireGuard template in this task:

```
- name: Template wireguard config file to a node
  ansible.builtin.template:
    src: "{{ 'wg-static.conf.j2' if 'static' in group_names else 'wg-dynamic.conf.j2' }}"
    dest: "{{ wg_conf_dir.path }}/wg0.conf"
```

## Summary by CodeRabbit

* **Performance Improvements**
  * More background package operations and reduced fact gathering to speed playbook runs.

* **Stability & Reliability**
  * Consolidated kernel, sysctl and systemd limits with a single reload for consistent node limits.
  * Centralized proxy environment management via block updates and standardized drop-in dirs.
  * Safer WireGuard key handling and idempotent interface/IP management; unattended-upgrades now masked.

* **Refactor**
  * Merged and simplified related tasks for cleaner, more consistent playbooks.

* **Chores**
  * Updated ansibler image tag.

### DIFF
--- a/docs/contributing/local-testing.md
+++ b/docs/contributing/local-testing.md
@@ -110,7 +110,7 @@ nodePools:
         region: nbg1
         zone: nbg1-dc3
       count: 1
-      serverType: cpx11
+      serverType: cpx22
       image: ubuntu-22.04
       storageDiskSize: 50
 
@@ -120,7 +120,7 @@ nodePools:
         region: nbg1
         zone: nbg1-dc3
       count: 1
-      serverType: cpx11
+      serverType: cpx22
       image: ubuntu-22.04
 
   static:
--- a/docs/getting-started/detailed-guide.md
+++ b/docs/getting-started/detailed-guide.md
@@ -339,7 +339,7 @@ This detailed guide for Claudie serves as a resource for providing an overview o
                 name: hetzner-1   # use your new hetzner provider hetzner-1 to create these nodes
                 region: hel1
                 zone: hel1-dc2
-            serverType: cpx51
+            serverType: cpx52
             image: ubuntu-22.04
             autoscaler:           # this node pool uses a claudie autoscaler instead of static count of nodes
                 min: 1
--- a/docs/input-manifest/example.md
+++ b/docs/input-manifest/example.md
@@ -108,7 +108,7 @@ spec:
           region: hel1
           zone: hel1-dc2
         count: 3
-        serverType: cpx11
+        serverType: cpx22
         image: ubuntu-24.04
         labels:
           country: finland
@@ -126,7 +126,7 @@ spec:
           region: hel1
           zone: hel1-dc2
         count: 2
-        serverType: cpx11
+        serverType: cpx22
         image: ubuntu-24.04
         storageDiskSize: 50
         labels:
@@ -140,7 +140,7 @@ spec:
           name: hetzner-1
           region: hel1
           zone: hel1-dc2
-        serverType: cpx11
+        serverType: cpx22
         image: ubuntu-24.04
         storageDiskSize: 50
         autoscaler:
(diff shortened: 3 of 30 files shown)


services/ansibler/server/ansible-playbooks/longhorn-req.yml
```diff
@@ -61,3 +46,11 @@
...
     - name: update node limits
       include_tasks: ./node-limits.yml
+
+    - name: Check async task result
+      async_status:
+        jid: "{{ async_job.ansible_job_id }}"
+      register: result
+      until: result.finished
+      retries: 10
+      delay: 5
```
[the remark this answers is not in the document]
Despire: I agree with this comment, increase the wait time for the async task here.

services/ansibler/server/ansible-playbooks/node-limits.yml
```diff
@@ -10,76 +10,44 @@
...
-    state: present
-    reload: yes
-
-- name: Set TCP fin timeout to 30s
-  ansible.builtin.sysctl:
-    name: net.ipv4.tcp_fin_timeout
-    value: "30"
-    state: present
-    reload: yes
+- name: Configure kernel parameters
```
Despire, line 35: I think this is okay, but won't this essentially depend on the file namings and which one will be loaded last ?

services/ansibler/server/ansible-playbooks/proxy/populate-proxy-envs.yml
```diff
@@ -1,25 +1,25 @@
 - hosts: all
+  strategy: free
   gather_facts: false
   become: true
   tasks:
-    - name: Verifying if /etc/apt/apt.conf.d/ exists
+    - name: Ensure apt and systemd drop-in directories exist
```
Despire, line 6: Why delete the check for `/etc/apt/apt.conf.d/` ?

services/ansibler/server/ansible-playbooks/node-limits.yml
m-brando: yes it depends that's why I put prefix 99- which effectively override previous defined params if any https://unix.stackexchange.com/questions/570770/what-does-the-number-in-sysctl-configuration-file-denote

services/ansibler/server/ansible-playbooks/proxy/populate-proxy-envs.yml
m-brando: man apt.conf
```
FILES
       /etc/apt/apt.conf
           APT configuration file. Configuration Item: Dir::Etc::Main.

       /etc/apt/apt.conf.d/
           APT configuration file fragments. Configuration Item: Dir::Etc::Parts.
```
This means that on a Debian-based distribution, the directory already exists, so there’s no need to check for it. Skipping that step will save a few seconds in the Ansible task.

services/ansibler/server/ansible-playbooks/longhorn-req.yml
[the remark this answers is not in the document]
m-brando: fixed

services/ansibler/server/ansible-playbooks/wireguard/tasks/main.yml
```diff
@@ -1,8 +1,8 @@
 ---
-- include_tasks: kill_unattended_upgrades.yml
+- include_tasks: install.yml
   tags: install
 
-- include_tasks: install.yml
+- include_tasks: kill_unattended_upgrades.yml
   tags: install
```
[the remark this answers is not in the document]
m-brando: fixed

PR #1952, opened 2026-01-27
feat(gcp): add GPU guest accelerator support for GCP nodepools

### LINKED ISSUE #1845: Feature: Enable GPU acceleration for instances where instance-type does not automatically enable the GPU's
### Motivation

Certain Cloud Providers like GCP and Azure, i.e. [guest accelerator](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance#argument-reference) for GCP

### Description

For Providers where the instance-type does not automatically make the GPU available, we should add an option into the nodepool and support into the templates themselves.

We already have something in place [Machine Spec](https://github.com/berops/claudie/blob/64a152b2728ba35f741d8e2dcbd85e88e8e12cfe/manifests/testing-framework/test-sets/test-set3/1.yaml#L70C9-L70C20), that is used only for a few cloud providers where more custom configuration is needed.

### Exit criteria

- [ ] Make GPU instances work with all supported providers

samuelstolicny:
## Summary

Closes #1845

This PR adds GPU guest accelerator support for GCP nodepools, allowing users to attach NVIDIA GPUs to GCP compute instances via Claudie's InputManifest.

### Changes

- **Proto**: Added `nvidiaGpuCount` (renamed from `nvidiaGpu`) and `nvidiaGpuType` fields to `MachineSpec`
- **Manifest**: Updated `MachineSpec` struct with new GPU fields and backward compatibility for `nvidiaGpu`
- **Validation**: Added `validateGCPGpuConfig()` to ensure GCP nodepools with GPUs have the required type specified
- **CRD**: Regenerated with new `nvidiaGpuCount` and `nvidiaGpuType` fields
- **Autoscaler**: Updated to use `NvidiaGpuCount` field name
- **Documentation**: Added GPU configuration docs for GCP provider
- **Makefile**: Added `kind-deploy` target for easier local testing

### Example Usage

```yaml
nodePools:
  dynamic:
    - name: gpu-workers
      providerSpec:
        name: gcp-provider
        region: europe-west1
        zone: europe-west1-b
      count: 1
      serverType: n1-standard-4
      image: ubuntu-2204-lts
      machineSpec:
        nvidiaGpuCount: 1
        nvidiaGpuType: nvidia-tesla-t4
```

### Backward Compatibility

- The old `nvidiaGpu` field is preserved as a deprecated alias for `nvidiaGpuCount`
- Non-GCP providers can still use GPU count without specifying type
- Proto field numbers maintained for wire compatibility

### Template Changes

GCP template changes implemented in https://github.com/berops/claudie-config/pull/22

## Summary by CodeRabbit

* **New Features**
  - GCP now supports NVIDIA GPUs with configurable GPU count and explicit GPU type; deprecated GPU field retained for compatibility.
  - Local "kind" cluster deploy target added to update and rollout workloads.

* **Documentation**
  - Added GCP GPU docs, examples, provider guidance and a GPU Operator deployment guide; updated provider matrix.

* **Validation**
  - GCP-specific validation requires GPU type when GPUs are requested.

* **Tests**
  - Added validation tests covering GCP and non‑GCP GPU scenarios.

### DIFF
--- a/manifests/claudie/kustomization.yaml
+++ b/manifests/claudie/kustomization.yaml
@@ -56,18 +56,18 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: ghcr.io/berops/claudie/ansibler
-  newTag: 1ab052c-3865
+  newTag: 54b543c-3870
 - name: ghcr.io/berops/claudie/autoscaler-adapter
-  newTag: 1ab052c-3865
+  newTag: 54b543c-3870
 - name: ghcr.io/berops/claudie/builder
-  newTag: 1ab052c-3865
+  newTag: 54b543c-3870
 - name: ghcr.io/berops/claudie/claudie-operator
-  newTag: 1ab052c-3865
+  newTag: 54b543c-3870
 - name: ghcr.io/berops/claudie/kube-eleven
-  newTag: 1ab052c-3865
+  newTag: 54b543c-3870
 - name: ghcr.io/berops/claudie/kuber
-  newTag: 1ab052c-3865
+  newTag: 54b543c-3870
 - name: ghcr.io/berops/claudie/manager
-  newTag: 1ab052c-3865
+  newTag: 54b543c-3870
 - name: ghcr.io/berops/claudie/terraformer
-  newTag: 1ab052c-3865
+  newTag: 54b543c-3870
--- a/manifests/testing-framework/kustomization.yaml
+++ b/manifests/testing-framework/kustomization.yaml
@@ -90,4 +90,4 @@ secretGenerator:
 
 images:
 - name: ghcr.io/berops/claudie/testing-framework
-  newTag: 1ab052c-3865
+  newTag: 54b543c-3870


docs/input-manifest/providers/gcp.md
```diff
@@ -69,6 +69,53 @@ If you wish to use GCP as your DNS provider where Claudie creates DNS records po
...
+| `nvidia-tesla-p100` | NVIDIA Tesla P100 |
+| `nvidia-tesla-k80` | NVIDIA Tesla K80 (legacy) |
+
+!!! note "GPU Availability"
+    GPU availability varies by zone. Check [GCP GPU regions and zones](https://cloud.google.com/compute/docs/gpus/gpu-regions-zones) for current availability in your desired region.
+
+!!! warning "GPU Instance Limitations"
+    - GPU instances cannot be live migrated and will be terminated during maintenance events
+    - Use `n1-standard-*` or `n1-highmem-*` machine types with GPUs (not `e2-*` types)
+    - Some GPU types have minimum vCPU and memory requirements
```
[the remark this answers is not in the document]
samuelstolicny: irrelevant

internal/api/manifest/manifest.go
```diff
@@ -135,9 +135,17 @@ type MachineSpec struct {
...
 	Memory int `validate:"required_with=CpuCount,gte=0" yaml:"memory" json:"memory"`
-	// Nvidia specifies the number of NVIDIA GPUs the provided instance type will have.
+	// NvidiaGpuCount specifies the number of NVIDIA GPUs the provided instance type will have.
 	// +optional
-	NvidiaGpu int `validate:"gte=0" yaml:"nvidiaGpu" json:"nvidiaGpu"`
+	NvidiaGpuCount int `validate:"gte=0" yaml:"nvidiaGpuCount" json:"nvidiaGpuCount"`
+	// NvidiaGpuType specifies the NVIDIA GPU accelerator type (required for GCP when using GPUs).
+	// Examples: nvidia-tesla-k80, nvidia-tesla-v100, nvidia-tesla-a100, nvidia-l4
+	// +optional
+	NvidiaGpuType string `validate:"omitempty" yaml:"nvidiaGpuType" json:"nvidiaGpuType,omitempty"`
```
m-brando, line 144: this should not be `validate:"omitempty"`
you should check if customer specify `NvidiaGpuCount` > 0. if so this parameter is required. 

check how we do validation for `ProviderSpec` struc for string `ExternalNetworkName`

internal/api/manifest/validate_node_pool.go
```diff
@@ -137,6 +142,36 @@ func (d *DynamicNodePool) Validate(m *Manifest) error {
 	return nil
 }
 
+// validateGCPGpuConfig validates that GCP nodepools with GPUs have the required type specified.
+// GCP requires the guest_accelerator block with both type and count to attach GPUs to instances.
+func (d *DynamicNodePool) validateGCPGpuConfig(m *Manifest) error {
+	providerType, err := m.GetProviderType(d.ProviderSpec.Name)
+	if err != nil {
+		// Provider validation is handled elsewhere
+		return nil
```
m-brando, line 152: Where is this validation handled?
I think we should `return err`, otherwise this error will be supressed and the process will continues with validation of external net on line 135

internal/api/manifest/manifest.go
samuelstolicny: @m-brando thats not really like that, when a user uses AWS provider, they need to specify only NvidiaGpuCount (deprecated nvidiaGpu). NvidiaGpuType is GCP-specific, not universal.
For AWS: Users specify NvidiaGpuCount only (deprecated nvidiaGpu) for AWS GPU instance types (like p3.2xlarge, g4dn.xlarge) already include the GPU hardware built into the instance type (if they need more). No separate GPU type specification is needed. https://docs.claudie.io/v0.9.15/input-manifest/api-reference/#:~:text=nvidiaGpu%3A%20specifies%20the%20number%20of%20nvidia%20GPUs%20used%20by%20the%20serverType
For GCP: Users need both NvidiaGpuCount and NvidiaGpuType because GCP uses a guest_accelerator mechanism that requires explicit GPU attachment.

proto/spec/nodepool.proto
```diff
@@ -91,9 +91,14 @@ message DynamicNodePool {
 
 // MachineSpec further specifies the requested server type.
 message MachineSpec {
+  // Number of CPU cores.
   int32 cpuCount = 1;
+  // Memory in MB.
   int32 memory = 2;
-  int32 nvidiaGpu = 3;
```
m-brando, line 96: I think we should keep the `nvidiaGpu` field at index 3 and mark it as `[deprecated = true]`. All new parameters should start from index 4.

https://medium.com/@sanhdoan/protocol-buffers-the-importance-of-reserved-and-deprecated-fields-bbb8d7d3211d
what do you think @Despire ?

internal/api/manifest/manifest.go
m-brando: understood, but we can still check if provider is GCP and, if so, than we should also validate if `NvidiaGpuType` is present, otherwise if we don't check this field, it can happend, user could omit `NvidiaGpuType` parameter (and only specify `NvidiaGpuCount`) and the execution would fail later, probably in terraformer stage.

This can be avoided by validating the manifest upfront and returning an error before any execution starts.

samuelstolicny: This validation already exists in the PR, see validateGCPGpuConfig() in validate_node_pool.go https://github.com/berops/claudie/blob/7f0110968bd42c1a419b8c0f05e3f3412fb608f0/internal/api/manifest/validate_node_pool.go#L147 which is called in DynamicNodePool.Validate() which runs during manifest vlidation. So users get an instant error if they specify NvidiaGpuCount without NvidiaGpuType for GCP before any terraformer execution. Can I even resolve provider type in validation tag?

proto/spec/nodepool.proto
samuelstolicny: Not sure what is the best practice here but since I only renamed the field (same number, same type), the binary format stays compatible. The Go struct already has both fields for backward compatibility.

internal/api/manifest/validate_node_pool.go
samuelstolicny: Kept return nil, returning err breaks existing unit tests that bypass the full validation flow. Provider existence is already validated at lines 28-31 before this runs. I want to avoid making the current unit tests more complicated just for this to pass. I think we are safe.

proto/spec/nodepool.proto
Despire: This is fine, both fields have the same meaning. And the deprecated one will not be used

internal/api/manifest/validate_node_pool.go
Despire: Agree, this cannot error out here.

internal/api/manifest/manifest.go
Despire: This is okay, validation will be checked for GCP on later stages. 👍

Despire: Not sure you would be able to resolve providerType at the `tag` level, at least not without some ugly "workarounds"

Despire (approved):
LGTM 👍 

Please also replace the old field here: https://github.com/berops/claudie/blob/master/docs/autoscaling/autoscaling.md

After we can merge.

PR #2066, opened 2026-04-21
fix: apply NAT hairpin workaround to OpenStack providers

### LINKED ISSUE #2064: Bug: extend NAT hairpin fix to OpenStack providers (Cleura)
## What's happening

When two VMs in the same OpenStack (Cleura) project talk to each other using their floating IPs, the traffic hairpins through Neutron's L3 router. The router rewrites the source port in the process. WireGuard remembers whatever source port it last saw from a peer, so it ends up sending keepalives to a random port that stops working a couple minutes later once the NAT entry ages out. The tunnel then sits in a constant re-handshake loop.

Cross-cloud traffic (OpenStack to AWS or the reverse) is fine. Only same-project peers reaching each other via floating IPs hit this.

Same class of problem as Cloudrift in #1982, and the existing fix there works here with a small change.

## How to reproducede

Tested on Cleura Cloud (Fra1):

- 1 Cleura control + 1 AWS control + 1 Cleura compute: cluster builds fine. Only one intra-Cleura tunnel, the churn is tolerable.
- 2 AWS control + 1 Cleura control + 3 Cleura compute + 1 AWS compute: kubeone gives up after 22 minutes of retries on the Cleura control's apiserver, because etcd can't stabilize across the several intra-Cleura tunnels that are all churning at once.

`wg show` on a Cleura node early in the build:

```
peer cleura-cmp
  endpoint: 185.52.156.127:63959    (should be :51820)
```

A few minutes later, same peer, same cluster:

```
peer cleura-cmp
  endpoint: 185.52.156.127:2807
```

New random port on every handshake. Keepalive counters show the classic "sending into a black hole" pattern: about 20 KiB sent, ~200 bytes received over 15 minutes, handshake last reached several minutes ago despite `PersistentKeepalive = 30`.

## Why it happens

Neutron's L3 agent runs iptables SNAT on traffic that hairpins between two floating IPs in the same project. Under some combinations of hardening (`--random-fully`) and plain conntrack collision fallback, the source port ends up randomized. WireGuard has an "endpoint roaming" feature that auto-updates the cached peer address to whatever source IP and port the last packet arrived from, so wg latches onto the randomized port. That port is an ephemeral NAT mapping that times out, and keepalives to it silently get dropped.

## The fix

Cloudrift already has a pattern for this:

- an Ansible playbook that adds `iptables -t nat -A OUTPUT -d <peer_public_ip> -j DNAT --to <peer_private_ip>` on each node, for every peer in the same `/24`
- a Go filter that runs the playbook only for Cloudrift nodepools

The playbook itself is already provider-agnostic (the `/24` match is what auto-scopes it to same-cloud peers). We just need to:

1. Rename `services/ansibler/ansible-playbooks/cloudrift-nat-fix.yml` to `nat-hairpin-fix.yml`. Contents unchanged.
2. In `services/ansibler/internal/worker/service/task_install_vpn.go`:
   - rename `fixCloudRiftNAT` to `fixHairpinNAT`,
   - swap the hardcoded `"cloudrift"` check for a set `{"cloudrift", "openstack"}`,
   - update the playbook constant and the call site in `installWireguardVPN`.

No changes to `claudie-config`. Self-hairpinning (a VM reaching its own public IP) already works on OpenStack out of the box, so the startup-script rule Cloudrift uses for that case is not needed here.

## What to verify after the fix

- [ ] Two or more OpenStack peers in the same project keep stable WireGuard tunnels with endpoints at `:51820` on both sides after bootstrap.
- [ ] Multi-node cluster with control plane split across OpenStack and another provider builds reliably. The failing 7-node topology above should pass.
- [ ] Cross-cloud tunnels (OpenStack ↔ AWS etc.) keep working unchanged.
- [ ] Cloudrift-only clusters are unaffected. The renamed playbook covers them identically.

## Related

- Inspired by #1982 (Cloudrift NAT hairpin fix)
- Discovered while working on #1900 (OpenStack provider tutorials)

samuelstolicny:
Closes #2064

## What this does

Extends the existing Cloudrift NAT hairpin fix to also cover OpenStack, so that WireGuard tunnels between OpenStack nodes in the same project stay stable instead of churning under Neutron's source-port randomization.

## How

- Renames `services/ansibler/ansible-playbooks/cloudrift-nat-fix.yml` to `nat-hairpin-fix.yml`.
- Adds a `file: state=directory` step to ensure `/etc/iptables` exists before `iptables-save` writes to it. Some OpenStack images (for example Cleura's Ubuntu 24.04) don't ship the directory.
- Renames `fixCloudRiftNAT` to `fixHairpinNAT` in `services/ansibler/internal/worker/service/task_install_vpn.go` and gates it on a provider set `{cloudrift, openstack}`.
- Matches peers by a new `hairpin_group` host variable with the shape `specName + clusterID + region`. DNAT rules are installed only between peers that share the same group, which is also the same private Neutron network in Claudie's OpenStack template. `region` is part of the key because one OpenStack provider can span multiple regions, each on its own private network with the same `10.0.0.0/16` CIDR, and cross-region LAN IPs are not mutually reachable.
- Adds a dedicated inventory template `nat-hairpin-inventory.goini` that emits the `hairpin_group` host var.
- Counts nodes per `hairpin_group` in the early-exit check, so the playbook runs only when at least one group has a peer pair.

No changes to `claudie-config` templates or any other service.

## Verification

Validated against two OpenStack providers (Cleura Fra1 and OVH GRA9 + SBG5) plus AWS, across four builds:

### 1. Cleura + AWS (7 nodes)

1 Cleura control + 2 AWS control + 3 Cleura compute + 1 AWS compute. Cleura floating IPs came from two `/24` pools in the same project (`89.46.80.0/24` for control, `91.123.203.0/24` for computes).

- Cluster reaches `DONE` in ~10 minutes, all 7 nodes `Ready`.
- `fixHairpinNAT` installs 3 DNAT rules on every Cleura node, including the control node that landed in a different `/24` from the computes.
- All 6 WireGuard peer endpoints on the Cleura control stay at `:51820` both at steady state and under 5 minutes of cross-node pod-to-pod traffic (DaemonSet nginx + busybox client loop). No randomized ports.
- 49 pod-to-pod curls from a client pod on the Cleura control to all 7 nginx pods succeed (100%).

### 2. OVH (2 nodes)

1 OVH control + 1 OVH compute in GRA9. Single-provider smoke test.

- Cluster reaches `DONE`, both nodes `Ready`.
- Bidirectional DNAT mesh installed (1 rule per node).
- WG peer endpoint stable at `:51820`, handshakes under 10 s.

### 3. OVH + Cleura + AWS (8 nodes, tri-cloud)

1 OVH control + 1 Cleura control + 1 AWS control + 4 OVH compute + 1 Cleura compute. OVH floating IPs came from two disjoint `/24` pools (`57.128.55.0/24` for `ovh-cmp-01`, `145.239.127.0/24` for the other 4 OVH nodes); OVH's `Ext-Net` in GRA9 has 62 subnets, so cross-pool allocation is common.

- Cluster reaches `DONE`, all 8 nodes `Ready`.
- Two independent DNAT meshes built in the same ansibler run, keyed by `hairpin_group`:
  - 5-node OVH mesh (4 rules per node)
  - 2-node Cleura mesh (1 rule per node)
- On `ovh-cmp-01` (the cross-pool node at `57.128.55.13`), all 4 DNAT rules to the `145.239.127.x` peers are installed.
- All 7 WG peer endpoints on `ovh-cmp-01` are at `:51820`, handshakes under 1 min.
- AWS node sees no DNAT rules, as expected.

### 4. OVH cross-region (4 nodes, GRA9 + SBG5)

One `ovh-1` provider, 1 control + 1 compute in GRA9, 2 compute in SBG5. Same provider, two regions, each with its own private `10.0.0.0/16` network.

- Cluster reaches `DONE`, all 4 nodes `Ready`.
- Playbook installs exactly 4 DNAT rule changes total, one per same-region peer direction.
- `iptables -t nat -S OUTPUT` on the GRA9 control shows exactly 1 DNAT rule (to the GRA9 compute's LAN IP) and zero rules for the two SBG5 peers.
- Same pattern on an SBG5 compute: 1 DNAT rule to the other SBG5 compute, zero for GRA9 peers.
- Cross-region WG peers use floating IPs as expected; the fix deliberately leaves cross-region traffic alone.

Cloudrift code path is untouched beyond the rename; the same playbook now serves three providers with provider-identity plus region scoping.

## Risk / rollback

- Cloudrift clusters are affected only by the rename plus the `/etc/iptables` mkdir step. The mkdir is idempotent and the file write was already conditional on the directory existing on Cloudrift images, so behavior there is unchanged.
- For OpenStack providers without the hairpin bug, the DNAT rules route same-provider peer traffic over the LAN instead of through the floating IP. Both paths work, so at worst this is a minor routing optimisation; no correctness regression expected.
- To revert, drop this commit - no migrations or state changes.

## Summary by CodeRabbit

* **Improvements**
  * Generalized NAT hairpin fix that targets providers by provider/cluster/region groups and runs only where groups have 2+ nodes.
  * More accurate matching so DNAT rules apply only between hosts in the same hairpin group.
  * Added a generated inventory for hairpin-targeted hosts.

* **Bug Fixes**
  * Ensure firewall rules directory exists before saving iptables rules.

* **Chores**
  * Updated deployment image tags used when rendering manifests.

### DIFF
--- a/services/ansibler/internal/worker/service/task_install_vpn.go
+++ b/services/ansibler/internal/worker/service/task_install_vpn.go
@@ -20,9 +20,19 @@ import (
 )
 
 const (
-	wireguardPlaybook       = "../../ansible-playbooks/wireguard.yml"
-	cloudriftNATFixPlaybook = "../../ansible-playbooks/cloudrift-nat-fix.yml"
+	wireguardPlaybook     = "../../ansible-playbooks/wireguard.yml"
+	natHairpinFixPlaybook = "../../ansible-playbooks/nat-hairpin-fix.yml"
 )
+
+// hairpinNATProviders lists cloud providers whose same-project, floating-IP
+// traffic hairpins through a NAT gateway that can rewrite the source port,
+// breaking WireGuard tunnels between peers on the same network. The fix
+// (see fixHairpinNAT) installs local DNAT rules so peers reach each other
+// over private IPs instead of the public hairpin path.
+var hairpinNATProviders = map[string]struct{}{
+	"cloudrift": {},
+	"openstack": {},
+}
 
 type VPNInfo struct {
 	ClusterNetwork string
@@ -164,15 +174,15 @@ func installWireguardVPN(clusterID string, vpnInfo *VPNInfo, processLimit *semap
 		}
 	}
 
-	// Fix CloudRift NAT before WireGuard so that peer public IPs are
+	// Fix hairpin NAT before WireGuard so that peer public IPs are
 	// redirected to LAN IPs, allowing the WireGuard tunnel to establish
 	// and kubeadm join to reach the API server.
-	if err := fixCloudRiftNAT(vpnInfo, clusterID, clusterDirectory, processLimit); err != nil {
-		return fmt.Errorf("error while fixing CloudRift NAT for %s : %w", clusterDirectory, err)
+	if err := fixHairpinNAT(vpnInfo, clusterID, clusterDirectory, processLimit); err != nil {
+		return fmt.Errorf("error while fixing hairpin NAT for %s : %w", clusterDirectory, err)
 	}
 
-	// Regenerate the full inventory since fixCloudRiftNAT overwrites it
-	// with CloudRift-only nodes.
+	// Regenerate the full inventory since fixHairpinNAT overwrites it
+	// with hairpin-provider-only nodes.
 	if err := utils.GenerateInventoryFile(templates.AllNodesInventoryTemplate, clusterDirectory, data); err != nil {
 		return fmt.Errorf("error while recreating inventory file for %s : %w", clusterDirectory, err)
 	}
@@ -191,24 +201,30 @@ func installWireguardVPN(clusterID string, vpnInfo *VPNInfo, processLimit *semap
 	return nil
 }
 
-// fixCloudRiftNAT adds iptables OUTPUT DNAT rules on CloudRift nodes so that
-// peer public IPs are redirected to their LAN IPs (discovered via Ansible
-// gather_facts). This works around CloudRift's local gateway sending TCP RST
-// instead of hairpinning traffic between VMs on the same subnet. Must run
+// fixHairpinNAT adds iptables OUTPUT DNAT rules on nodes of providers whose
+// floating-IP hairpin path can rewrite the source port (see hairpinNATProviders),
+// so that peer public IPs are redirected to their LAN IPs (discovered via Ansible
+// gather_facts). This works around provider-side NAT behavior (CloudRift's local
+// gateway sending TCP RST, OpenStack Neutron's SNAT port randomization) that
+// otherwise breaks WireGuard tunnels between VMs on the same subnet. Must run
 // before WireGuard so that the tunnel endpoints resolve over the LAN, and
 // before KubeEleven so that kubeadm join can reach the API server.
-func fixCloudRiftNAT(vpnInfo *VPNInfo, clusterID, clusterDirectory string, processLimit *semaphore.Weighted) error {
-	var cloudriftInfos []*NodepoolsInfo
+func fixHairpinNAT(vpnInfo *VPNInfo, clusterID, clusterDirectory string, processLimit *semaphore.Weighted) error {
+	var hairpinInfos []*NodepoolsInfo
 
 	for _, npi := range vpnInfo.NodepoolsInfos {
 		var dynPools []*spec.NodePool
 		for _, np := range npi.Nodepools.Dynamic {
-			if dyn := np.GetDynamicNodePool(); dyn != nil && dyn.Provider.CloudProviderName == "cloudrift" {
+			dyn := np.GetDynamicNodePool()
+			if dyn == nil {
+				continue
+			}
+			if _, ok := hairpinNATProviders[dyn.Provider.CloudProviderName]; ok {
 				dynPools = append(dynPools, np)
 			}
 		}
 		if len(dynPools) > 0 {
-			cloudriftInfos = append(cloudriftInfos, &NodepoolsInfo{
+			hairpinInfos = append(hairpinInfos, &NodepoolsInfo{
 				Nodepools:      NodePools{Dynamic: dynPools},
 				ClusterID:      npi.ClusterID,
 				ClusterNetwork: npi.ClusterNetwork,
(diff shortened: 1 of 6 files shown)


services/ansibler/templates/nat-hairpin-inventory.goini
```diff
@@ -0,0 +1,9 @@
+[dynamic]
+{{- range $nodepoolInfo := .NodepoolsInfo }}
+    {{- range $nodepool := $nodepoolInfo.Nodepools.Dynamic }}
+        {{- $dyn := $nodepool.GetDynamicNodePool }}
+        {{- range $node :=  $nodepool.Nodes }}
+{{ trimPrefix (printf "%s-" $nodepoolInfo.ClusterID) $node.Name }} ansible_user=root ansible_host={{ $node.Public }} ansible_port={{ sshPort $nodepool }} private_ip={{ $node.Private }} netmask={{ extractNetmaskFromCIDR $nodepoolInfo.ClusterNetwork }} ansible_ssh_private_key_file={{ $nodepool.Name }}.pem ansible_ssh_extra_args="-o IdentitiesOnly=yes" hairpin_group={{ $dyn.Provider.SpecName }}__{{ $nodepoolInfo.ClusterID }}
```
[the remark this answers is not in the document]
samuelstolicny: Fixed in f1fbab4b. Added Region to the hairpin_group key in both the inventory template and the Go count helper. Kept Zone and ExternalNetworkName out since Claudie's OpenStack template puts all same-region nodepools (per specName per cluster-type) on one private Neutron network, so neither field affects whether a peer's LAN IP is reachable. K8s vs LB clusters are already separated via different ClusterIDs.

services/ansibler/internal/worker/service/task_install_vpn.go
```diff
@@ -191,56 +201,76 @@ func installWireguardVPN(clusterID string, vpnInfo *VPNInfo, processLimit *semap
...
+	var hairpinInfos []*NodepoolsInfo
 
 	for _, npi := range vpnInfo.NodepoolsInfos {
 		var dynPools []*spec.NodePool
 		for _, np := range npi.Nodepools.Dynamic {
-			if dyn := np.GetDynamicNodePool(); dyn != nil && dyn.Provider.CloudProviderName == "cloudrift" {
+			dyn := np.GetDynamicNodePool()
+			if dyn == nil {
+				continue
+			}
```
Despire: You don't have to check for `nil` here, you are already looping through `Dynamic` nodepools you can assume that if there is a nodepool that it is dynamic.

services/ansibler/internal/worker/service/task_install_vpn.go
```diff
@@ -191,56 +201,76 @@ func installWireguardVPN(clusterID string, vpnInfo *VPNInfo, processLimit *semap
...
+	// Count nodes per hairpin group (specName + clusterID + region). DNAT
+	// rules only fire between peers in the same group, so the fix is only
+	// useful if at least one group has 2+ nodes.
+	groupCounts := make(map[string]int)
+	for _, npi := range hairpinInfos {
 		for _, np := range npi.Nodepools.Dynamic {
-			totalNodes += len(np.Nodes)
+			dyn := np.GetDynamicNodePool()
+			if dyn == nil {
+				continue
```
Despire: ditto

Despire (approved):
LGTM

PR #2070, opened 2026-04-23
chore: make taints unique

Despire:
## Summary by CodeRabbit

* **Bug Fixes**
  * Node pool taints are now normalized: deterministically sorted and duplicate entries removed for consistent node configurations.
* **Chores**
  * Updated container image tags for Claudie components and the testing framework to new versions used during deployment.


samuelstolicny (approved):
Looks good

PR #2086, opened 2026-05-12
chore: Switch GitHub Actions authentication to claudie-runner app token

samuelstolicny:
Replace the default GITHUB_TOKEN with a token generated from the claudie-runner GitHub App to enable write access to the berops organization's package registry, facilitating image pushes to ghcr.io across various workflows.

image

1. Jobs run on separate runners. Each job has an isolated environment, so a token minted in one job isn't available in another unless you explicitly pass it through job outputs.
2. Tokens are short-lived (1 hour default). If the pipeline runs longer than that (very plausible with
the E2E tests), a token minted at the start would be expired by the time the edit-kustomization or release jobs need it.
3. actions/create-github-app-token auto-revokes the token at job end as a security cleanup. So even if you passed it between jobs, the next job would receive an already-revoked token.

Generating per job is the standard, recommended pattern (recommended by claude).

## Summary by CodeRabbit

* **Chores**
  * CI, docs, and release workflows updated to authenticate with a generated GitHub App installation token for all repo operations (checkout, git actions, container registry login, and release uploads), replacing the previous default token to ensure consistent authenticated workflows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/berops/claudie/pull/2086)

### DIFF
--- a/.github/workflows/CI-pipeline.yml
+++ b/.github/workflows/CI-pipeline.yml
@@ -147,7 +147,6 @@ jobs:
     steps:
       - name: Generate app token
         id: app-token
-        if: (needs.check-changes.outputs.ARRAY_OF_CHANGES != '' && github.event.pull_request.draft == false)
         uses: actions/create-github-app-token@v1
         with:
           app-id: 3688229
@@ -226,7 +225,6 @@ jobs:
     steps:
       - name: Generate app token
         id: app-token
-        if: ${{ needs.build-and-push.outputs.ARRAY_OF_CHANGES != '' && github.event.pull_request.draft == false }}
         uses: actions/create-github-app-token@v1
         with:
           app-id: 3688229
--- a/.github/workflows/CI-pipeline.yml
+++ b/.github/workflows/CI-pipeline.yml
@@ -15,15 +15,24 @@ jobs:
     if: github.event.pull_request.draft == false && github.event.pull_request.head.repo.full_name == github.repository
     runs-on: self-hosted
     steps:
+      - name: Generate app token
+        id: app-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: 3688229
+          private-key: ${{ secrets.CLAUDIE_CI_APP_PRIVATE_KEY }}
+          owner: berops
+
       - uses: actions/checkout@v4
         with:
           ref: ${{ github.head_ref }}
           repository: ${{ github.event.pull_request.head.repo.full_name }}
+          token: ${{ steps.app-token.outputs.token }}
 
       - name: Merge with master branch
         uses: everlytic/branch-merge@1.1.5
         with:
-          github_token: ${{ github.token }}
+          github_token: ${{ steps.app-token.outputs.token }}
           source_ref: master
           target_branch: ${{ github.head_ref }}
   #--------------------------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


.github/workflows/automatic-docs-update.yml
```diff
@@ -24,17 +24,25 @@ jobs:
     if: ${{ (github.event.issue.pull_request && contains(github.event.comment.body, '/refresh-docs')) || github.event.label.name == 'refresh-docs' }}
     runs-on: ubuntu-22.04
     steps:
+      - name: Generate app token
+        id: app-token
+        uses: actions/create-github-app-token@v1
```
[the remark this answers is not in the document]
samuelstolicny: Skipping, all actions in this repo use floating major tags. SHA pinning should be a separate repo-wide change.

.github/workflows/automatic-docs-update.yml
```diff
@@ -24,17 +24,25 @@ jobs:
...
+        with:
+          app-id: 3688229
```
m-brando: pls comment this app-id 3688229, to which  app name it belongs for future reference  and tshoot

m-brando: or maybe we can use GH variable for this one.

.github/workflows/CI-pipeline.yml
```diff
@@ -15,15 +15,24 @@ jobs:
     if: github.event.pull_request.draft == false && github.event.pull_request.head.repo.full_name == github.repository
     runs-on: self-hosted
     steps:
+      - name: Generate app token
+        id: app-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: 3688229
```
m-brando: same as previous mentioned comment.

.github/workflows/CI-pipeline.yml
```diff
@@ -136,6 +145,14 @@ jobs:
     runs-on: self-hosted
     needs: [merge-branch, check-changes]
     steps:
+      - name: Generate app token
+        id: app-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: 3688229
```
m-brando: here also

.github/workflows/CI-pipeline.yml
```diff
@@ -206,11 +223,20 @@ jobs:
     runs-on: self-hosted
     needs: [merge-branch, check-changes, build-and-push, golangci, gotest]
     steps:
+      - name: Generate app token
+        id: app-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: 3688229
```
m-brando: here as well

.github/workflows/release-docs.yml
```diff
@@ -21,26 +21,34 @@ jobs:
     name: Create a new  docs release
     runs-on: ubuntu-22.04
     steps:
+      - name: Generate app token
+        id: app-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: 3688229
```
m-brando: here

.github/workflows/release.yml
```diff
@@ -12,6 +12,14 @@ jobs:
     name: Create a new release
     runs-on: ubuntu-22.04
     steps:
+      - name: Generate app token
+        id: app-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: 3688229
```
m-brando: here

m-brando (changes requested):

.github/workflows/automatic-docs-update.yml
samuelstolicny: Thanks, I added it as a GH var.

m-brando (approved):

PR #2152, opened 2026-06-26
fix nil panic

Despire:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved credential reconciliation to safely handle missing DNS settings for load balancer clusters.
  * Prevented potential crashes by only propagating DNS provider credentials when DNS details exist on both sides.
* **Chores**
  * Updated Claudie-related manifest image tags for both the Claudie and testing framework deployments.

### DIFF
--- a/services/manager/internal/service/reconciliation.go
+++ b/services/manager/internal/service/reconciliation.go
@@ -893,7 +893,7 @@ func updateCredentials(current, desired *spec.Clusters) (updated bool) {
 		}
 
 		// DNS can be nil if they fail to be built.
-		if clb.Dns == nil || dlb.Dns == nil { return }
+		if clb.Dns != nil && dlb.Dns != nil {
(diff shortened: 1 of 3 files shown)


services/manager/internal/service/reconciliation.go
```diff
@@ -892,9 +892,12 @@ func updateCredentials(current, desired *spec.Clusters) (updated bool) {
 			}
 		}
 
-		if !clb.Dns.Provider.CredentialsEqual(dlb.Dns.Provider) {
-			if clb.Dns.Provider.CopyCredentials(dlb.Dns.Provider) {
-				updated = true
+		// DNS can be nil if they fail to be built.
+		if clb.Dns != nil && dlb.Dns != nil {
```
bernardhalas:
```suggestion
		if clb.Dns == nil || dlb.Dns == nil { return }
```
would be easier on the eyes :)

bernardhalas (approved):
Otherwise LGTM

PR #2177, opened 2026-08-10
Feat/statefile per nodepool

### LINKED ISSUE #1788: Feature: Move provider versions to external template repository
### Motivation

Currently the provider versions are part of claudie itself, resulting in the claudie version dictating what provider version is to be used for the external templates. Ideally this would be part of the external templates itself so that they also control the version of the supported providers.

### Description

Move https://github.com/berops/claudie/blob/master/services/terraformer/server/domain/utils/templates/providers.tpl into https://github.com/berops/claudie-config to have the templates also specify the versions of the provider for terraform, while also keeping it inside claudie for backwards compatibility

### Exit criteria

- [ ] Add https://github.com/berops/claudie/blob/master/services/terraformer/server/domain/utils/templates/providers.tpl into https://github.com/berops/claudie-config
- [ ] https://github.com/berops/claudie/blob/82389fb6faac3e052d2b38f76da3277125b40ed8/services/terraformer/server/domain/utils/templates/provider.go#L21 move into  and properly document it https://github.com/berops/claudie/blob/master/services/terraformer/server/domain/utils/templates/structures.go

- [ ] Adjust the generation of the templates inside terraformer, to prefer using the versions from the external template repository, if present, if not default to using the one inside claudie.

**NOTE: for backwards compatibility, the `providers.tpl` should stay inside claudie but it should also look at the external templates and if its present there it should prefer using that one, For now we need to make it part of claudie for backwards compatibility, but it will be removed in future versions**

### LINKED ISSUE #1895: Feature: Statefile per nodepool
### Motivation
Internally, Claudie uses OpenTofu for provisioning cloud infra. Currently, it stores one statefile per cluster.
This is a problem for scenarios, where there's an OpenTofu template upgrade that needs to be synced with OpenTofu provider upgrade. The nature of the problem lies in the fact, that template upgrades are rolled-out via a rolling update on a nodepool basis. 

This means, if there's a template update, that concerns an existing nodepool, Claudie creates a new nodepool with the new template version before deleting the original nodepool created by the older template version. This creates a conflict as in one OpenTofu statefile can be just one tofu provider version.

### Description
Break up statefiles into one per nodepool.

### Exit criteria

- [ ] Claudie creates one statefile per nodepool
- [ ] OpenTofu templates can pin tofu provider version.

Despire:
Closes https://github.com/berops/claudie/issues/1895
Closes https://github.com/berops/claudie/issues/1788

Splits the single per-cluster statefile into one for the common networking infrastructure and one per nodepool (DNS unchanged), so nodepool-scoped changes no longer apply the whole cluster.

Nodepool modules consume the common infrastructure exclusively through its terraform outputs; any operation that writes the common module escalates to a full reconcile of all nodepools, keeping them consistent without tracking state.

Nodepool deletion retracts LB DNS records first, destroys the nodepool, then shrinks the common module — keeping departed providers configured until their resources are destroyed.

Templates: each of networking/nodepool/dns now carries its own provider.tpl, provider versions are pinned via provider_version.tpl, and templates of the same provider are grouped per templates version (rolling updates).

**BREAKING: existing clusters must be recreated (the old statefile layout is not migrated) and template repositories must be updated to the new contract.**

Increased the limits on the terraformer pod due to consuming more than the currently limited 2GB

```
values from reading /sys/fs/cgroup/memory.current
...
14:18:05 anon=212MiB ws=1683MiB
14:18:10 anon=580MiB ws=2056MiB
14:18:15 anon=644MiB ws=2119MiB
14:18:21 anon=644MiB ws=2119MiB
14:18:26 anon=645MiB ws=2120MiB
14:18:31 anon=645MiB ws=2120MiB
14:18:36 anon=492MiB ws=1967MiB
14:18:41 anon=526MiB ws=2019MiB
14:18:46 anon=297MiB ws=1787MiB
14:18:51 anon=736MiB ws=2228MiB
14:18:56 anon=573MiB ws=2064MiB
14:19:01 anon=350MiB ws=1838MiB
14:19:06 anon=350MiB ws=1854MiB
14:19:11 anon=350MiB ws=1854MiB
14:19:16 anon=351MiB ws=1855MiB
14:19:21 anon=261MiB ws=1763MiB
...
```

## Summary by CodeRabbit

* **New Features**
  * Infrastructure updates support targeted reconciliation and removal of individual node pools.
  * Added improved DNS record reconciliation, cleanup, and validation.
  * OpenTofu output handling now supports retrieving all outputs.
* **Bug Fixes**
  * Infrastructure-update stages are now included consistently when adding load-balancer node pools.
* **Improvements**
  * Terraformer concurrency and OpenTofu parallelism now default to 30.
  * OpenTofu initialization upgrades provider versions automatically.
* **Documentation**
  * Expanded guidance for external OpenTofu templates and environment configuration.


internal/extemplates/extofu/download.go
```diff
@@ -93,18 +98,26 @@ func Download(downloadInto string, provider *spec.Provider) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to read worktree %q: %w", gitDirectory, err)
 	}
 
-	// Worktree is not supported by go-git, thus check if worktree is set.
-	return extemplates.UnsetWorktree(gitDirectory)
+	return w.Checkout(&git.CheckoutOptions{
+		SparseCheckoutDirectories: []string{providerTemplates},
+		Hash:                      h,
+	})
```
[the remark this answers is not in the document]
Despire: this check is done in the manager service when fetching the provider

services/terraformer/internal/worker/service/internal/cluster-builder/cluster_builder.go
```diff
@@ -6,311 +6,387 @@ import (
...
-	if err := tofu.ProvidersLock(); err != nil {
-		log.Warn().Msgf("Error while locking providers from local FS mirror\n" +
-			"Continue to retrieve providers and generate hash from remote registry.")
-	}
+	var err error
+	defer func() {
+		if err != nil {
+			c.Cleanup()
+		}
+	}()
```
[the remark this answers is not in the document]
Despire: only one task for a cluster is scheduled by the manager itself, while it can happen that if may schedule 2 tasks (if the NATS ttl runs to 0) they will fight each other and end up in error and manager will reschedule again without conflicts.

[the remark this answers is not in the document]
Despire: while it may change in the future, for now it stays as is

services/terraformer/internal/worker/service/task_delete_loadbalancer_nodes.go
```diff
@@ -36,44 +36,69 @@ func deleteLoadBalancerNodes(
...
 		deleted := nodepools.FindByName(action.Delete.Nodepool, current.ClusterInfo.NodePools)
-
-		// deleting whole nodepool, if the nodepool is not found there are no side-effects.
 		current.ClusterInfo.NodePools = nodepools.DeleteByName(current.ClusterInfo.NodePools, action.Delete.Nodepool)
 
-		// Include the deleted nodepools for generating the provider that was used
-		// in that nodepool so that the infrastructure will get correctly destroyed.
-		if deleted != nil {
-			lb.GhostNodePools = append(lb.GhostNodePools, deleted)
+		if err := lb.DestroyNodePool(ctx, logger, deleted, stores.s3); err != nil {
```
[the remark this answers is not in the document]
Despire: A task is always scheduled for a nodepool that exists, if it is the case that it is nil as you describe then nothing will be done in the task with the nodepool.

services/terraformer/internal/worker/service/internal/loadbalancer/loadbalancer.go
```diff
@@ -41,114 +57,443 @@ type LBcluster struct {
...
 
+	nodeIPs := nodepools.PublicEndpoints(l.Cluster.ClusterInfo.NodePools) // include all of the nodes, not just dynamic.
+	if err := dnsBuilder.Init(logger, nodeIPs, l.Cluster.Dns); err != nil {
+		return err
+	}
+	defer dnsBuilder.Cleanup()
+
+	if err := dnsBuilder.ReconcileRecords(); err != nil {
+		return fmt.Errorf("error while reconciling dns records for cluster %q: %w", builder.ClusterId, err)
+	}
```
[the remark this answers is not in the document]
Despire: The nil check is inside the DnsBuilder.Init function that will fail and the whole task will end up with an error and thus the manager will revert back the task. In fact this would be a logic error within the manager which scheduled nodepools operations before a valid DNS existed.

services/terraformer/internal/worker/service/internal/kubernetes/kubernetes.go
```diff
@@ -1,90 +1,347 @@
...
+	latest, err := builder.OutputOnlyCommon()
+	if err != nil {
+		return fmt.Errorf("failed to read output of common nodepool infrastructure: %w", err)
+	}
+
+	group, _ /* ctx Currently not used */ := errgroup.WithContext(ctx)
+	group.SetLimit(k.NodePoolConcurrencyLimit)
+
+	for i, np := range dynamic {
+		nodepoolStateFileKey := store.ObjectKey(builder.InputManifest, cluster_builder.NodePoolStateKey(builder.ClusterId, np.Name))
```
m-brando, line 195: The pre-split statefile key <project>/<clusterId> in not referenced anywhere.

so deleting a cluster created before the split reports a clean success while every VM and network keeps running.

services/terraformer/internal/worker/service/internal/loadbalancer/loadbalancer.go
```diff
@@ -41,114 +57,445 @@ type LBcluster struct {
...
-	group := errgroup.Group{}
+	// Exclude nodes with no IP's, as those are nodes that were not successfully built.
+	nodeIPs = slices.DeleteFunc(nodeIPs, func(s string) bool { return s == "" })
+
+	if err := lbuilder.Init(logger, dynamic); err != nil {
+		return cluster_builder.ExplainUnknownCommit(err, lbuilder.ClusterId)
+	}
+	defer lbuilder.Cleanup()
+
+	group, ctx := errgroup.WithContext(ctx)
```
m-brando, line 266: `group, ctx := errgroup.WithContext(ctx)` shares the cancellable context between the DNS goroutine and the nodepool/common goroutine, so a DNS failure cancels unrelated concurrent work.

services/manager/internal/service/reconciliate_loadbalancers.go
```diff
@@ -548,29 +548,27 @@ func ScheduleAdditionLoadBalancerNodePools(
 	diff *NodePoolsDiffResult,
 	opts LoadBalancerNodePoolsOptions,
 ) *spec.TaskEvent {
-	pipeline := []*spec.Stage{}
+	var pipeline []*spec.Stage
```
m-brando, line 552: Removing the `if !opts.IsStatic` guard schedules a terraformer UPDATE_INFRASTRUCTURE subpass for static LB nodepool additions, but static additions emit Update_AddedLoadBalancerNodes_, which has no case in reconcileInfrastructure.

could pls verify ?

m-brando:
Not about this PR's changes — this landed in `master` via #2183 — but flagging it here since it is adjacent to the secret/state handling and you authored both.

`SecretMetadata` (`services/manager/internal/service/managementcluster/secret.go:27`) went from `<clusterName>-<hash>-<type>` to `<projectName>-<clusterName>-<type>`. Two consequences:

**Pre-existing secrets are never cleaned up.** `delete_kubeconfig.go:26` and `delete_metadata.go:24` were updated to the new name in the same change, so they no longer match anything written by an earlier version. Deleting a cluster that was created before the upgrade leaves its `<clusterName>-<hash>-kubeconfig` behind permanently — a live kubeconfig for infrastructure that no longer exists. Relevant to #2177 specifically, since that PR requires existing clusters to be deleted and recreated, so every current install takes this path.

**The new name is ambiguous.** Without the hash, project `a-b` + cluster `c` and project `a` + cluster `b-c` both render `a-b-c-kubeconfig`. Project names have no charset restriction (`validate:"required"`, `internal/api/manifest/manifest.go:42`; cluster names only `max=15`), so hyphens are allowed in both halves and two unrelated clusters can silently overwrite each other's credentials.

Keeping the hash (e.g. `<projectName>-<clusterName>-<hash>-<type>`) fixes the ambiguity. For the orphans, `claudie.io/cluster-id` is already a label on the secret, so a label-selector delete would cover both naming schemes.

Happy to move this to its own issue if you'd rather not carry it on this PR.

services/manager/internal/service/reconciliate_loadbalancers.go
Despire: Addition for loadbalancers emit 

https://github.com/berops/claudie/blob/0ae2a6d27c72cedd0c3c485b14adca01cec008ec/services/manager/internal/service/reconciliate_loadbalancers.go#L698

which is handled in:

https://github.com/berops/claudie/blob/0ae2a6d27c72cedd0c3c485b14adca01cec008ec/services/terraformer/internal/worker/service/task_update_infrastructure.go#L69

Despire:
>Pre-existing secrets are never cleaned up. delete_kubeconfig.go:26 and delete_metadata.go:24 were updated to the new name in the same change, so they no longer match anything written by an earlier version. Deleting a cluster that was created before the upgrade leaves its <clusterName>-<hash>-kubeconfig behind permanently — a live kubeconfig for infrastructure that no longer exists. Relevant to https://github.com/berops/claudie/pull/2177 specifically, since that PR requires existing clusters to be deleted and recreated, so every current install takes this path.

Currently that is the intended way. After merging this PR, the old clusters will no longer be supported, Simply because of the changes of how statefiles are handled after merging the changes in this PR.

>The new name is ambiguous. Without the hash, project a-b + cluster c and project a + cluster b-c both render a-b-c-kubeconfig. Project names have no charset restriction (validate:"required", internal/api/manifest/manifest.go:42; cluster names only max=15), so hyphens are allowed in both halves and two unrelated clusters can silently overwrite each other's credentials.

The new name is the <k8s-namespace>-<inputmanifest-name>-<cluster-name> pattern.

Within a single InputManifest you cannot have multiple clusters with the same name so there cannot be any collisions.
Within a single k8s namespace you cannot have two InputManifest with the same name, they always just reference the same Inputmanifest.

services/terraformer/internal/worker/service/internal/loadbalancer/loadbalancer.go
Despire: Since this is in the `DestroyAll` function, which is for destroying all of the infrastructure, this is okay.

On failure the manager simply reschedules the whole deletion again.

services/terraformer/internal/worker/service/internal/kubernetes/kubernetes.go
Despire: If we want the next release of claudie to be able to delete old clusters, I think we can create another issue for that to implement it. But in the context of this issue, this is the expected behavior.

Despire: cc @bernardhalas

bernardhalas: As this behavior concerns just a single release, I think it's acceptable to mention that in the release notes - that the cluster + secrets need to be deleted prior the upgrade.

m-brando (approved):

PR #2191, opened 2026-08-14
Feat/ssh pings

### LINKED ISSUE #2181: Feature: SSH pings for loadbalancers
### Manager: replace ICMP pings with SSH-port pings for loadbalancer node healthchecks

The manager healthchecks loadbalancer nodes via raw ICMP echo.

ICMP is blocked or rate-limited by many cloud firewalls and requires raw-socket privileges, producing false "unreachable" results, while the SSH port is claudie's actual management channel and is guaranteed open by our own security groups.

Shared-IP nodepools break ICMP entirely: nodes behind one NAT address are indistinguishable by echo.

Replace with a TCP dial to each node's `ip:SshPort` keeping the existing timeout/retry/concurrency semantics.

Despire:
Closes https://github.com/berops/claudie/issues/2181

Changes in this Pull Requests drops all previous usage of raw ICMP packets for pinging (only used for LoadBalancer nodes) and replaces them with SSH pings that timed-out after ~4 seconds if the node is unreachable.

## Summary by CodeRabbit

* **New Features**
  * Node reachability checks now use SSH with retries and endpoint-specific diagnostics.
  * Node deletion supports static, dynamic, and untracked nodes.
  * SSH username and port defaults are applied consistently.

* **Bug Fixes**
  * Improved handling of unreachable nodes and partial health-check results.

* **Documentation**
  * Updated concurrent health-check settings and reduced the default worker count from 20 to 10.

* **Security**
  * Removed unnecessary raw-network access from service containers.

* **Chores**
  * Updated container image versions.

### DIFF
--- a/internal/clusters/ping4.go
+++ b/internal/clusters/ping4.go
@@ -23,7 +23,7 @@ const (
 
 	// PingRetryCount is the number of times the ping will be retried
 	// to determine if the node is healthy of not.
-	PingRetryCount = 3
+	PingRetryCount = 4
 )
 
 var (
@@ -94,6 +94,7 @@ func SSHPing(logger zerolog.Logger, retries int, dst NodeEndpoint) error {
 
 	var err error
 	for i := range retries {
+		logger.Debug().Msgf("SSH ping %s:%s", dst.Ip, dst.Port)
 		if err = sshPing(PingTimeout, dst); err == nil {
 			break
 		}
@@ -101,7 +102,7 @@ func SSHPing(logger zerolog.Logger, retries int, dst NodeEndpoint) error {
 			break
 		}
 		wait := 1 * time.Second
-		logger.Err(err).Msgf("failed to ssh ping node %q, retrying again in %s", dst.Ip, wait)
+		logger.Warn().Msgf("failed to ssh ping node %q: %v, retrying again in %s", dst.Ip, err, wait)
 		time.Sleep(wait)
 	}
 	if err != nil {
--- a/internal/clusters/ping4.go
+++ b/internal/clusters/ping4.go
@@ -1,290 +1,205 @@
 package clusters
 
 import (
-	"context"
 	"errors"
 	"fmt"
 	"maps"
 	"net"
-	"os"
 	"slices"
 	"sync"
 	"time"
 
 	"github.com/berops/claudie/internal/envs"
+	"github.com/berops/claudie/internal/nodepools"
 	"github.com/berops/claudie/proto/pb/spec"
 	"github.com/rs/zerolog"
 
-	"golang.org/x/net/icmp"
-	"golang.org/x/net/ipv4"
+	"golang.org/x/crypto/ssh"
 )
 
 const (
 	// How long to wait before considering the ping packet to be lost.
-	PingTimeout = 3 * time.Second
+	PingTimeout = 2 * time.Second
+
 	// PingRetryCount is the number of times the ping will be retried
 	// to determine if the node is healthy of not.
-	PingRetryCount = 5
+	PingRetryCount = 3
 )
 
-// How many goroutines will be used to ping nodes of a cluster.
-var pingConcurrentWorkers = envs.GetOrDefaultInt("PING_CONCURRENT_WORKERS", 20)
+var (
+	// How many goroutines will be used to ping nodes of a cluster.
+	pingConcurrentWorkers = envs.GetOrDefaultInt("PING_CONCURRENT_WORKERS", 10)
 
-// ErrEchoTimeout is returned when the reply is not received within the requested timeout.
-var ErrEchoTimeout = errors.New("icmp request timeout")
+	// ErrUnreachable is returned when a node cannot be reached via an SSH ping.
+	ErrUnreachable = errors.New("failed to reach node")
+)
 
-func ping4(logger zerolog.Logger, conn *icmp.PacketConn, id, seq int, dst *net.UDPAddr, timeout time.Duration) error {
-	m := icmp.Message{
-		Type: ipv4.ICMPTypeEcho,
-		Code: 0,
-		Body: &icmp.Echo{
-			ID:   id,
-			Seq:  seq,
-			Data: []byte("ping"),
-		},
-	}
-
-	b, err := m.Marshal(nil)
-	if err != nil {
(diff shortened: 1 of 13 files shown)


internal/clusters/ping4.go
```diff
@@ -1,290 +1,206 @@
...
-					lbsip[lb][np] = append(lbsip[lb][np], ip)
-				}
-			}
+		if i == retries-1 {
+			break
 		}
+		wait := 1 * time.Second
+		logger.Warn().Msgf("failed to ssh ping node %q: %v, retrying again in %s", dst.Ip, err, wait)
+		time.Sleep(wait)
 	}
```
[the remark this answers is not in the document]
Despire: This is logically, true if there are no tries then all nodes can be considered as reachable, this is up to the caller of that function to make that decision

internal/clusters/ping4.go
```diff
@@ -1,290 +1,206 @@
...
+					Username: nodepools.NodeSSHUsername(n),
+					SSHKey:   "",
+				}
+
+				switch t := np.Type.(type) {
+				case *spec.NodePool_DynamicNodePool:
+					ep.SSHKey = t.DynamicNodePool.PrivateKey
+				case *spec.NodePool_StaticNodePool:
+					ep.SSHKey = t.StaticNodePool.NodeKeys[n.Public]
+				}
```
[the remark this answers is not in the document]
Despire: missing credentials lead to unreachable nodes. That would signal an inner bug within Claudie itself.
Therefore this is working as expected. And if there are inconsistencies, it possible to remove the nodepools from the InputManifest to unblock.

services/kuber/internal/worker/service/task_delete_nodes.go
```diff
@@ -190,33 +157,197 @@ func deleteNodesFromCluster(
...
+	var ni []nodes.NodeInfo
+
+	for i := range n {
+		ni = append(ni, nodes.NodeInfo{
+			Fullname: n[i].Name,
+			K8sName:  strings.TrimPrefix(n[i].Name, fmt.Sprintf("%s-", k8s.ClusterInfo.Id())),
+			// Zero value, treat as unreachable by default.
+			Ep: clusters.NodeEndpoint{},
+		})
 	}
```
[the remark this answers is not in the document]
Despire: The VM was removed from the cluster due to it not being reachable (for example VM was shut down) and then it came back online and re-joined the. cluster. But that VM is no longer tracked inside the Claudie state and therefore no longer expected to have workloads/volumes on it. Now we could read the readiness check of that node for this but if the Node was already removed from the cluster, we shouldn't really expect to have workloads on that node

internal/clusters/ping4.go
```diff
@@ -1,290 +1,206 @@
...
 
-	"golang.org/x/net/icmp"
-	"golang.org/x/net/ipv4"
+	"golang.org/x/crypto/ssh"
 )
 
 const (
 	// How long to wait before considering the ping packet to be lost.
-	PingTimeout = 3 * time.Second
+	PingTimeout = 2 * time.Second
```
m-brando, line 22: PingTimeout was reduced from 3s to 2s, but it is also used by an unrelated single-shot probe at healthcheck.go:316 (net.DialTimeout to the control node's 6443) which has no retries to absorb the tighter budget.

wondering if 2 seconds without any in  healthcheck.go:316  retry is enough

services/manager/internal/service/healthcheck.go
```diff
@@ -191,7 +190,7 @@ func CheckNodesStatus(logger zerolog.Logger, state *spec.Clusters, hc HealthChec
 
 	result.UnknownLoadBalancersNodes, err = clusters.PingLoadBalancerNodes(logger, state)
 	if err != nil {
-		if !errors.Is(err, clusters.ErrEchoTimeout) {
+		if !errors.Is(err, clusters.ErrUnreachable) {
```
m-brando, line 193: The `if !errors.Is(err, clusters.ErrUnreachable)` bail-out can never fire, because `PingLoadBalancerNodes` now returns only joined SSHPing errors and all of them match ErrUnreachable
So the setup-failure escape hatch the comment below describes is dead code.

services/kuber/internal/worker/service/internal/nodes/delete.go
```diff
@@ -141,78 +117,69 @@ func (d *Deleter) DeleteNodes(logger zerolog.Logger) error {
...
 
-		if err := clusters.Ping(logger, UnreachableNodesPingCount, master.publicEndpoint); err != nil {
-			if errors.Is(err, clusters.ErrEchoTimeout) {
-				logger.
-					Info().
-					Msgf(
-						"Node %s is unreachable, marking node with `out-of-service` taint "+
-							"before deleting it from the cluster", master.k8sName,
-					)
+		if err := clusters.SSHPing(logger, clusters.PingRetryCount, master.Ep); errors.Is(err, clusters.ErrUnreachable) {
```
m-brando, line 145: The rewrite to `errors.Is(err, clusters.ErrUnreachable)` dropped the else arm that logged:

"Failed to determine if node %s is reachable or not, proceeding with default deletion"

m-brando (changes requested):

services/kuber/internal/worker/service/internal/nodes/delete.go
Despire: Yes, because the else branch was dead_code, since now SSH pings either succeed or return an error that is `clusters.ErrUnreachable`

services/manager/internal/service/healthcheck.go
Despire: Yes, this branch is dead_code as with the latest changes.

However I still want to keep it, 'case if the funcion `PingLoadBalancerNodes` returns a non `clusters.ErrUnreachable` error in the future this will still hold and will block the reconciliation, in case
of continuing.

internal/clusters/ping4.go
Despire: Yes it is enough. A round-robin packet takes roughly 150ms over half of the globe (https://www.norvig.com/21-days.html#answers) 

Therefore a tcp dial (3-way handshake) should be established in 3 packets which should take 3x (150) ~ 450ms
2 seconds should be enough.

Also the healthcheck is executed in a loop and can be retried on the next iteration on failure.

m-brando (approved):

PR #2193, opened 2026-08-14
fix unreachable node deletion pipeline

Despire:
For scheduling deletion of unreachable nodes if the node is no longer tracked by the Claudie state, irrelevant pipeline stages
are scheduled for the task which end up in an endless loop (due to the node/nodepool not being in the state)

A node/nodepool may not be in the state if it was removed due to being unreachable but the VM (if it was a static one) came back online and re-joined the cluster. Claudie then  scheduled deletion of the node again but as its no longer tracked in the state we need to only schedule one stage for the pipeline.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Kubernetes node and node pool deletion handling when tracked infrastructure information is incomplete.
  - Prevented unnecessary cleanup actions for node pools that are no longer tracked.
  - Ensured complete node pool deletions consistently perform all required cleanup and control-plane checks.
  - Improved handling of unreachable infrastructure details during deletion workflows.

- **Tests**
  - Added comprehensive coverage for partial and complete deletions across static and dynamic node pools, proxy configurations, and load balancer scenarios.

### DIFF
--- a/services/manager/internal/service/reconciliate_kubernetes.go
+++ b/services/manager/internal/service/reconciliate_kubernetes.go
@@ -1326,35 +1326,35 @@ func ScheduleDeletionsInNodePools(
 	}
 
 	for np, nodes := range diff.Deleted {
+		// Nodepools deleted as a whole are always part of the claudie tracked
+		// state. [NodePoolsDiff] fills [NodePoolsDiffResult.Deleted] from the
+		// current state and the reconciliate_unreachable_nodes drift logic only
+		// ever fills [NodePoolsDiffResult.PartiallyDeleted], thus the lookup
+		// cannot return nil.
 		currentStateNodePool := nodepools.FindByName(np, current.K8S.ClusterInfo.NodePools)
-		if currentStateNodePool != nil {
-			// If the ApiServer is on the kubernetes cluster on deletion of the
-			// control plane nodes the Kubeadm config map needs to be updated which
-			// is used during the join operation of new nodes.
-			if opts.HasApiServer && currentStateNodePool.IsControl {
-				kuber.Kuber.SubPasses = append(kuber.Kuber.SubPasses, []*spec.StageKuber_SubPass{
-					{
-						Kind: spec.StageKuber_PATCH_KUBEADM,
-						Description: &spec.StageDescription{
-							About:      "Updating Kubeadm certSANs",
-							ErrorLevel: spec.ErrorLevel_ERROR_FATAL,
-						},
-					},
-					{
-						Kind: spec.StageKuber_CILIUM_RESTART,
-						Description: &spec.StageDescription{
-							About: "Rollout restart of cilium pods",
-							// Rollout restart failing is not a fatal error.
-							ErrorLevel: spec.ErrorLevel_ERROR_WARN,
-						},
-					},
-				}...)
-			}
-		}
 
-		// Contrary to addition, when deleting we do not skip if the nodepool is not found.
-		// As deleting from a not found nodepool should be handled gracefully by the existing
-		// services.
+		// If the ApiServer is on the kubernetes cluster on deletion of the
+		// control plane nodes the Kubeadm config map needs to be updated which
+		// is used during the join operation of new nodes.
+		if opts.HasApiServer && currentStateNodePool.IsControl {
+			kuber.Kuber.SubPasses = append(kuber.Kuber.SubPasses, []*spec.StageKuber_SubPass{
+				{
+					Kind: spec.StageKuber_PATCH_KUBEADM,
+					Description: &spec.StageDescription{
+						About:      "Updating Kubeadm certSANs",
+						ErrorLevel: spec.ErrorLevel_ERROR_FATAL,
+					},
+				},
+				{
+					Kind: spec.StageKuber_CILIUM_RESTART,
+					Description: &spec.StageDescription{
+						About: "Rollout restart of cilium pods",
+						// Rollout restart failing is not a fatal error.
+						ErrorLevel: spec.ErrorLevel_ERROR_WARN,
+					},
+				},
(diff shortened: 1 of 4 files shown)


services/manager/internal/service/reconciliate_kubernetes_test.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+}
+
+// The pipeline shape decides which services participate in a node/nodepool
+// deletion. For nodepools missing from the current state (drift, e.g. a node
+// re-joined the cluster after being removed from the state and the input
+// manifest) only the first kuber stage may be scheduled: it deletes the nodes
+// from the cluster without committing an update, and since the task delta is
+// then never consumed, any update committed by a later stage would be refused
+// by the manager, re-running that stage indefinitely.
```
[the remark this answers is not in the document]
Despire: It does if the node is untracked.

services/manager/internal/service/reconciliate_kubernetes_test.go
```diff
@@ -0,0 +1,241 @@
...
+				kuberDelete,
+				{ansibler: proxyPasses},
+				{terraformer: []spec.StageTerraformer_SubPassKind{spec.StageTerraformer_UPDATE_INFRASTRUCTURE}},
+			},
+			wantNodepool: p1,
+			wantNodes:    []string{"node-1"},
+		},
+		{
+			name:         "untracked nodepool schedules only the tracked state removal stage",
+			nodepool:     namedStaticNodePool(p1),
```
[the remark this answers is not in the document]
Despire: fixed in lates commit

services/manager/internal/service/reconciliate_kubernetes.go
```diff
@@ -1407,21 +1425,33 @@ func ScheduleDeletionsInNodePools(
...
+		// reconciliate_unreachable_nodes logic was involved.
+		//
+		// Scheduling a node/nodepool deletion that is not part
+		// of the tracked state is a valid operation, as the node
+		// may have been unreachable, claudie decided to delete it
+		// but it may have come back online and re-connected to the
+		// cluster. Therefore the following stages are only to be
+		// considered when the node/nodepool is part of the claudie
+		// tracked state.
+		if currentStateNodePool != nil {
```
m-brando: This guard is unreachable, and would not be safe if it were reached.

**Unreachable:** every `diff.Deleted` key comes from the current state. `diff.go:293-296` fills that map by iterating `old`, which is `NodePoolsView(old.GetClusterInfo())` — the same `current.K8S` looked up here. `PendingDynamic/StaticDeletions` and both `reconciliate_unreachable_nodes.go` call sites fill only `PartiallyDeleted`. So `currentStateNodePool != nil` is always true here, and the comment about `reconciliate_unreachable_nodes` does not apply to this loop.

**Not safe either:** this loop sets `WithNodePool: true` (line 1470), which takes the *first* branch of kuber's `deleteFromState`. `nodepools.DeleteByName` is a no-op for an unknown name, then `update.Commit()` runs regardless — `ConsumeUpdateResult` rejects it and the task reschedules under a new UUID forever, i.e. the bug this PR fixes. And with `removeFromClusterStage` dropped, nothing removes the node from the cluster either. The drift handling exists only in the `else` branch, which `WithNodePool: true` never reaches.

Suggest dropping the guard here and leaving a comment that `diff.Deleted` names are always tracked.

services/manager/internal/service/reconciliate_kubernetes_test.go
```diff
@@ -0,0 +1,285 @@
...
+					[]spec.StageAnsibler_SubPassKind{spec.StageAnsibler_REMOVE_CLAUDIE_UTILITIES},
+					proxyPasses...,
+				)},
+			},
+			wantWithNodePool: true,
+			wantNodepool:     p2,
+			wantNodes:        []string{"node-1", "node-2"},
+		},
+		{
+			name:             "untracked whole nodepool deletion schedules only the tracked state removal stage",
```
m-brando: This case and the one at line 239 describe a scenario that cannot occur — `diff.Deleted` never carries an untracked nodepool (see the comment on `reconciliate_kubernetes.go:1443`). They pass, but assert nothing. Remove them along with the guard.

services/manager/internal/service/reconciliate_kubernetes_test.go
```diff
@@ -0,0 +1,285 @@
...
+type expectedStage struct {
+	kuber       []spec.StageKuber_SubPassKind
+	ansibler    []spec.StageAnsibler_SubPassKind
+	terraformer []spec.StageTerraformer_SubPassKind
+}
+
+func assertStage(t *testing.T, i int, got *spec.Stage, want expectedStage) {
+	t.Helper()
+
+	switch {
```
m-brando, line 67: No `default` arm, so an expectation with all three slices nil accepts any stage, including a wrong kind.

Also `kuberDelete` matches both `removeFromTrackedStage` and `removeFromClusterStage`, since only subpass kinds are compared — the tracked cases would still pass if the code scheduled `removeFromTrackedStage` twice, which is exactly what this PR changes. Asserting `Description.About` plus a failing `default` arm would pin both.

m-brando (changes requested):
Requesting changes on one item, the rest are nits.

The fix in the `PartiallyDeleted` loop is correct. The copy of it in the `diff.Deleted` loop is not — inline.

Nit outside the diff, `reconciliate_kubernetes.go:1361`: the Longhorn subpasses are appended to `kuber.Kuber.SubPasses` before the `currentStateNodePool != nil` check, so on the untracked path they are built and discarded with the stage. Harmless today, but a subpass added there later that is needed for that path would vanish silently.

m-brando (approved):

PR #2199, opened 2026-09-01
fix: environment variables across the repo

Despire:
## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive environment variable reference organized by service, with defaults, descriptions, and configuration wiring.
  * Simplified customization guidance and linked it to the new reference.

* **Configuration**
  * Renamed the Kuber worker setting to `KUBER_CONCURRENT_WORKERS`.
  * Removed obsolete operator host and manually configured NATS cluster URL settings.
  * Renamed the webhook TLS setting and updated the bucket name default.

* **Deployment Updates**
  * Updated component and testing framework image versions across deployment manifests.


docs/environment-settings/environment.md
```diff
@@ -1,74 +1,105 @@
...
-```
+- Each Deployment references specific keys of the `env` ConfigMap explicitly (there is no `envFrom`), so adding
+  a new key to the ConfigMap alone has no effect unless the Deployment also references it.
+- Any variable that is not set falls back to the default value built into the service (listed below).
+- `DATABASE_URL` (manager) and `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` (terraformer) are sourced from the
+  `mongo-secret` and `minio-secret` Secrets, not from the ConfigMap.
+- `NAMESPACE` is injected automatically from the pod metadata, and `NATS_CLUSTER_URL` is set directly in each
+  Deployment to `nats://nats.$(NAMESPACE).svc.cluster.local`.
+- To set a variable that is not wired into a Deployment (e.g. `CLAUDIE_NAMESPACES` or `PING_CONCURRENT_WORKERS`),
+  add it to the respective Deployment.
```
bernardhalas, line 17: I know it's off-topic from the md file, but still - I don't understand why do we source some env vars from the ConfigMap and some not. Is that a good approach?

Despire: All of the envs are sourced from the .env file, the only exception are `mongo-secret` and `minio-secret` which are referenced via a secret. As for why, they are the only ones carrying credentials.

m-brando (approved):

Issue #2180, opened 2026-08-10
Feature: Improve 0-sized nodepools


Despire:
### Investigate: terraformer-only workflow for 0-sized nodepools

Autoscaled nodepools allow `min: 0`, but the manager schedules the full pipeline (terraformer → ansibler → kube-eleven → kuber) for any nodepool diff regardless of node count — `NodePoolsDiff` reports an added nodepool even with zero nodes, so e.g. adding an autoscaled nodepool that starts empty runs whole-cluster ansible playbooks and kubeone apply for nothing.

Investigate whether operations whose diff only touches 0-sized nodepools (new autoscaled nodepool at zero, rolling update of a scaled-to-zero nodepool) can schedule just the terraformer stage, and which flows must keep the full pipeline.

Outcome: a written summary of affected flows plus a follow-up implementation issue if feasible.

Despire:
The whole context for adding new nodepool into the cluster is: 

https://github.com/berops/claudie/blob/7b7bac98ecf0df64308365952580a4266bf4e3aa/services/manager/internal/service/reconciliate_kubernetes.go#L587

The decision should be made within the function.

PR #2214, opened 2026-09-08
feat/0-sized-nodepools

misosvec:
Issue #2180 

This change prevents running `ansibler`, `kube-eleven`, and `kuber` when the autoscaler creates a node pool with 0 nodes.

This is irrelevant to static node pools and creation of dynamic node pools with fixed size 0 is not allowed. Only the autoscaler scenario applies.

## Summary by CodeRabbit

- **Bug Fixes**
  - Node pool provisioning now runs for both static and dynamic node pools when nodes are added.
  - Empty node pool additions no longer trigger provisioning or load balancer updates.

- **Maintenance**
  - Updated platform and testing framework components to newer versions, including improvements to node management, cluster operations, and infrastructure provisioning.

### DIFF
--- a/services/manager/internal/service/reconciliate_kubernetes.go
+++ b/services/manager/internal/service/reconciliate_kubernetes.go
@@ -867,14 +867,14 @@ func ScheduleAdditionsInNodePools(
 			pipeline = append(pipeline, terraformerStage)
 		}
 
-		if !opts.IsStatic && len(toAdd.Nodes) != 0 {
+		if len(toAdd.Nodes) != 0 {
 			pipeline = append(pipeline, ansiblerStage)
 			pipeline = append(pipeline, kubeElevenStage)
 			pipeline = append(pipeline, kuberStage)
-		}
 
-		if updateLoadBalancersStage != nil {
-			pipeline = append(pipeline, updateLoadBalancersStage)
+			if updateLoadBalancersStage != nil {
+				pipeline = append(pipeline, updateLoadBalancersStage)
+			}
 		}
 
 		return &spec.TaskEvent{
--- a/services/manager/internal/service/reconciliate_kubernetes.go
+++ b/services/manager/internal/service/reconciliate_kubernetes.go
@@ -867,9 +867,11 @@ func ScheduleAdditionsInNodePools(
 			pipeline = append(pipeline, terraformerStage)
 		}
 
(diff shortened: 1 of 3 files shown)


services/manager/internal/service/reconciliate_kubernetes.go
```diff
@@ -867,9 +867,11 @@ func ScheduleAdditionsInNodePools(
...
 		}
 
-		pipeline = append(pipeline, ansiblerStage)
-		pipeline = append(pipeline, kubeElevenStage)
-		pipeline = append(pipeline, kuberStage)
+		if !opts.IsStatic && len(toAdd.Nodes) != 0 {
+			pipeline = append(pipeline, ansiblerStage)
+			pipeline = append(pipeline, kubeElevenStage)
+			pipeline = append(pipeline, kuberStage)
+		}
```
Despire: remove the `!opts.IsStatic`. 
(The `opts.IsStatic` flag is set if the `diff` that is passed into the function is for static nodepools)

Right now what this results into is that, If a new static nodepool is added then nothing gets scheduled for it.

 
Also include the `updateLoadBalancersStage`. 

```go
if len(toAdd.Nodes) != 0 {
...
  if updateLoadBalancersStage != nil {
     pipeline = append(pipeline, updateLoadBalancersStage)
  }
}
```

Despire (changes requested):
This needs some changes before merging.

services/manager/internal/service/reconciliate_kubernetes.go
```diff
@@ -867,9 +867,11 @@ func ScheduleAdditionsInNodePools(
 			pipeline = append(pipeline, terraformerStage)
 		}
 
-		pipeline = append(pipeline, ansiblerStage)
-		pipeline = append(pipeline, kubeElevenStage)
-		pipeline = append(pipeline, kuberStage)
+		if !opts.IsStatic && len(toAdd.Nodes) != 0 {
```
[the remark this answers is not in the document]
Despire: Incorrect I'have described the issue with `opts.IsStatic`.

services/manager/internal/service/reconciliate_kubernetes.go
misosvec: I made the changes.

Despire (approved):
LGTM

PR #2220, opened 2026-09-10
fix: reject unusable AutoscalerConfig min/max in nodepool validation

### LINKED ISSUE #2218: Bug: Autoscaler Config validation
### Current Behaviour
`DynamicNodePool.AutoscalerConfig` is optional, but two definitions of "defined" are in play. The struct tags treat any non-zero block as present, so `required_without=Count` is satisfied and `count` is no longer required. Runtime uses `isDefined()` (`Min >= 0 && Max > 0`). A block that is non-zero but not usable passes validation, is then ignored by `CreateNodepools`, and the nodepool is created with `count: 0` and no autoscaler. No error anywhere.

Accepted today:
- `autoscaler: {min: 3}` (max omitted)
- `autoscaler: {min: -1, max: 5}`
- `autoscaler: {max: -3}`
- `autoscaler: {min: 5, max: 2}` is usable per `isDefined()` and reaches the cluster-autoscaler with min > max.

### Expected Behaviour
A non-zero autoscaler block must be usable: `0 <= min <= max <= 255` and `max >= 1`, otherwise validation fails. Omitting the block keeps requiring `count`. `min: 0` stays valid for scale-from-zero.

### Steps To Reproduce
1. Define a dynamic nodepool with `autoscaler: {min: 3}` and no `count`.
2. Apply the InputManifest.
3. Validation passes; the nodepool ends up with 0 nodes and no autoscaler.

### Anything else to note
Fix is two tags in `api/manifest/manifest.go` (`internal/api/manifest` on master), no logic change:

```go
Min int32 `validate:"gte=0" yaml:"min" json:"min,omitempty"`
Max int32 `validate:"gte=0,max=255,gtefield=Min" yaml:"max" json:"max,omitempty"`
```

Include unit tests for this change.

pujitha24:
Motivation:
DynamicNodePool.AutoscalerConfig is meant to be optional and mutually
exclusive with Count, but the struct validation tags did not actually
constrain Min/Max to a usable range. A non-zero but unusable autoscaler
block (e.g. {min: 3} with max omitted, {min: -1, max: 5}, {max: -3}, or
{min: 5, max: 2}) satisfied "required_without=Count" and passed struct
validation, since the runtime isDefined() check (Min >= 0 && Max > 0)
does not catch min > max or a max that defaults to 0. Such a nodepool
is then silently created with count: 0 and no autoscaler, with no
validation error anywhere.

Approach:
Add validate tags to AutoscalerConfig.Min ("gte=0") and Max
("gte=0,max=255,gtefield=Min") so a non-zero autoscaler block must be
usable: 0 <= min <= max <= 255. Omitting the block still requires
Count via the existing required_without=Count tag. min: 0 stays valid
for scale-from-zero, since Max's gtefield=Min still allows max >= 0.
No other logic changes.

Validation:
- go build ./... passes for the whole repo.
- go test -short ./... passes for the whole repo.
- Added unit tests to api/manifest/validate_test.go covering the
  accepted-today-but-unusable cases from the issue (min without max,
  negative min, negative max, min > max) as errors, and confirmed
  min: 0 with a valid max still passes.
- Confirmed the new no-max test case fails before the fix ("An error
  is expected but got nil") and passes after, by temporarily reverting
  only the two struct tags and re-running
  `go test ./api/manifest/... -run TestNodepool -v`.
- golangci-lint (v2.13.2, matching the version pinned in
  .github/workflows/CI-pipeline.yml) run against ./api/manifest/...
  reports 0 issues.

Report: https://github.com/berops/claudie/issues/2218
Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>
Assisted-by: claude-sonnet-5 (via Claude Code)

Fixes #2218

## Summary by CodeRabbit

* **Bug Fixes**
  * Autoscaler settings now reject negative minimum or maximum values.
  * Maximum values can no longer be lower than the configured minimum.
  * A minimum value of zero is now accepted when valid.

### DIFF
--- a/api/manifest/manifest.go
+++ b/api/manifest/manifest.go
@@ -254,9 +254,9 @@ type DynamicNodePool struct {
 // Autoscaler configuration on per nodepool basis. Defines the number of nodes, autoscaler will scale up or down specific nodepool.
 type AutoscalerConfig struct {
 	// Minimum number of nodes in nodepool.
-	Min int32 `yaml:"min" json:"min,omitempty"`
+	Min int32 `validate:"gte=0" yaml:"min" json:"min,omitempty"`
 	// Maximum number of nodes in nodepool.
-	Max int32 `validate:"max=255" yaml:"max" json:"max,omitempty"`
+	Max int32 `validate:"gte=0,max=255,gtefield=Min" yaml:"max" json:"max,omitempty"`
 }
 
 // Provider spec is further specification build on top of the data from any of the provider instance.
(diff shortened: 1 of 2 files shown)


Despire (approved):
LGMT

PR #2221, opened 2026-09-11
fix: require externalNetworkName by OpenStack provider type

### LINKED ISSUE #2217: Bug: validateExternalNet not working as intended
### Current Behaviour
`validateExternalNet` (`internal/api/manifest/validate_node_pool.go`, `api/manifest/...` after #2216) compares `ProviderSpec.Name` with the literal `"openstack"`. `ProviderSpec.Name` is the provider *instance* name, so:

- an OpenStack provider named anything else (e.g. `openstack-1`, as in the docs example) skips the check and `externalNetworkName` is never required;
- a non-OpenStack provider that happens to be named `openstack` is wrongly required to set it.

### Expected Behaviour
`externalNetworkName` is required when the referenced provider's *type* is `openstack`, regardless of the instance name.

### Steps To Reproduce
1. Define an OpenStack provider named `openstack-1` and a dynamic nodepool referencing it without `externalNetworkName`.
2. Apply the InputManifest.
3. Validation passes although the field is documented as required for OpenStack.

### Anything else to note
drop the `external_net` tag, callback and pretty-print case, add a test. 

`test-set1` sets the field on an OVH nodepool; drop it since this is intended for openstack.

MrBeldum:
## Summary
`validateExternalNet` compared `ProviderSpec.Name` (instance name) to the literal `"openstack"`, so OpenStack providers with other names skipped the check and non-OpenStack providers named `openstack` were wrongly required to set `externalNetworkName`.

- Validate via `Manifest.GetProviderType` (same pattern as spot/GPU checks)
- Drop the `external_net` tag, callback, and pretty-print case
- Remove `externalNetworkName` from OVH nodepools in `test-set1`
- Add focused unit tests

Fixes #2217

## Test plan
- [x] `go test ./api/manifest/ -run TestValidateExternalNet -v`
- [x] `go test ./api/manifest/`

## Summary by CodeRabbit

* **Bug Fixes**
  * OpenStack node pool validation now determines requirements by provider type rather than provider instance name.
  * OpenStack configurations still require an external network name, with validation errors reported consistently.
  * Non-OpenStack providers are no longer incorrectly subject to the OpenStack external network requirement.
  * Test configurations now verify that valid OpenStack and non-OpenStack scenarios are handled correctly.

### DIFF
--- a/api/manifest/manifest.go
+++ b/api/manifest/manifest.go
@@ -270,7 +270,7 @@ type ProviderSpec struct {
 	Zone string `yaml:"zone" json:"zone"`
 	// Name of the external provider network to which the nodes will be connected to. Currently only required for OpenStack.
 	// +optional
-	ExternalNetworkName string `validate:"external_net" yaml:"externalNetworkName" json:"externalNetworkName"`
+	ExternalNetworkName string `yaml:"externalNetworkName" json:"externalNetworkName"`
 }
 
 // StaticNodePool List of static nodepools of already existing machines, not created by Claudie, used for Kubernetes or loadbalancer clusters.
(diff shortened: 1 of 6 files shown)


Despire (approved):
LGTM, thanks.

